PR #124, opened 2022-05-19
Implement inline C++ blocks

linusg:


linusg:
Continued in #125 as this PR died in mysterious ways.

PR #126, opened 2022-05-20
Move file write responsibility from Compiler to main

ADKaster:
It'll be pretty hard to integrate into other repositories if the output is always called "output.cpp" :^)


ADKaster:
hmm this isn't baked enough. Needs a default binary dir.

ADKaster:
There we go. Now with yoinking a command line parser library to parse out --binary-dir or -o, and defaulting to $PWD/build.

linusg (changes requested):

tests/main.rs
```diff
@@ -17,9 +17,14 @@ fn test_samples(path: &str) -> Result<(), JaktError> {
...
 
+                let output_dir = path
+                    .parent()
+                    .expect("Input file must live somewhere")
+                    .to_path_buf();
+
                 if output_path.exists() || error_output_path.exists() {
                     // We have an output to compare to, let's do it.
-                    let mut compiler = Compiler::new();
+                    let mut compiler = Compiler::new(output_dir);
```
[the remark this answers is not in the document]
ADKaster: Makes sense to me, the change set looks a lot simpler now.,

Cargo.toml
```diff
@@ -8,6 +8,9 @@ edition = "2021"
 [dev-dependencies]
 uuid = {version = "1.0.0", features=["v4"]}
 
+[dependencies]
+clap = { version = "3.1.18", features = ["derive"] }
```
kleinesfilmroellchen, line 12: Clap is nice but extremely bloated. Look at the cargo.lock stuff it pulled in. There are a lot of other nice minimal command line argument libraries in Rust that should be by far enough until we rewrite the compiler in jakt.

src/main.rs
```diff
@@ -1,14 +1,46 @@
-use std::path::PathBuf;
+use std::{io::Write, path::PathBuf};
```
kleinesfilmroellchen, line 1: Not a problem but just noting for @jntrnr and others: how do we want to handle multi-layered use's? There's a rustfmt config but I can't remember whether it's stable yet, and there are arguments against and for both approaches.

src/main.rs
```diff
@@ -1,14 +1,46 @@
...
+
+use clap::Parser;
 
 use jakt::{Compiler, JaktError, Span};
 
+#[derive(Parser, Debug)]
+struct CLIArgs {
+    /// Binary directory
+    #[clap(long = "binary-directory", short = 'o')]
```
kleinesfilmroellchen, line 10: No spaces around key-value pairs in attribute macros, usually

src/main.rs
```diff
@@ -1,14 +1,46 @@
...
+
 fn main() -> Result<(), JaktError> {
+    let cli = CLIArgs::parse();
+
+    let binary_directory = match cli.binary_directory {
+        Some(dir) => dir,
+        None => {
+            let mut dir = std::env::current_dir().expect("Current directory unusable!");
+            dir.push("build");
+            dir
```
kleinesfilmroellchen, line 26: Can you not collapse this into one statement, or does that create a dropped temporary variable?

(to be fair I haven't worked much with Rust's path API yet)

kleinesfilmroellchen (changes requested):
Rust rating 9/10, just some nitpicks

src/main.rs
ADKaster: rustfmt complained in CI when I removed the spaces

src/main.rs
ADKaster: push doesn't return a value, if there's a better way i'm all ears

src/main.rs
jntrnr: I lean towards just using the default `rustfmt` style

src/main.rs
kleinesfilmroellchen: ah strange, seems like I don't write enough of these :)

src/main.rs
kleinesfilmroellchen: Sadly those APIs don't return this and the struct itself is the right choice. Then I'm fine with it.

Cargo.toml
ADKaster: Thanks for adding a suitable pico-args implementation! I need more rust under my belt to figure out a "clean" solution for using that.

linusg (commented):
Needs a commit prefix please (`compiler+main:`)

src/compiler.rs
```diff
@@ -104,13 +104,8 @@ impl Compiler {
...
 
-    pub fn compile(&mut self, fname: &Path) -> Result<(), JaktError> {
-        let cpp_string = self.convert_to_cpp(fname)?;
-
-        let mut out_file = std::fs::File::create("output.cpp")?;
-        out_file.write_all(cpp_string.as_bytes())?;
-
-        Ok(())
+    pub fn compile(&mut self, fname: &Path) -> Result<String, JaktError> {
+        self.convert_to_cpp(fname)
```
linusg, line 108: Now that this function only calls one other function, I'm not sure if it makes sense to actually keep it?

ADKaster: The rationale I told myself was that we'd later add other ways to compile, or other output formats, so it made sense to keep the wrapper. But if that's too forward thinking, for sure can remove :)

linusg:
It werks! :^)

PR #130, opened 2022-05-20
Improve error message when calling static methods like instance methods

drunderscore:
Previously, this would error with `wrong number of arguments`. While this is correct, calling a static method through an object instance doesn't make sense, so let's be more specific as to what is actually wrong with the call.

```
class Foo {
    function bar() {
        println("I am missing a this parameter")
    }
}

function main() {
    let foo = Foo()
    foo.bar() // TypecheckError "wrong number of arguments"
}
```

### DIFF
--- a/samples/classes/static_method_called_on_object_instance.error
+++ b/samples/classes/static_method_called_on_object_instance.error
@@ -1,0 +1,1 @@
+Cannot call static method on an instance of an object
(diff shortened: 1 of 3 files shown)


linusg (changes requested):
Please add a test for this :)

drunderscore:
@linusg Test added!

PR #135, opened 2022-05-20
Add support for underscores in numeric literals

jdahlin:
This will allow underscores in normal, binary and hex literals, e.g:

1234_5678
0x01_23_45
0b0000_1111

Which helps readablity and is supported by in many other popular languages.


samples/basics/binary_literals.jakt
```diff
@@ -2,4 +2,6 @@ function main() {
     println("{}", 0b10000000)
     println("{}", 0b111)
     println("{}", 0b11111111111110101)
+    println("{}", 0b0101_0101_0101_0101)
+    println("{}", 0b0101_0101_0101_0101_)
```
linusg, line 6: Leading or trailing underscore should not be valid.

jdahlin: Leading are not valid, rust allows trailing. Any particular reasoning for not allowing them?

linusg: This is not rust, and there's no point in making it valid, so just don't :)

linusg: and neither should more than one `_` be valid, FWIW

alimpfard: (am I the only one that prefers `123'456'789`?)

mustafaquraish: > (am I the only one that prefers `123'456'789`?)

That's always looked weird to me, especially considering we separate words in variable names with an underscore too

jdahlin: FWIW:
- Allows trailing underscore: Rust, Swift
- Allows subsequent underscores in the middle: Java, Rust, Swift
- Disallows trailing underscore: Python, Java, JavaScript, C#, Kotlin, Go, PHP, Ruby

jdahlin: @linusg same argument goes for this-is-not-Python-or-JavaScript. I'd argue for the current version as it's simpler to implement and slightly more flexible. I see no downside to allowing subsequent underscores and some minor upsides (multiple groups: `1111_1111__2222_2222`).

linusg: > same argument goes for this-is-not-Python-or-JavaScript

I only said that because you tried to justify this decision with "rust does it" 🤷

The issue with it is readability - `1______________2` isn't readable, so it's not needed.

> I'd argue for the current version as it's simpler to implement

Sorry, but that's a terrible argument for making language design decisions :)

linusg (commented):
LGTM, thanks! Please fix the rustfmt error and squash the commits :)

Issue #137, opened 2022-05-20
Array creation in functions that can't throw errors generates invalid cpp


HectorPeeters:
When allocating arrays in functions that don't return an error type, the compiler generates incorrect cpp code as it uses the `TRY` macro.

It might also be nice to annotate that the main function returns and `ErrorOr<T>` as that is not clear from the current syntax. In the example below, it looks like the two functions both return void but in main, the array allocation generates correct code while in test if does not.

Example:
```jakt
function test() {
    let mutable data = [0u32; 256]
}

function main() {
    test()
}
```

linusg:
Arguable this is a typechecker bug more that a codegen one, as the compiler should point out that a fallible operation (array creation) inside a function no marked `throws` is not permitted unless explicitly ignored (made "infallible", a.k.a crashing).

linusg:
See #153

lanmonster:
fixed as of d73ef3e

there is now an error message "Array initialization inside non-throwing scope"

PR #139, opened 2022-05-20
Convert the sample apps to explicit int comparisons

### LINKED ISSUE #131: Current sample apps don't compile
On the latest main branch, both the cat and crc32 sample programs fail with the following error: `Condition must be a boolean expression`.

To prevent this in the future, it might be a good idea to add these to `cargo test`.

## Cat

For the cat example, compilation fails because of the following line:
```jakt
20:    while not feof(file) {
```

Since feof returns a c_int and while expects a boolean. Changing the condition to `feof(file) == 0` does not work as comparisons between `c_int` and normal numeric types is not supported.

## CRC

For the crc example, the incorrect line is the following:
```jakt
14:    if value & 1 {
```

Here, `value & 1` does not result in a boolean value. Replacing this with `value & 1 != 0` would solve the problem but that again needs the `c_int` and numeric comparisons.

HectorPeeters:
- Change the sample apps to use explicit `c_int` comparisons instead of relying on an implicit cast which is not implemented at the moment
- Add support for `c_int` types in numeric compare expressions
- Make `make_lookup_table` throws (see #137)

Closes #131

### DIFF
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -16,6 +16,7 @@ use crate::{
         UnaryOperator, UncheckedType,
     },
 };
+use std::os::raw::{c_char, c_int};
 
 pub type StructId = usize;
 pub type EnumId = usize;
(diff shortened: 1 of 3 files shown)


src/typechecker.rs
```diff
@@ -54,8 +54,10 @@ pub fn can_fit_integer(type_id: TypeId, value: &IntegerConstant) -> bool {
     match *value {
         IntegerConstant::Signed(value) => match type_id {
             I8_TYPE_ID => value >= i8::MIN as i64 && value <= i8::MAX as i64,
+            CCHAR_TYPE_ID => value >= i8::MIN as i64 && value <= i8::MAX as i64,
             I16_TYPE_ID => value >= i16::MIN as i64 && value <= i16::MAX as i64,
             I32_TYPE_ID => value >= i32::MIN as i64 && value <= i32::MAX as i64,
+            CINT_TYPE_ID => value >= i32::MIN as i64 && value <= i32::MAX as i64,
```
linusg, line 60: - Let's put these at the top and next to each other, like in `is_integer
- Pretty sure we should be using `https://doc.rust-lang.org/std/os/raw/type.c_int.html` (and `c_char`) for MIN and MAX here, they're not guaranteed to be i8 and i32 on all platforms and architectures

src/typechecker.rs
```diff
@@ -453,8 +455,10 @@ impl IntegerConstant {
         let new_constant = match self {
             IntegerConstant::Signed(value) => match type_id {
                 I8_TYPE_ID => NumericConstant::I8(*value as i8),
+                CCHAR_TYPE_ID => NumericConstant::I8(*value as i8),
```
linusg, line 458: Likewise, and might need a conditional based on `c_int`'s actual size

HectorPeeters: I changed this one to work with the amount of bits and the signedness of the value instead of the individual types. This does not add any additional logic for the cchar, cint, and usize types.

src/typechecker.rs
```diff
@@ -50,9 +51,38 @@ pub fn is_integer(type_id: TypeId) -> bool {
     }
 }
 
+pub fn is_signed(type_id: TypeId) -> bool {
+    match type_id {
+        USIZE_TYPE_ID | U8_TYPE_ID | U16_TYPE_ID | U32_TYPE_ID | U64_TYPE_ID => false,
```
linusg, line 56: This is not sufficient:

> https://doc.rust-lang.org/std/os/raw/type.c_char.html
> On modern architectures this type will always be either `i8` or `u8`, as they use byte-addresses memory with 8-bit bytes.

so we need to check for that.

linusg (changes requested):

src/typechecker.rs
HectorPeeters: That is indeed a good point. From what I can see in the documentation, it looks like the only way to check if c_char is signed is by checking if `c_char::MIN` is smaller than zero. This seems like a hacky solution but I don't see another way to check for that.

linusg (commented):
LGTM now, thanks! Can you clean up the commit history a bit and squash the first, third and fourth commit into one? We don't need to preserve these iterations once merged :)

PR #140, opened 2022-05-20
Parser: Fix method call capturing spans incorrectly

### LINKED ISSUE #107: Method call expression captures incorrect span end value
image

image

- Off by one token
- They should be consistent with including or excluding the parentheses and arguments

reaganmcf:
Closes #107 

Prior to this patch, there was a one-off error for the end span, as well as inconsistent span capturing for method calls compared to function calls. This patch fixes this, making it more consistent.

By the way, huge fan of SerenityOS and @jntrnr's NuShell project. I hope my contribution is one of many to this project -  been on the lookout for a project like this for a while :)


jntrnr:
thanks!

PR #141, opened 2022-05-20
Add support for block comments

Undefine-Unity:


linusg:
There were multiple objections to this on Discord. Please also note that all new language features should come with tests :)

PR #143, opened 2022-05-20
Check if lexer index is out of bounds before checking for DotDot

Lubrsi:
When the lexer sees the dot, it increments the index into the `bytes`
array and then use the incremented index to check for another dot.
However, it did not check if this was out-of-bounds, meaning if a dot
is followed by EOF, the lexer would panic with an array OOB.


linusg (commented):
use -> uses in commit message

Lubrsi:
Fixed :+1:

PR #144, opened 2022-05-20
Add clippy check to github actions

HectorPeeters:
This might be too strict for the repository right now, but I think it would be beneficial to have this setup as the repository grows in size.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,11 +44,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - uses: actions-rs/toolchain@v1
-        with:
-            toolchain: nightly
-            components: clippy
-            override: true
+      - run: rustup component add clippy
       - uses: actions-rs/clippy-check@v1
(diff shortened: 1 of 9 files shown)


HectorPeeters:
Since the clippy action now fails for every single warning, there might be some things we want to disable like: https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call

Disabling this lint and auto-fixing the others leaves 23 warnings that have to be fixed by hand

.github/workflows/ci.yml
```diff
@@ -40,3 +40,16 @@ jobs:
         with:
           command: test
           args: --workspace ${{ matrix.flags }}
+  clippy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/toolchain@v1
+        with:
+            toolchain: nightly
```
jntrnr, line 49: can we use the stable one instead? that way people don't need the nightly one on their machines to double-check for upcoming clippy lints

jntrnr:
thanks!

PR #146, opened 2022-05-20
runtime: Add a simple File class for basic I/O

awesomekling:
Here's something a little bit nicer than using the standard C library.
Note that we're limited by what's supported on all of macOS, Linux and Windows.

### DIFF
--- a/runtime/IO/File.h
+++ b/runtime/IO/File.h
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2022, Andreas Kling <kling@serenityos.org>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <AK/Array.h>
+#include <AK/Error.h>
+#include <AK/RefCounted.h>
+#include <AK/String.h>
+#include <stdio.h>
+
+class File final : public RefCounted<File> {
+public:
+    static ErrorOr<NonnullRefPtr<File>> open_for_reading(String path);
+    static ErrorOr<NonnullRefPtr<File>> open_for_writing(String path);
+
+    ErrorOr<size_t> read(Array<u8>);
+    ErrorOr<size_t> write(Array<u8>);
+
+    ErrorOr<Array<u8>> read_all();
+
+    ~File();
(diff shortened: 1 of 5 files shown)


runtime/IO/File.h
```diff
@@ -0,0 +1,25 @@
...
+#include <AK/Array.h>
+#include <AK/Error.h>
+#include <AK/RefCounted.h>
+#include <AK/String.h>
+#include <stdio.h>
+
+class File final : public RefCounted<File> {
+public:
+    static ErrorOr<NonnullRefPtr<File>> open_for_reading(String path);
+    static ErrorOr<NonnullRefPtr<File>> open_for_writing(String path);
```
linusg, line 12: can these be const references, given that the parameters in the jakt declaration are not `mutable`?

linusg (commented):
Also, do we perhaps want to start adding copyright headers to source files like in serenity? :)

jntrnr:
> Also, do we perhaps want to start adding copyright headers to source files like in serenity? :)

assuming you mean the C++ we pull in and not the C++ we emit, that's probably not a bad idea...

linusg:
Right, human-written files only! :^)
In serenity we use SPDX headers, like this: https://github.com/SerenityOS/serenity/blob/74695ce76ea5f6b88d8cb166621edafd1abedb47/Userland/Libraries/LibJS/Runtime/Temporal/ZonedDateTime.cpp#L1-L6

runtime/IO/File.h
awesomekling: Hmm, interesting question. Of course they *could* because you and I know that it's safe.

However, jakt has no concept of "const references", it sees `String` as a value type. For that reason, it feels natural to pass `String` rather than `String const&`.

awesomekling:
> Also, do we perhaps want to start adding copyright headers to source files like in serenity? :)

Roger that!

runtime/IO/File.h
linusg: fair enough!

linusg (approved):
Nice, LGTM!

PR #149, opened 2022-05-20
Add initial support for sets!

mustafaquraish:
This PR adds support for `Set<T>`. You can construct sets using the `{1, 2, 3}` shorthand as suggested by Linus. There are a few limitations for now, which would require some overall changes to the way stuff is done, so I will save these for later:

1) You cannot define an empty set with `let set: Set<i64>  = {}`, this is because the code currently doesn't do anything to infer the type of the RHS based on the LHS of assignments. There's also a similar issue for arrays currently, which results in cpp code that doesn't compile.

2) There's no short-hand for sets, such as `{String}` instead of `Set<String>`. Should be easy enough to add in the future, but due to some minor parser trouble I'm leaving it out for the PR, I wanted to get a sense if the rest of this looks good enough first.

3) No reference-semantics for Sets currently, will do this in a future PR.


runtime/prelude.jakt
```diff
@@ -48,6 +48,18 @@ extern struct Dictionary<K, V> {
     function Dictionary<A, B>() -> Dictionary<A, B>
  }
 
+ extern struct Set<V> {
```
linusg, line 51: leading space (`jaktfmt` can't come soon enough!)

runtime/prelude.jakt
```diff
@@ -48,6 +48,18 @@ extern struct Dictionary<K, V> {
...
+    function contains(this, anonymous value: V) ->bool
```
linusg, line 52: `-> bool` pls, same below

runtime/prelude.jakt
```diff
@@ -48,6 +48,18 @@ extern struct Dictionary<K, V> {
...
+    function add(mutable this, anonymous value: V)
```
linusg, line 53: this should probably be marked `throws` to account for OOM?

runtime/prelude.jakt
```diff
@@ -48,6 +48,18 @@ extern struct Dictionary<K, V> {
...
+    function remove(mutable this, anonymous value: V) ->bool
+    function ensure_capacity(mutable this, anonymous capacity: usize)
```
linusg, line 55: same

runtime/AK/Set.h
```diff
@@ -0,0 +1,54 @@
...
+        for (auto& item : list)
+            add(item);
+    }
+
+    using HashTableType::contains;
+
+    HashSetResult add(const T& value) { return HashTableType::set(value); }
+    HashSetResult add(T&& value) { return HashTableType::set(move(value)); }
+    ErrorOr<HashSetResult> try_add(const T& value) { return HashTableType::try_set(value); }
+    ErrorOr<HashSetResult> try_add(T&& value) { return HashTableType::try_set(move(value)); }
```
linusg, line 34: For these we probably don't need infallible versions and can drop the `try_` prefix, that's mostly a artefact from moving towards OOM fallibility

mustafaquraish:
That should hopefully fix all the issues pointed out!

runtime/AK/Set.h
```diff
@@ -0,0 +1,52 @@
...
+    Set(std::initializer_list<T> list)
+    {
+        MUST(ensure_capacity(list.size()));
+        for (auto& item : list)
+            MUST(add(item));
+    }
+
+    using HashTableType::contains;
+
+    ErrorOr<HashSetResult> add(const T& value) { return HashTableType::try_set(value); }
```
linusg, line 31: Please run clang-format on this, which will format as east const (needs to be added to CI...)

linusg (approved):
one more tiny nitpick :)

runtime/AK/Set.h
mustafaquraish: Hmm, I did run clang-format on this, it doesn't seem to want to make it east-const. I'll do it manually.

mustafaquraish: Pushed the fix.

samples/sets/add.jakt
```diff
@@ -0,0 +1,9 @@
+function main() {
+    let mutable set = {"b", "c"}
+    set.add("a")
+    println("{}", set.size())
+
+    set.add("b")
+    set.add("c")
+    println("{}", set.size())
+}
```
MacDue, line 9: Don't know if this is considered an issue, but these files are ending missing newlines

samples/sets/unique_numbers.jakt
```diff
@@ -0,0 +1,8 @@
+function main() {
+    let nums = [1, 2, 1, 2, 1, 2, 3, 4, 5, 1, 2]
+    let mutable set = {nums[0]}
+    for i in 0..nums.size() {
+        set.add(nums[i])
+    }
+    println("{}", set.size())
+}
```
MacDue, line 8: Here too

PR #156, opened 2022-05-20
Add syntax highlighting for vim

dk949:
Currently vim plugin is in the `editors` directory, but vscode one is not. I think in the end vscode plugin should also be moved to `editors`, but it was somewhat out side of the scope of this PR.


jntrnr:
thanks!

PR #157, opened 2022-05-20
Create a VSCode launch configuration to debug current file compilation

kleinesfilmroellchen:
This launch configuration utilizes the CodeLLDB extension, the default debugging extension for Rust in VSCode. It will debug the jakt compiler running on the open file. So to debug any given file that has bugs in its compilation, simply open it and start the launch configuration from the "Run/Debug" window.


jntrnr:
thanks :)

PR #160, opened 2022-05-20
Fix infinite loop while parsing illegal function call arguments

### LINKED ISSUE #155: Parser hangs infinitely on unclosed brackets in statements
```
function main() {
    println("test"
}
```

Trace output:

```
...
parse_file
parse_function: Token { contents: Name("function"), span: Span { file_id: 1, start: 2, end: 10 } }
parse_block: Token { contents: LCurly, span: Span { file_id: 1, start: 18, end: 19 } }
parse_statement: Token { contents: Name("println"), span: Span { file_id: 1, start: 24, end: 31 } }
parsing expression from statement parser
parse_expression: Token { contents: Name("println"), span: Span { file_id: 1, start: 24, end: 31 } }
parse_operand: Token { contents: Name("println"), span: Span { file_id: 1, start: 24, end: 31 } }
parse_call: Token { contents: Name("println"), span: Span { file_id: 1, start: 24, end: 31 } }
parse_expression: Token { contents: QuotedString("test"), span: Span { file_id: 1, start: 32, end: 38 } }
parse_operand: Token { contents: QuotedString("test"), span: Span { file_id: 1, start: 32, end: 38 } }
...
parse_expression: Token { contents: RCurly, span: Span { file_id: 1, start: 39, end: 40 } }
parse_operand: Token { contents: RCurly, span: Span { file_id: 1, start: 39, end: 40 } }
ERROR: unsupported expression
parse_operator: Token { contents: RCurly, span: Span { file_id: 1, start: 39, end: 40 } }
ERROR: unsupported operator (possibly just the end of an expression)
...
```

kleinesfilmroellchen:
When the parser tries to parse an expression from "}" (for the second argument of the function), it errors out and doesn't increment the token index. But because the inner loop that repeatedly parses function parameter expressions never considered errors, it would get stuck in an infinite loop failing to parse "}".

The solution is simply to check for errors and bail out of the loop if there is one. A regression test has been added to the parser tests.

Fixes #155

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3898,7 +3898,16 @@ pub fn parse_call(tokens: &[Token], index: &mut usize) -> (Call, Option<JaktErro
                             index,
                             ExpressionKind::ExpressionWithoutAssignment,
                         );
-                        error = error.or(err);
+
+                        if let Some(jakt_error) = err {
+                            trace!(
+                                "ERROR: error while parsing expression in function call parameter"
+                            );
+                            error = error.or(Some(jakt_error));
+                            break;
+                        } else {
+                            error = error.or(err);
+                        }
 
                         call.args.push((param_name, expr));
                     }
--- a/tests/parser/unclosed-paren-in-function.jakt
+++ b/tests/parser/unclosed-paren-in-function.jakt
@@ -1,0 +1,7 @@
+// When the parser tries to parse an expression from "}" (for the second argument of the function),
+// it errors out and doesn't increment the token index.
+// But because the inner loop that repeatedly parses function parameter expressions never considered errors,
+// it would get stuck in an infinite loop failing to parse "}".
+function main() {
+    println("test"
+}
--- a/tests/parser/unclosed-paren-in-function.error
+++ b/tests/parser/unclosed-paren-in-function.error
@@ -1,0 +1,1 @@
+expected complete block


src/parser.rs
```diff
@@ -3898,6 +3898,12 @@ pub fn parse_call(tokens: &[Token], index: &mut usize) -> (Call, Option<JaktErro
                             index,
                             ExpressionKind::ExpressionWithoutAssignment,
                         );
+
+                        if let Some(jakt_error) = &err {
```
alimpfard, line 3902: How about checking if the index is changed instead?

kleinesfilmroellchen: Well, it might be perfectly fine to not increment the index sometimes? Either way I think it's more generic to handle the sudden appearance of an error, no matter the reason.

tests/parser/unclosed-paren-in-function.jakt
```diff
@@ -0,0 +1,7 @@
+// When the parser tries to parse an expression from "}" (for the second argument of the function),
+// it errors out and doesn't increment the token index.
+// But because the inner loop that repeatedly parses function parameter expressions never considered errors,
+// it would get stuck in an infinite loop failing to parse "}".
+function main() {
```
jntrnr, line 5: how is this test run?

jntrnr: Oops, I see it now. I missed we added other directories earlier

src/parser.rs
```diff
@@ -3898,6 +3898,14 @@ pub fn parse_call(tokens: &[Token], index: &mut usize) -> (Call, Option<JaktErro
...
```
jntrnr, line 3902: we probably want to fix the warnings for `jakt_error` not being used. Guessing this should be passed along to `error`?

kleinesfilmroellchen: Ah yes, sorry. I'm using rust-analyzer which doesn't give me Clippy warnings atm. I think I'll make an if-else out of this and just move err.

jntrnr:
thanks!

PR #162, opened 2022-05-21
Make Dictionary methods use fallible versions

mustafaquraish:
Previously Dictionary::set and Dictionary::ensure_capacity did not
propagate errors back up, this patch makes it so that we now use
the fallible versions


linusg:
Method -> methods in commit message

mustafaquraish:
> Method -> methods in commit message

woops, fixed

PR #163, opened 2022-05-21
Add shorthand for sets, and example in README

mustafaquraish:
What it says on the box. Not sure if the shorthand is the best idea, but personally I think it's nice, makes sets feel very much integrated into the language.


jntrnr:
looks good

PR #164, opened 2022-05-21
Type check invalid declarations

### LINKED ISSUE #154: Type checking does not reject invalid declarations
```
function main() {
    let foo: i8 = "Boo!";
}
```

Raises no errors and produces:

```cpp
#include "runtime/lib.h"



ErrorOr<int> _jakt_main(Array<String>)
{
{
        const i8 foo = String("Boo!");
    }
    return 0;
}
```

weltschildkroete:
This patch expands the typechecker to reject programs with incorrect type hints by calling `unify_with_type_hint` in more branches of the typechecking pass and propagating unification errors. This also fixes #154.


jntrnr:
Thanks for digging into this. Looks like you need to run `cargo fmt` so you can use the standard formatting.

weltschildkroete:
Thanks for the review! It seems I messed auto formatting in my editor somehow. Sorry!

In other news, I updated the PR to be up to date with main, and fixed some new tests that I believe were broken by this patch.

linusg (commented):
Please squash the last commit into the one where it broke, and use prefixes like `typechecker: ...` :)

tests/typechecker/return-value-type-mismatch.jakt
```diff
@@ -0,0 +1,3 @@
+function foo() -> u64 {
+        return "Foo"
```
linusg, line 2: Four spaces for indentation please

linusg (changes requested):

tests/typechecker/return-value-type-mismatch.jakt
weltschildkroete: Whoops, my bad. Still got to fix my editor. Fixed. :^)

tests/typechecker/return-value-type-mismatch.jakt
```diff
@@ -1,3 +1,3 @@
 function foo() -> u64 {
-        return "Foo"
+    return "Foo"
```
linusg, line 2: Please fix these in the commit that added them instead if introducing weird formatting and fixing it in a later commit

weltschildkroete: That makes sense, of course. Thanks! Fixed it, I think

tests/typechecker/return-value-type-mismatch.error
```diff
@@ -0,0 +1 @@
+Parameter type mismatch: String vs u64
```
drunderscore, line 1: I don't like this error message, which type is it expecting? How about something along the lines of:
`Expected "Foo", got "Bar"`

weltschildkroete: Yeah, I agree. I had just propagated the error message up from `check_types_for_compat`, but since that function is used in contexts other than function calls anyway, I changed the error message there to a clearer and more general type mismatch error message, as suggested. Thanks for the review!

PR #166, opened 2022-05-21
typechecker+tests: Implement return analysis

### LINKED ISSUE #152: Code paths with missing return are not rejected
```jakt
function foo() -> i64 {
    if true {
        return 42
    }
}
```

Generates:

```cpp
#include "runtime/lib.h"


i64 foo();

i64 foo(){
using _JaktCurrentFunctionReturnType = i64;
{
        if (true) {
            return (static_cast<i64>(42LL));
        }
    }
}
```

sin-ack:
This commit adds return analysis to the type checker. After each
statement in a block is checked, we check whether the statement we just
analyzed definitely returns. If it does, then we also mark the current
checked block as definitely returning. This allows us to propagate the
"definitely returns" information up to the function checker, which can
then verify that the function returns if it's not of a void type.

Closes #152


linusg (approved):
🎉

PR #172, opened 2022-05-21
Runtime integer overflow detection, the first

kleinesfilmroellchen:
As per the README, we want to make integer overflow or underflow a runtime error. This PR is the first step in this direction, as it adds overflow detection and a hard error (VERIFY) to integer addition. The implementation in the code generator is slightly hacked, as we can only simplify the match statements once we have the checked arithmetic functions for all operations. But nevertheless, it detects both integer overflow and underflow for all integer types.

Note that due to typechecker bugs/incompleteness, we can generate code that adds integers of different sizes. That's what would happen in the new underflow test if it weren't for the literal cast. The usage of __checked_add makes this a C++ compilation failure due to template substitution failure. As it is much more involved to check overflow with two different types being added, along with the implied language semantics (which of the two is the resulting type?), I believe this is fine, especially because it doesn't regress any tests and the typechecker should make the literal have the correct type anyways.

### DIFF
--- a/runtime/jakt_math.h
+++ b/runtime/jakt_math.h
@@ -1,0 +1,15 @@
+#pragma once
+
+#include <AK/Assertions.h>
+#include <AK/Concepts.h>
+#include <AK/NumericLimits.h>
+
+template<Integral T>
+T __checked_add(T lhs, T rhs)
+{
+    if ((rhs > 0) && (lhs > NumericLimits<T>::max() - rhs))
+        VERIFY(false && "Integer overflow");
+    if ((rhs < 0) && (lhs < NumericLimits<T>::min() - rhs))
+        VERIFY(false && "Integer underflow");
+    return lhs + rhs;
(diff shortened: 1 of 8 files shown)


runtime/jakt_math.h
```diff
@@ -0,0 +1,15 @@
...
+
+template<Integral T>
+T __checked_add(T lhs, T rhs)
+{
+    if ((rhs > 0) && (lhs > NumericLimits<T>::max() - rhs))
+        VERIFY(false && "Integer overflow");
+    if ((rhs < 0) && (lhs < NumericLimits<T>::min() - rhs))
+        VERIFY(false && "Integer underflow");
+    return lhs + rhs;
+}
```
linusg, line 15: We already have `AK::Checked`, let's not reinvent the wheel :)

linusg (changes requested):

PR #174, opened 2022-05-21
Add overflow checks to arithmetic operations

linusg:


src/codegen.rs
```diff
@@ -1507,6 +1614,26 @@ fn codegen_expr(indent: usize, expr: &CheckedExpression, project: &Project) -> S
...
+                    if is_integer(lhs.ty()) =>
+                {
+                    output.push_str(&codegen_checked_binary_op(indent, lhs, rhs, op, project))
+                }
+                BinaryOperator::AddAssign
+                | BinaryOperator::SubtractAssign
+                | BinaryOperator::MultiplyAssign
+                | BinaryOperator::DivideAssign
+                | BinaryOperator::ModuloAssign
+                    if is_integer(lhs.ty()) =>
```
kleinesfilmroellchen, line 1631: I think we should check the type of the expression itself instead of just the left hand side

kleinesfilmroellchen (commented):
I totally forgot we had Checked wrappers in AK. I guess this is the better solution, but I still like my stderr test capabilities in #172

src/codegen.rs
linusg: Done!

PR #175, opened 2022-05-21
Add the ability to test stderr + improve integer overflow stuff

kleinesfilmroellchen:
Do not merge before #174


kleinesfilmroellchen:
(cc @linusg )

linusg (commented):
LGTM, but let's get into the habit of using prefixes like `runtime:` and `codegen:` :)

PR #183, opened 2022-05-21
Handle out-of-range literals correctly

kleinesfilmroellchen:
If I'm not mistaken, this should handle all out-of-range literals with errors instead of silent truncation or overflow/underflow.


linusg:
same comment about using prefixes (`lexer: ...` etc) here :)

PR #185, opened 2022-05-21
Implement must unary operator

ccapitalK:
This PR adds a `must` keyword which acts as a unary operator that corresponds to AK's `MUST` macro. I'm not entirely sure if having a separate keyword for this is a good idea however, people I've talked to have stated that it might be better to fold this into the `!` operator. I mostly wanted to be consistent with how AK's abstractions for `Error`s works (since I am planning on proposing a similar keyword for `try`, which would also be a unary operator). What are people's thoughts on this?


awesomekling:
I like it.

ccapitalK:
I just realized I'll need to change this to work with our new way of handling errors (with try catch). I think what would make more sense is to make it something like a must block. Something like:

```
function main() {
  try {
    function_that_can_fail1()
    function_that_can_fail2()
  } catch error {
    println("Caught error {}", error.code)
  }

  must {
    function_that_can_fail1()
    ...
    function_that_can_fail2()
    ...
  }
}
```

awesomekling:
Yeah, it could be nice to have `must {xyz }`, but let's also support `must xyz`. :)

ccapitalK:
The syntax of the proposed must keyword has been updated. It now can be used in two forms:
- The must unary operator, used like `let x = must foo();`, `y + must(x + 3) + must 4`, etc. 
- Must blocks, which are like try blocks without the corresponding catch:
```
    must {
        foo();
        bar();
        baz();
    }
```

src/codegen.rs
```diff
@@ -1887,6 +1887,22 @@ fn codegen_statement(indent: usize, stmt: &CheckedStatement, project: &Project)
             output.push_str(&codegen_expr(indent, expr, project));
             output.push_str(";\n");
         }
+        CheckedStatement::Must(block) => {
+            // TODO: We should be showing the jakt source file/line number when an error is caught,
+            //       fails, not the source file/line number of the generated C++.
```
ccapitalK, line 1892: Will fix the extra 'fails,' in the next revision.

AtkinsSJ:

PR #186, opened 2022-05-21
Bounds check generic parameters before accessing

### LINKED ISSUE #180: Calling non-generic function with type name crashes typechecker
```jakt
function foo() {}

function main() {
    foo<String>()
}
```

```
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', src/typechecker.rs:3554:49
```

mattisboeckle:
Resolves an Issue, where calling a non-generic function with a type crashes the typechecker
resolves #180

### DIFF
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -3550,6 +3550,14 @@ pub fn typecheck_call(
                         typecheck_typename(type_arg, caller_scope_id, project);
                     error = error.or(err);
 
+                    if callee.generic_parameters.len() <= idx {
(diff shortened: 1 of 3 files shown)


linusg (commented):
Please prefix this with `typechecker: ...` :)

mattisboeckle:
Done

linusg (commented):
there's a rustfmt error unfortunately... also, I think a simple test would be nice to ensure this doesn't regress

PR #192, opened 2022-05-21
Syntax: Use literals for inline-cpp

Hendiadyoin1:
This comes with the extra benefit of having syntax highlighting for these blocks now

If we want to retain the string logic, I could try getting the same running in with the strings
but this way seems nicer.

The syntax was tested in VSCode by me and in CLion by @alimpfard

~~I could not fully verify this work because `cargo test` does not seem to do work out of the box for me~~


linusg (commented):
Please stick to the existing lowercase commit prefixes (`parser:` and `compiler:` in this case)

Hendiadyoin1:
Sure!

linusg (commented):
one more small typo: `editors` -> `editor's`. Could probably also say `inline C++` for readability :)

Implementation wise it's a nice change I guess, but falls apart really quickly:

```
cpp {
    /* } */
}
```

...so I'm not sure how I feel about it.

jntrnr:
We'd need a C++ lex of the block for this to work, I think. Cases like `{ "}" }`, `{ '}' }` and so on would also cause problems

awesomekling:
> 



> one more small typo: `editors` -> `editor's`. Could probably also say `inline C++` for readability :)
> 
> Implementation wise it's a nice change I guess, but falls apart really quickly:
> 
> ```
> cpp {
>     /* } */
> }
> ```
> 
> ...so I'm not sure how I feel about it.

Can we avoid this problem by using markers similar to C++ raw string literals?
Something like this:
```
cpp ~~~{


~~~}
```

It looks a bit weird, but it's the closest thing we already have today that I could think of :shrug:

awesomekling:
Let's close this as it's currently completely outdated. We can bring it back if we intend to work on this in the future.

PR #201, opened 2022-05-21
Use {K:V} for Dictionaries instead of [K:V]

mustafaquraish:
Based on a discussions we had on Discord yesterday, where I believe we came to a conclusion that is Sets are defined as `{a, b, c}`, then it makes sense to have Dictionaries be `{a:x, b:y, c:z}`. Also updating the shorthand types and README to reflect the syntax changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ let values = ["foo", "bar", "baz"]
 
 ```jakt
 function main() {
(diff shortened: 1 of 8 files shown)


awesomekling:
Directories use `[K:V]` notation matching the way you access them (`dict[key]` not `dict{key}`)

PR #204, opened 2022-05-21
typechecker+codegen: Infer types of empty Array/Set literals

mustafaquraish:
Previously, the following code did not generate C++ code that compiles:

```
let x: [i64] = []
let y: {i64} = {}
```

Since the template types for `Set` and `Array` could not be inferred
from empty initializer lists. The workaround for this was to use the
type of the first element of the literal to generate the type, but
this did not work for empty literals.

This commit looks at the actual type that's stored within the project
and extract the inner template types when generating code, so it can
handle empty literals as well.


src/typechecker.rs
```diff
@@ -2415,8 +2419,12 @@ pub fn typecheck_expression(
                 .find_struct_in_scope(0, "Set")
                 .expect("internal error: Set builtin definition not found");
 
-            let type_id =
-                project.find_or_add_type_id(Type::GenericInstance(set_struct_id, vec![inner_ty]));
+            // FIXME: Make sure type_hint is of type Set<T>
+            let type_id = if inner_ty == UNKNOWN_TYPE_ID && type_hint.is_some() {
+                type_hint.unwrap()
```
mustafaquraish, line 2424: This is a little ugly, but I can't think of a nice way to not use `.unwrap()` here without breaking up into something like:
```rust
                let type_id = if let Some(type_hint) = type_hint {
                if inner_ty == UNKNOWN_TYPE_ID {
                    type_hint
                } else {
                    project.find_or_add_type_id(Type::GenericInstance(set_struct_id, vec![inner_ty]))
                }
            } else {
                project.find_or_add_type_id(Type::GenericInstance(set_struct_id, vec![inner_ty]))
            };
```
which doesn't necessarily look that much better

src/typechecker.rs
```diff
@@ -2380,8 +2380,12 @@ pub fn typecheck_expression(
                 .find_struct_in_scope(0, "Array")
                 .expect("internal error: Array builtin definition not found");
 
-            let type_id =
-                project.find_or_add_type_id(Type::GenericInstance(array_struct_id, vec![inner_ty]));
+            // FIXME: Make sure type_hint is of type Array<T>
```
alimpfard, line 2383: My suggestion is to do something similar to `unify_with_type_hint` instead of the current "either the type hint or the known type", this allows things like generics to resolve correctly too.

linusg (commented):
Inference -> Infer, and needs a prefix like `typechecker+codegen:` please :)

jntrnr:
Yeah, looks okay. At some point we'll want start moving away from UNKNOWN_TYPE_ID towards a more complete inference. For now, working around the limitations so we can get better examples in place seems like a good trade-off to me.

PR #208, opened 2022-05-22
typechecker+tests: check throwing of function calls

### LINKED ISSUE #153: Point out missing `throws` for functions that can fail
Currently, if a `foo()` not marked with `throws` calls another function `bar()` which does throw an error, the compiler doesn't point this out, and instead generates incorrect cpp code. This should ideally be caught before we generate any code. 

Example:
```
function bar() throws -> i64 { return 5 }
function foo() -> i64 { return bar() }
```

AndrewBregger:
Function calls get checked for if they throw and if they can be safely handled at the call site.
Situations:
* calling function throws: call as normal
* calling function doesn't throw: must be called within a try-catch block

Adds error message: `try to call a throwing function <called-function> in a non-throw function <calling-function>`

All tests were updated with try-catch blocks that were calling throwing function/methods in non-throwing functions.
Resolves #153

### DIFF
--- a/samples/arrays/array_push.jakt
+++ b/samples/arrays/array_push.jakt
@@ -2,7 +2,7 @@ function main() {
     let mutable values = [0; 0]
 
     for i in 0..10 {
-        values.push(i)
+        try values.push(i) catch err {}
     }
 
     while not values.is_empty() {
--- a/samples/basics/is.jakt
+++ b/samples/basics/is.jakt
@@ -7,21 +7,23 @@ class Bar {
 }
 
 function main() {
-    let foo = Foo(x: 1)
+    try {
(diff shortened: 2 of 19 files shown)


linusg (changes requested):
Please largely follow the [SerenityOS code submission policy](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy) for commits, including:

- no spaces around `+` in commit categories
- Description after category starts with uppercase letter
- Ensure commit messages don't exceed a length of 72 chars

Additionally, please fix any broken tests in the commits that break them, not separately. For each given commit, all tests must pass at that point in time.

PR #210, opened 2022-05-22
Meta: Add script for build+running Jakt programs

jedieaston:
I thought the two commands to get a program built all the way was
somewhat unwieldy, so I added a little bash script to do it.

Much like Serenity, it's in a meta folder off the root. If there's anything that's super wrong please let me know, my bash scripting is rustier than I'd like to admit.

### DIFF
--- a/meta/run.sh
+++ b/meta/run.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+
+print_help() {
+    NAME=$(basename "$0")
+    cat <<EOF
+Usage: $NAME [arguments...] [arguments to Jakt...]
+Build and run a Jakt program!
+Optional arguments:
+    --build-compiler: Build the Jakt compiler before compiling the program (equivalent to running cargo build first).
+    --help: Print this message.
+EOF
+}
+
+if [ $# -eq 0 ]; then
+    print_help
+    exit 1
+fi
+
+BUILD_JAKT=0
+
+for argument in "$@"
+do
+    if [ $argument = "--help" ]; then
+        print_help
+        exit
+    fi
+    if [ $argument = "--build-compiler" ]; then
+        BUILD_JAKT=1
+        shift
+    fi
+done
+
+# Check to make sure clang is installed, since we need it to run the program.
+if ! command -v clang++ &> /dev/null
+then
+    echo "clang++ was not found on your PATH, and it's required for building the output C++ program. Install it, then come back here :^)"
+    exit 1
+fi
+
+# Assuming that jakt is in the normal place after it's built...
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+if [ $BUILD_JAKT = 1 ] || [ ! -f $SCRIPT_DIR/../target/debug/jakt ]; then
+    cargo build
+fi
+
+$SCRIPT_DIR/../target/debug/jakt "$@"
+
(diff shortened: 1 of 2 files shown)


meta/run.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+
+print_help() {
+    NAME=$(basename "$ARG0")
+    cat <<EOF
+Usage: $NAME [arguments...] [SOURCEFILE]
```
kleinesfilmroellchen, line 6: Mind you that jakt might gain other arguments in the future. Just replace "SOURCEFILE" with "arguments to jakt" or something. Also see #213 .

kleinesfilmroellchen (changes requested):

meta/run.sh
jedieaston: Fixed.

kleinesfilmroellchen (approved):

meta/run.sh
```diff
@@ -0,0 +1,56 @@
...
+$SCRIPT_DIR/../target/debug/jakt "$@"
+
+JAKT_EXIT=$?
+if [ $JAKT_EXIT != 0 ]; then
+    echo "Compilation failed :^("
+    exit $JAKT_EXIT
+fi
+
+clang++ -std=c++20 -Iruntime -Wno-user-defined-literals -o jakt.out output.cpp 
+./jakt.out
```
linusg, line 56: No trailing whitespace, but trailing newline please

linusg (changes requested):
> I thought the two commands to get a program built all the way was
somewhat unwieldy, so I added a little bash script to do it.

...but forgot to document its existence anywhere?

Also, please not period at the end of commit titles.

meta/run.sh
```diff
@@ -0,0 +1,56 @@
...
+
+$SCRIPT_DIR/../target/debug/jakt "$@"
+
+JAKT_EXIT=$?
+if [ $JAKT_EXIT != 0 ]; then
+    echo "Compilation failed :^("
+    exit $JAKT_EXIT
+fi
+
+clang++ -std=c++20 -Iruntime -Wno-user-defined-literals -o jakt.out output.cpp
```
ADKaster, line 55: I broke this in adcaeb5a459a1fc147437b4d934041ad965d4f9f , it now will be "$@".cpp or.. since that's all the arguments, it'll be $1.cpp ? $last-arg.cpp ? something like that.

mustafaquraish: For what it's worth, this is what I've been using in my bash script to run stuff:
```bash
filename=$(basename -- "$1")
filename="${filename%.*}.cpp"

clang++ ... $filename ...
```

meta/run.sh
jedieaston: Fixed, along with adding a note about it to README.md. Sorry I completely spaced on documentation, I wasn't 100% awake when I was doing this :)

meta/run.sh
jedieaston: I used a modification of that snippet to find the filename, thanks @mustafaquraish! The way I have it if multiple Jakt files are passed to the compiler we will try to tell clang to use the last one passed.

mustafaquraish:
Since #299 has been merged, I think we can leverage that functionality instead of also mirroring it in this shell script

awesomekling:
> Since #299 has been merged, I think we can leverage that functionality instead of also mirroring it in this shell script

Agreed.

PR #211, opened 2022-05-22
codegen: Place the 'static' keyword in the right position

alimpfard:
Previously we put it right before 'template' if the function was
templated.

Also add a test I forgot to commit for enums.


alimpfard:
Of note: this makes all regular functions static (i.e. ones with internal linkage).

samples/enums/underlying.jakt
```diff
@@ -0,0 +1,13 @@
+enum CuteLittleEnum: u64 {
```
kleinesfilmroellchen, line 1: my little enum can't possibly be this cute

samples/enums/underlying.jakt
```diff
@@ -0,0 +1,13 @@
...
+    A = 1
+    B
+    C = 69
```
kleinesfilmroellchen, line 4: What about a test for an initial enum variant without an explicit value? That should be 0, right, how about we check that?

PR #213, opened 2022-05-22
main: Use pico-args for better CLI

kleinesfilmroellchen:
pico-args is a zero-dependency minimal argument parsing library that just handles the bare basics of argument parsing. We use it here to provide a basic CLI that's a little more ergonomic than before, plus some help output. This is mainly preparatory work for subcommands, specifying output files, etc. which require more sophistication than manual argsparsing.

The main two reasons for choosing pico-args is (1) it's super tiny and fast, keeping the development cycle snappy even with full builds, and (2) most argparsing things are handled in our own code, making it easier to port to jakt later on. We'd just have to replicate the low-level parsing facilities of pico-args, which are not too complex in comparison to things that structopt or clap would handle.

CC @ADKaster, can you rebase your work in #126 on this?

Compile times before:
```
❯ time cargo build; time cargo build --release
   Compiling jakt v0.1.0 (/home/kleines/jakt)
    Finished dev [unoptimized + debuginfo] target(s) in 1.96s

real    0m2.015s
user    0m3.539s
sys     0m0.237s
   Compiling jakt v0.1.0 (/home/kleines/jakt)
    Finished release [optimized] target(s) in 3.74s

real    0m3.797s
user    0m10.165s
sys     0m0.117s
```

Compile times after:
```
❯ time cargo build; time cargo build --release
   Compiling pico-args v0.4.2
   Compiling jakt v0.1.0 (/home/kleines/jakt)
    Finished dev [unoptimized + debuginfo] target(s) in 2.10s

real    0m2.156s
user    0m3.837s
sys     0m0.347s
   Compiling pico-args v0.4.2
   Compiling jakt v0.1.0 (/home/kleines/jakt)
    Finished release [optimized] target(s) in 3.83s

real    0m3.892s
user    0m10.717s
sys     0m0.209s
```

Binary size before and after is 4.2 MiB, so a difference of less than 100K (pico-args is 11K according to crates.io)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,13 +1,17 @@
-use std::path::PathBuf;
+use std::{path::PathBuf, process::exit};
+
+use pico_args::Arguments;
 
 use jakt::{Compiler, JaktError, Span};
 
 fn main() -> Result<(), JaktError> {
+    let arguments = parse_arguments();
+
     let mut compiler = Compiler::new();
     let mut first_error = None;
 
-    for arg in std::env::args_os().skip(1) {
-        match compiler.compile(&PathBuf::from(&arg)) {
+    for file in arguments.input_files {
+        match compiler.compile(&file) {
             Ok(_) => {}
             Err(err) => {
                 match &err {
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -1,13 +1,17 @@
-use std::path::PathBuf;
+use std::{path::PathBuf, process::exit};
 
 use jakt::{Compiler, JaktError, Span};
 
+use pico_args::Arguments;
```
linusg, line 5: Isn't the usual `use` order standard library, third-party, first-party?

src/main.rs
```diff
@@ -28,6 +32,68 @@ fn main() -> Result<(), JaktError> {
...
+const HELP: &str = "\
+Flags:
+  -h, --help           Print this help and exit 
+
+Options:
+  FILES...             List of files to compile. The output is `output.cpp`
+                       in the cwd.
+
+Jakt programming language (c) 2022 the SerenityOS developers.
+Licensed under BSD 2-Clause, see LICENSE.
```
linusg, line 48: I don't think this needs to appear in CLI output

kleinesfilmroellchen: Maybe I've seen this too often in help output

src/main.rs
kleinesfilmroellchen: idk, rustfmt doesn't complain 🤷

linusg: Might not be the default, but I think there's a config option for it, I remember using it for serenity-rust's config.

PR #214, opened 2022-05-22
Typechecker: Add type inferring to return values of functions

### LINKED ISSUE #151: Return type inference doesn't reject incompatible returns
```jakt
function foo() {
    return true
    return "foo"
}
```

```cpp
#include "runtime/lib.h"


String foo();

String foo(){
{
        return (true);
        return (String("foo"));
    }
}
```

mattisboeckle:
Should resolve #151 when finished


mattisboeckle:
Still nees functionality for conditional returns.
Also, two of the old tests use (wrong) functionality that this pull request removes. Those will need to be fixed. See the issue for information on which ones

linusg (changes requested):
> "=>" functions can no longer return void

Who agreed to such a change?

---

Please:

- Fix tests in the same commit as you break them, not separately. Each commit should have all tests at that point in time passing
- Reword your commit messages to largely match the [SerenityOS guidelines](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#code-submission-policy) (working on making something Jakt-specific) including wrapping lines at 72 chars and using `Add`, not added

mattisboeckle:
@linusg Thanks for the review, still new to this.

I'm not even sure, if my ideas are right or I misunderstood the Issue.

Should functions that don't have an explicit return type be inferred by return statements or should there never be returns in those functions as they should return void.

For example is this valid
```
function foo() {
    return true
}
```
because it is the same as
```
function foo() -> bool {
    return true
}
```
or should it be rejected?

If those functions are valid then some of the assumptions made in my code don't make sense and we should scratch this PR

mattisboeckle:
- Squashed / Fixuped commits into 2
- reworded my commit messages
- "=>" functions can return void values again

linusg (commented):
First still says Added, third starts with lowercase word, and I see lines that are not wrapped at 72 chars

PR #215, opened 2022-05-22
Typechecker: Infer return type from first return in function body

Lubrsi:
Marked as draft as I'm not sure how to test this.

Previously, if the return type was unknown, we would infer the return
type from the last return statement in the top level block of a
function body. However, this would not account for child blocks, so
this example would be inferred as `void`:

```jakt
function foo() {
    if (true) {
        return 1
    }
}
```

Now, if the return type is unknown, and the function contains a return
statement anywhere in the function body, we infer the return type from
the first return statement in the function body. For example:

```jakt
function foo() {
    if (false) {
        return 1
    } else {
        return true
    }
    return "foo"
}
```

We will infer the return type to be i64 from `return 1`, as it appears
before `return true` and `return "foo"`, even though `return 1` is
unreachable.

This is because we cannot know ahead of time which blocks may be
executed, for example, if the if statement was `if (rand() % 2 == 0)`
instead of `if (false)`.

### DIFF
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -1765,6 +1765,136 @@ fn typecheck_function_predecl(
     error
 }
 
+fn get_return_type_of_statement(statement: &CheckedStatement) -> TypeId {
+    match statement {
+        CheckedStatement::Return(ret) => ret.ty(),
+
+        CheckedStatement::If(_, then_block, maybe_else_statement) => {
+            // Examples:
+
+            // if (true) {
+            //     true
+            // }
+            // Inferred type: void
+
+            // if (true) {
+            //     true
+            // } else {
+            //     2
+            // }
+            // Inferred type: void
+
+            // if (true) {
+            //     true
+            // } else {
+            //     return 2
+            // }
+            // Inferred type: i64
+
+            // if (true) {
+            //     return true
+            // } else {
+            //     return 2
+            // }
+            // Inferred type: bool
+
+            let then_block_return_type =
+                get_return_type_from_first_return_statement_in_block_tree(then_block);
+
+            if then_block_return_type != VOID_TYPE_ID {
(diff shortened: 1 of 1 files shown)


src/typechecker.rs
```diff
@@ -1765,6 +1765,136 @@ fn typecheck_function_predecl(
...
+            let statement_return_type = get_return_type_of_statement(statement.as_ref());
+
+            if statement_return_type != VOID_TYPE_ID {
+                return statement_return_type;
+            }
+
+            get_return_type_from_first_return_statement_in_block_tree(catch_block)
+        }
+
+        _ => VOID_TYPE_ID,
```
alimpfard, line 1867: `Match` may contain returns too.

mattisboeckle:
I think you misunderstood the Issue. Your examples should not compile, because there is a return in a void function.

Always choosing the first return does not fix the issue. There would always be runtime errors when something like 
```
function foo() {
    if (false) {
        return 1
    } else {
        return true
    }
}
```
is called, as the inferred type would be i64 (or whatever is the default) and the return value is a bool.
Defeating the purpose of the Typechecker

Lubrsi:
Yeah, this is supposed to be combined with another part of the typechecker that takes this inferred return type and rejects any future returns that doesn't match the type.

mattisboeckle:
You were working on #151 right? I made a Pull-Request for that Issue, but now I'm not sure if my solution is correct.

Does no explicit return type mean that the function has a void return type or does a return statement imply the return type

Lubrsi:
No explicit return type depends if there is a return statement or not. If there isn't, the inferred type is `void`. If there is one, the inferred type is the type the return statement is returning. The typechecker in main does this, but only for the very last return statement in the top level block of the function body.

AtkinsSJ:

PR #219, opened 2022-05-22
Typechecker: add typechecking for binary operations

### LINKED ISSUE #217: Binary operators not typechecked
At the moment, the normal binary operators (+, -, *, /, ...) are not typechecked at all.

Implementation wise, I think it would be best if we only allow the same types on both the left and right side without any upcasting. This adds some additional type casts but makes it more clear what exactly is happening in the code.

An example of what wouldn't work in that scenario:

```jakt
function main() {
    let x: u8 = 12;
    let y: u16 = 34;
    println("{}", x + y) // ERROR: binary operation between incompatible types
}
```

In the future it would be a good addition to allow casting literals to the correct type so the following expression would work:

```jakt
let x: u8 = 12;
println("{}", x + 34);
```

HectorPeeters:
Add typechecking for:

- Comparison operators (left and right have to be the same type)
- Logical and and or (left and right have to be booleans)
- +, -, *, /, % (left and right have to be the same type)

The mathematical binary operators (+, -, *, /, ...) are still somewhat explicit as casts have to be inserted in some places. While this makes it more explicit, especially when working with multiple different integer types, this might not be desired.

Right now it is not possible to do the following:

```jakt
let x: i64 = array.size() + 12;
```

A cast would be needed to cast the `usize` to an `i64`.

Closes #217

### DIFF
--- a/samples/arrays/prefill.jakt
+++ b/samples/arrays/prefill.jakt
@@ -2,7 +2,7 @@ function main() {
     let v = [85; 3]
     println("{}", v.size())
     let mutable i = 0
-    while i < v.size() {
+    while i < v.size() as! i64 {
         println("{}", v[i++])
     }
 }
--- a/samples/arrays/reference_semantics.jakt
+++ b/samples/arrays/reference_semantics.jakt
@@ -7,7 +7,7 @@ function main() {
     change_value(vector: v)
 
     let mutable i = 0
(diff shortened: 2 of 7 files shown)


HectorPeeters:
Also, I'm getting some clippy errors for code that I didn't touch so I guess there might be some unlinted code in the main branch.

PR #223, opened 2022-05-22
parser+typechecker: Visibility round two, electric boogaloo

kleinesfilmroellchen:
Now with 100% more JoJo references in tests, this PR brings you visibility modifiers on fields + some refactoring niceties to avoid rightwards drift. Visibility can now be considered fully implemented, unless we want `protected` once inheritance lands.

### DIFF
--- a/samples/classes/method.jakt
+++ b/samples/classes/method.jakt
@@ -1,6 +1,6 @@
 class Person {
-    name: String
-    age: i64
+    public name: String
+    public age: i64
 
     public function birthday(mutable this) {
         ++this.age
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -589,6 +589,7 @@ pub struct CheckedVarDecl {
     pub type_id: TypeId,
     pub mutable: bool,
     pub span: Span,
+    pub visibility: Visibility,
 }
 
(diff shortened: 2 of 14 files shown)


samples/classes/method.jakt
```diff
@@ -1,6 +1,6 @@
 class Person {
     name: String
```
linusg, line 2: I know it doesn't matter for this example, but for consistency with the previous and following test I'd make them both public.

src/typechecker.rs
```diff
@@ -994,6 +996,79 @@ impl Scope {
...
+    span: &Span,
+    project: &Project,
+) -> Option<JaktError> {
+    if member.visibility() != Visibility::Public
+        && !Scope::can_access(own_scope, member_scope, project)
+    {
+        Some(JaktError::TypecheckError(
+            // FIXME: Improve this error
+            format!(
+                "Can't access {} `{}` from scope {:?}",
```
linusg, line 1060:
```suggestion
                "Can't access {} '{}' from scope {:?}",
```

linusg (commented):
Last commit can be squashed IMO. Thanks for fixing tests as you're breaking them, I've seen so many PRs not doing that :^)

Issue #170, opened 2022-05-21
Weak references


awesomekling:
`weak T?` should behave essentially like `Optional<T&>` in C++ but with the added feature that it automatically empties itself when the `T` is destroyed.

`weak T` is not allowed, since that implies it's always in a dereferenceable state.

sin-ack:
Hmm, shouldn't `weak T` imply `T?`? Because there is no other possible state for a weak pointer, I don't really see the point of requiring additional tokens when `weak` provides adequate meaning regardless.

awesomekling:
The `?` token is for readability, and symmetry with other optional types.

PR #227, opened 2022-05-22
Add weak references

awesomekling:
You can now declare a weak reference to T via `weak T`

These are represented as `WeakPtr<T>` in the resulting codegen.
The goal is for them to feel and behave like `Optional<T>`, despite
being backed by a different C++ class.

Weak references can be force-unwrapped with postfix bang (`foo!`)

Work towards #170


samples/weak/assign_weak_to_weak.jakt
```diff
@@ -0,0 +1 @@
+OK
```
sin-ack, line 1: Seems like this accidentally became a .jakt file rather than an .out file.

samples/weak/basic.jakt
```diff
@@ -0,0 +1,19 @@
+class Foo {
+    public function hello(this) => "friends"
+}
+
+function main() {
+    let mutable weak_foo: weak Foo = None
```
sin-ack, line 6: Per earlier discussion it was agreed that this would use `weak Foo?`, did that change?

samples/weak/basic.jakt
```diff
@@ -0,0 +1,19 @@
...
+    {
+        let foo = Foo()
+        weak_foo = foo
+
+        println("weak_foo has_value? {}", weak_foo.has_value())
+        println("weak_foo hello: {}", weak_foo!.hello())
+    }
+
+    println("weak_foo has_value? {}", weak_foo.has_value())
+}
```
sin-ack, line 19: Missing newline at the end of file

src/codegen.rs
```diff
@@ -378,8 +378,8 @@ fn codegen_struct(structure: &CheckedStruct, project: &Project) -> String {
     match structure.definition_type {
         DefinitionType::Class => {
             output.push_str(&format!(
-                "class {} : public RefCounted<{}> {{\n",
-                structure.name, structure.name
+                "class {} : public RefCounted<{}>, public Weakable<{}> {{\n",
```
sin-ack, line 381: In the future, it would be nice to only enable `Weakable` for types that have `weak T`s pointing to them.

samples/weak/basic.jakt
awesomekling: That didn't change. Hence "Work towards #170" instead of "Fixes #170" :^)

src/codegen.rs
awesomekling: That seems inconvenient to support across library boundaries

Issue #184, opened 2022-05-21
Compiler should reject incorrect main function type annotations


linusg:
Given that the main function is special and hardcodes `ErrorOr<int>` as its return type in codegen, the typechecker should reject any attempts of giving it:

- A return type that is not `throws -> c_int`
- Parameters other than the current `Array<String>`

E.g.

```jakt
function main(x: i32) -> String {
    return ":^)"
}
```

Generates:

```cpp
#include "runtime/lib.h"

ErrorOr<int> _jakt_main(const i32 x)
{
    using _JaktCurrentFunctionReturnType = ErrorOr<int>;
    {
        return (String(":^)"));
    }
    return 0;
}
```

PR #228, opened 2022-05-22
typechecker: Check jakt `main` type declaration

Pythonen:
Add separate typecheck for main functions as stated in https://github.com/SerenityOS/jakt/issues/184
Also fix tests to use this new declaration.

### DIFF
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -1794,6 +1794,9 @@ fn typecheck_function(
         .find_function_in_scope(parent_scope_id, &function.name)
         .expect("Internal error: missing previously defined function");
 
+    if function.name == "main" {
+        error = typecheck_jakt_main(function);
+    }
     let checked_function = &mut project.functions[function_id];
     let function_scope_id = checked_function.function_scope_id;
     let function_linkage = checked_function.linkage;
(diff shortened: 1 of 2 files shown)


linusg (changes requested):
>  Also fix tests to use this new mandatory declaration.

I never said they should be mandatory...

Pythonen:
> > Also fix tests to use this new mandatory declaration.
> 
> I never said they should be mandatory...

Sorry, now I'm a bit lost. So the type declaration for main should be inferred if not explicitly typed by the user or?

linusg:
Right, I think inheriting a bit of magic from C/C++ isn't a bad thing here. Personally I think it should be fine to:

- Omit the return type declaration and not return anything -> implicit `return 0`
- Omit the return type declaration and return a number
- Omit the args parameter if they are not being used

In other words, 

```jakt
function main() {}
```

is the same as

```jakt
function main(args: [String]) throws -> c_int {}
```

but that does not mean that

```jakt
function main(foo: i32, bar: i32) -> String {}
```

should be accepted.

Pythonen:
Right, my bad. Misunderstood it.

src/typechecker.rs
```diff
@@ -1844,6 +1847,53 @@ fn typecheck_function(
...
+        "Main function must take a single array of strings as its parameter".to_string(),
+        function.name_span,
+    ));
+
+    if function.params.len() > 1 {
+        return param_type_error;
+    }
+
+    if !function.params.is_empty() {
+        match &function.params[0].variable.ty {
```
linusg, line 1861: Just noticed I missed this in my `ty` removal & renamed it now, please rebase/update

src/typechecker.rs
```diff
@@ -1844,6 +1847,53 @@ fn typecheck_function(
...
+                        return None;
+                    }
+                }
+                _ => {
+                    return param_type_error;
+                }
+            },
+            _ => {
+                return param_type_error;
+            }
```
linusg, line 1877:
```suggestion
            ParsedType::Array(ty, _) => match &**ty {
                ParsedType::Empty => {}
                ParsedType::Name(ty, _) if ty == "String" => {}
                _ => return param_type_error,
            },
            _ => return param_type_error,
```

src/typechecker.rs
```diff
@@ -1844,6 +1847,53 @@ fn typecheck_function(
...
+        function.name_span,
+    ));
+    match &function.return_type {
+        ParsedType::Empty => {}
+        ParsedType::Name(name, _) => {
+            if name != "c_int" {
+                return return_type_error;
+            }
+        }
+        _ => return return_type_error,
```
linusg, line 1892:
```suggestion
        ParsedType::Empty => {}
        ParsedType::Name(name, _) if name == "c_int" => {}
        _ => return return_type_error,
```

linusg (commented):
Looks much better, thanks! Please rebase, and squash the commits while you're at it :)
For each given commit in time, all tests should be passing, so fixing one after the fact is not desirable and makes bisecting harder.

src/typechecker.rs
Pythonen: Is it okay though to use it in 
`ParsedType::Array(ty, _) => match &**ty`
or should these destructurings also have more verbose names?

linusg: it's not ok to use anywhere :)

linusg (changes requested):
Typo: `programs'` -> `program's`, and please clean up the commit message, it now contains all the messages from the squashed commits

Pythonen:
Sorry for the hassle, my git-fo is a bit rusty (heh)

PR #230, opened 2022-05-22
Typechecker: Prepare typenames before any typechecking for better name binding

### LINKED ISSUE #203: Typechecker isn't making all user types available before checking them
Example from discord:

```jakt
Error: unknown type
-----
struct Foo {
    function bar(mutable this, baz: mutable Bar) {} // <-- error here on Bar
}

class Bar {}

function main() {
    let foo = Foo()
    let baz = Bar()
    foo.bar(baz)
}
```

What we currently do is check method prototypes in `typecheck_struct_predecl` but that doesn't allow all the user types to be known in that scope before they're looked up. We should go through all the user types for the scope *first* before doing the predecl so that the names are available, then we have something to bind to.

jntrnr:
This makes all typenames visible and ready for binding before any proper typechecking begins. This prevents issues where types have to be ordered in a particular way for visibility.

fixes #203


src/typechecker.rs
```diff
@@ -1044,6 +1038,19 @@ pub fn typecheck_namespace(
         {
             error = error.or(Some(err));
         }
+    }
+
+    for (struct_id, structure) in parsed_namespace.structs.iter().enumerate() {
+        //Ensure we know the types ahead of time, so they can be recursive
```
linusg, line 1044:
```suggestion
        // Ensure we know the types ahead of time, so they can be recursive
```

linusg (commented):
Let's keep things consistent and use commit prefixes, i.e. `typechecker: Description goes here` :)

PR #234, opened 2022-05-22
Make non-`public` static methods actually private

### LINKED ISSUE #232: Standard library cat sample is not compiling
It seems like there is something wrong with the visibility modifiers. None of the functions of the `File` class are set to public, so I suspect that would have to be changed. However, it is weird that the `open_for_reading` doesn't throw an error. Is there a difference between functions that take `this` as an argument and functions that do not?

```jakt
Error: Can't access function `read` from scope None
-----
function main(args: [String]) {
    if args.size() <= 1 {
        eprintln("usage: cat <path>")
        return 1
    }

    let mutable file = File::open_for_reading(args[1])
    let mutable array: [u8] = [0u8]

    while file.read(array) != 0 {
        for idx in 0..array.size() {
            print("{:c}", array[idx])
        }
    }
}

-----
Error: TypecheckError("Can't access function `read` from scope None", Span { file_id: 1, start: 222, end: 231 })
```

linusg:
Closes #232.

### DIFF
--- a/runtime/prelude.jakt
+++ b/runtime/prelude.jakt
@@ -77,11 +77,11 @@ extern struct Error {
 }
 
 extern class File {
-    function open_for_reading(anonymous path: String) throws -> File
-    function open_for_writing(anonymous path: String) throws -> File
+    public function open_for_reading(anonymous path: String) throws -> File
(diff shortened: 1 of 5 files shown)


jntrnr:
Nice :)

PR #235, opened 2022-05-22
Everywhere: Add very basic MVP Variadic Generics

alimpfard:
Note: very early proof of concept

Implemented:
- `Ts...` generic parameter
- `(tuple)...` splat for variables and tuple literals (only if the type is a known tuple type **at function definition time**)
- calls correctly distribute arguments taking splats into account
- variadic functions

Not implemented:
- `Ts...` generic argument
- tuple splat on type vars (can't re-splat)


alimpfard:
This patchset is unrecoverable, will make a new PR later and see that one become useless in an hour too

PR #237, opened 2022-05-22
Everywhere: Stop using 'ty' for type-related variable names

linusg:
From today's video:

> "ty" is used throughout the codebase because type is a reserved word in Rust

How about we don't :^)


nico:
ty!

linusg:
yw!

PR #240, opened 2022-05-22
Tests: Update enum testcase to latest syntax

jntrnr:
This updates the enum test case that tries the different supported enum cases and also adds an output file so that the testcase is actually run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ function main() {
 
 ```jakt
 enum MyOptional<T> {
-    Some: T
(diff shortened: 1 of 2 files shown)


linusg (commented):
I think we'll also need to update this example from the docs:

```
enum Foo {
    StructLikeThingy {
        field_a: i32
        field_b: i32
    }
}
```

jntrnr:
@linusg - good catch. updated

Issue #182, opened 2022-05-21
Add instructions on how to run rustfmt as a pre-commit hook


jdahlin:
This seems to be a common issue in PRs and it would be helpful.

For instance, using [pre-commit](https://pre-commit.com/):

`$ pip install pre-commit`

Adding this to `.pre-commit-config.yaml`
```yaml
-   repo: https://github.com/doublify/pre-commit-rust
    rev: master
    hooks:
    -   id: fmt
```

It also makes it trivial to add cargo-check and clippy.

Thoughts? I can create a PR with the hook config and README.md updates if needed.

linusg:
Sure, a pre-commit config would be appreciated :)

PR #243, opened 2022-05-22
Meta: Add lint-commits script and lintcommits Github Actions workflows

ADKaster:
Copied from serenity with some light modifications :^)

Also add a pre-commit config to use the lint commits script and run cargo fmt.


linusg:
Closes #182

PR #245, opened 2022-05-23
errors: Make error reporting nicer

jntrnr:
This changes the error reporter to:

* print out only nearby lines rather than the whole file
* print line numbers for each line
* print a label on the error

This PR also cleans up the missing quote error message.

Example:

image

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -42,8 +42,8 @@ fn main() -> Result<(), JaktError> {
         }
     }
 
-    if let Some(error) = first_error {
-        Err(error)
+    if first_error.is_some() {
+        std::process::exit(1)
     } else {
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -42,8 +42,8 @@ fn main() -> Result<(), JaktError> {
         }
     }
 
-    if let Some(error) = first_error {
-        Err(error)
+    if first_error.is_some() {
+        std::process::exit(1)
```
linusg, line 46: Just curious, what's the reason for this?

src/main.rs
```diff
@@ -123,24 +123,101 @@ fn parse_arguments() -> JaktArguments {
...
+                );
+            }
+
+            break;
+        } else {
+            line_index += 1
+        }
+    }
+
+    if span.start == span.end && span.start == file_contents.len() {
```
linusg, line 182: Clippy warning here about collapsible this if statement

linusg (commented):
Love this 🎉

src/main.rs
jntrnr: Ahh, this is a cleanup of what we had before. Previous, we panicked if there was an error, which prints debug info to the screen (after the wire)

I believe the more correct solution is non-zero exit code

jntrnr:
[reviewed earlier by linusg (not sure why the comment disappeared...)]

PR #253, opened 2022-05-23
vscode: Add VSCode extension file icon

Jantje19:
Add a logo to Jakt files so you can differentiate them more easily from other file types


linusg (changes requested):
Using the Twemoji bison image with no attributions seems a bit lame

Jantje19:
> Using the Twemoji bison image with no attributions seems a bit lame

I agree, but I was unsure how I would go about that. Do you have a recommendation?

awesomekling:
Let's not add anything that could be interpreted as a logo at this point :)

PR #254, opened 2022-05-23
parser: Parser enters infinite loop when setting enum value to unclosed bracket

### LINKED ISSUE #233: Parser infinite loop when setting enum value to unclosed bracket
For example:
```
enum a { b=[ }
```

cybergsus:
Should close #233.

The parser for `parse_array` didn't bail when `parse_expression` failed and it 
could not recover from the error, resulting in an endless loop.

Likewise, `parse_enum` didn't contain a loop itself but the fact that it 
didn't error out when encountering `Eof` when expecting a `}` makes outer-level 
parsers not aware that `Eof` has been reached, which makes them continue their
loops.

The 'infinite loop' problem has been fixed for those two, although it might 
need a second look at `parse_array` or other parsers for enclosed lists to make 
sure they bail if their last token doesn't close the list. However, the errors 
that get spit out don't match with the actual error that happened, so please 
don't merge this until that problem is solved.

Another problem that occurs to me is that I don't know how to test this,
since checking that an infinite loop doesn't happen is not trivial, or at least
I haven't been able to think of a good solution. Help there will be appreciated.


src/parser.rs
```diff
@@ -3664,6 +3664,12 @@ pub fn parse_array(tokens: &[Token], index: &mut usize) -> (ParsedExpression, Op
             _ => {
                 let (expr, err) =
                     parse_expression(tokens, index, ExpressionKind::ExpressionWithoutAssignment);
+
+                if err.is_some() {
+                    // bail out with no expression and an error
+                    return (ParsedExpression::OptionalNone(tokens[*index].span), err);
```
linusg, line 3670:
```suggestion
                    return (ParsedExpression::Garbage(tokens[*index].span), err);
```

linusg (changes requested):
> Another problem that occurs to me is that I don't know how to test this, since checking that an infinite loop doesn't happen is not trivial, or at least I haven't been able to think of a good solution. Help there will be appreciated.

It is trivial - just add a test for the expected output, and if it never completes, it fails.

`bug:` is not a valid prefix, please use `parser:` and remove the merge commits.

cybergsus:
Ok. How do I remove those merge commits?

linusg:
look up how `git rebase` works, there's plenty of information on that out there

cybergsus:
Parser tests pass, but rest of tests currently fail. I'll be fixing those

cybergsus:
Looks like all the tests pass right now.

cybergsus:
It currently fails since `ref` was not yet implemented when I forked. How do I get those changes here?

Lubrsi:
In case you haven't already, you will have to add the base repo as a remote:
```
git remote add upstream https://github.com/SerenityOS/jakt.git
```
Then, fetch upstream:
```
git fetch upstream
```
Then, whilst on your branch, perform:
```
git rebase upstream/main
```

cybergsus:
Ok, now it runs. I'm now running the whole test suite to know if I messed up somewhere else.

cybergsus:
image
Ok, it works now!

cybergsus:
I've also walked down the parser code and checked for more possible mishaps regarding reaching EOF in loops, since the lexer emits an extra EOF token and therefore `*index < tokens.len()` remains true if EOF is not handled, and the loop continues unless the function bails on a subparser error.

linusg (changes requested):
https://github.com/SerenityOS/jakt/blob/master/CONTRIBUTING.md#code-submission-policy

awesomekling:
This looks like a valuable PR, it just needs a refresh since `ref enum` was renamed to `boxed enum` :)

cybergsus:
Tests for `test_dictionary` continue failing.
]

PR #256, opened 2022-05-23
typechecker: Reject `return` inside a `defer` as an error

### LINKED ISSUE #216: Disallow returns in defer statements
For example:
```
function foo() {
    defer {
        return 2
    }

    return 1
}
```

This will return 1, because it's essentially doing:
```cpp
ScopeGuard blah([] {
    return 2;
});

return 1;
```
The `return 2;` only exits the ScopeGuard lambda and not the top level function.

But this is not intuitive and hard to reason about, and like Agni says:
> but it's probably better to disallow that in the language semantics, because it was mentioned that C++ is not going to be the only backend

AtkinsSJ:
Closes #216

This is my first attempt at Rust so there's probably a smarter way of doing this!
image

### DIFF
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -153,6 +153,7 @@ pub struct Project {
     pub types: Vec<Type>,
 
     pub current_function_index: Option<usize>,
+    pub inside_defer: bool,
 }
 
 impl Project {
@@ -168,6 +169,7 @@ impl Project {
             scopes: vec![project_global_scope],
             types: Vec::new(),
             current_function_index: None,
+            inside_defer: false,
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/typechecker.rs
```diff
@@ -2446,8 +2451,16 @@ pub fn typecheck_statement(
                     .current_function_index
                     .map(|i| project.functions[i].return_type_id),
             );
+            error = error.or(err);
+
+            if project.inside_defer {
+                error = error.or(Some(JaktError::TypecheckError(
+                    "`return` is not allowed inside `defer`".to_string(),
```
linusg, line 2458:
```suggestion
                    "'return' is not allowed inside 'defer'".to_string(),
```

linusg: (I don't think having markdown in compiler error messages is a good choice, people will inevitably copy and paste that into github and accidentally alter the string, possibly breaking search and all that :))

awesomekling: Yeah, I copied what GCC does and used the nifty open/close single quote characters.

PR #259, opened 2022-05-23
Parser: Fix an extern keyword followed by EOF entering an infinite loop

### LINKED ISSUE #231: Parser infinite loop when parsing unfinished extern
For example:
```
extern
```

supex0fan:
Add the missing token index increment in the default cases of the extern parsing match statements.

Resolves #231.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -235,7 +235,7 @@ fn display_message_with_span(
     println!("{}", "-".repeat(width + 3));
 
     while line_index < line_spans.len() {
-        if span.start >= line_spans[line_index].0 && span.start < line_spans[line_index].1 {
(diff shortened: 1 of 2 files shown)


supex0fan:
This seems to be an issue with the new error printing system, it misses errors that occur at the end of a line that isn't the last line in the file. 
@jntrnr The solution seems to be to make the `span.start < line_spans[line_index].1` check on line 145 in main.rs a `<=` check. Then delete the block at 189-197 which checks for errors at the end of the file as that is now done implicitly through the first change. Can you see any issues here?

jntrnr:
@supex0fan worth a try

awesomekling:
This has a conflict

PR #262, opened 2022-05-23
typechecker: Make throw statement expect Error

### LINKED ISSUE #147: Throw expression is not type checked sufficiently
This compiles:

```jakt
function foo() throws -> u32 {
    throw 1
}

function main() {
    let x = foo()
    println("{}", x)
}
```

To:

```cpp
#include "runtime/lib.h"


ErrorOr<u32> foo();

ErrorOr<u32> foo(){
using _JaktCurrentFunctionReturnType = ErrorOr<u32>;
{
        return static_cast<i64>(1LL);    }
}

ErrorOr<int> _jakt_main(Array<String>)
{
{
        const u32 x = TRY(foo());
        outln(String("{}"),x);
    }
    return 0;
}
```

Meaning it will unexpectedly print `1`.

mattisboeckle:
A throw statement requires the type of its
expression to be the Error struct.
Other types of expression will be rejected
by the typechecker.

Add a test to check for this behaviour.

Resolves  #147

### DIFF
--- a/runtime/prelude.jakt
+++ b/runtime/prelude.jakt
@@ -73,7 +73,7 @@ extern struct Range {}
 
 extern struct Error {
     function code(this) -> i32
-    function from_errno(anonymous errno: i32)
+    function from_errno(anonymous errno: i32) -> Error
 }
 
 extern class File {
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -2160,7 +2160,17 @@ pub fn typecheck_statement(
                 typecheck_expression(expr, scope_id, project, safety_mode, None);
             error = error.or(err);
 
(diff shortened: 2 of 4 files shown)


linusg (changes requested):
Please wrap your commit messages at 72 chars and write them as sentences, including punctuation in the description and using imperative mood for the subject. https://github.com/SerenityOS/jakt/blob/main/CONTRIBUTING.md

linusg:
Not sure what you're expecting me to say now - that's neither written in imperative mood, nor wrapped at 72 chars.

mattisboeckle:
Sorry, I didnt't really get it. The title should now (hopefully) be imperative and the lines were automatically wrapped by a tool.

PR #263, opened 2022-05-23
runtime: Open files as binary

nbvdkamp:
This ensures the bytes written are not modified, i.e. `\n` being replaced by `\r\n` on Windows.

Since the current File API only supports reading and writing raw bytes it feels like a sensible default.


nbvdkamp:
Fixed :^)

PR #269, opened 2022-05-23
errors: Make lines 1-based in error output

sppmacd:


src/main.rs
```diff
@@ -210,7 +210,7 @@ fn print_source_line(
 
     let width = format!("{}", largest_line_number).len();
 
-    print!(" {:<width$} | ", line_number);
+    print!(" {:<width$} | ", line_number + 1);
```
linusg, line 213: The correction should be done at the call site, or the argument renamed to line_index

linusg (changes requested):
`main:`, not `errors:`

src/main.rs
jntrnr: This would change the `largest_line_number` calculation also

sppmacd: I renamed the argument to line_index since it is used like this on the call site.

jntrnr: This doesn't fix what we suggest above. Since you're changing the meaning of the value, you need to start the counting with the initial and fix the calculation of `largest_line_number`

mustafaquraish (approved):

Issue #268, opened 2022-05-23
Parser OOB if `ref` is not followed by another token


Lubrsi:
For example:
```
ref
```

PR #270, opened 2022-05-23
Parser: Check that ref is followed by enum keyword

bartlomieju:
Fixes https://github.com/SerenityOS/jakt/issues/268


tests/parser/ref-followed-by-eof-crash.error
```diff
@@ -0,0 +1,5 @@
+Error: expected enum keyword
+----
+ 0 | ref
+     ^- expected enum keyword
```
bartlomieju, line 4: This actually is not correct, the highlight should be after the `ref` word, unfortunately using the next token's span causes error message not to show up:
```
Error: expected enum keyword
----
 0 | ref_
----

```
Seems like a bug in error reporting logic

linusg (commented):
This fails CI

jntrnr:
looks good, thanks!

PR #271, opened 2022-05-23
compiler: Show error if main function is not defined

corecrypt:
This prints an error that is a bit friendlier than the one that you get
from the linker when main is not defined.

### DIFF
--- a/samples/basics/no_main_function.error
+++ b/samples/basics/no_main_function.error
@@ -1,0 +1,1 @@
+the main function needs to be defined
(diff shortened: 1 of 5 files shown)


linusg (commented):
This fails CI

AtkinsSJ:
Now it has a conflict. 😅

awesomekling:
This error only makes sense for projects that have their own `main()` function. It's okay for the linker to be the one who tells you that `main` is missing :)

PR #273, opened 2022-05-23
Everywhere: Typecheck generic functions after instantiating them

alimpfard:
This allows them to be typechecked as needed, which makes it possible to
use operations that aren't valid for all types on a value of a generic
type.

This will make it possible to actually do _something_ with type packs in the future :P


alimpfard:
forgot to make member functions work, stand by...

jntrnr (approved):

tests/typechecker/generic_late_type_check.jakt
```diff
@@ -0,0 +1,21 @@
...
+struct Test {
+    a: i32
+}
+function baz<T>(anonymous x: T) -> i32 {
+    return x.a
+}
+
+function main() {
+    println("{} {} {}", foo(1, 2), bar(1 as! i8), baz(Test(a: 1)))
+}
```
mattisboeckle, line 21: Does this need a newline?

alimpfard: Seems like vscode doesn't do The Right Thing :tm: by default, fixed, thanks for noticing it.

PR #275, opened 2022-05-23
typechecker: Move project_function_len down

### LINKED ISSUE #212: Bogus typecheck error message if unrelated class is present
When compiling this code, an error message is produced:
```
function popcount( bitfield: u32) -> u8 {
    let mutable popcnt : u8 = 0u8;
    return popcnt
}

class Clif {
    function init(mutable this) {}
}
```
Error message:
```Error: TypecheckError("Type mismatch: expected void, but got u8", Span { file_id: 1, start: 367, end: 373 })```

The error goes away if the function `init()` of class `Clif` is commented out, or the Clif class is deleted altogether.

The error is highlighted at the return statement of the popcount() function.

### PR DESCRIPTION
**Pull request by mattisboeckle**
This was causing issues when a function was declared before a class with
a function.

The total amount of functions was saved before the functions of classes
are added to the scope, causing the previously declared functions to
look up their types at the wrong indices.

Resolves #212

### CODE REVIEW

File: tests/typechecker/function_class_function.jakt [hunk 1]
```diff
@@ -0,0 +1,11 @@
...
+    return i
+}
+
+class TestClass {
+    function init(mutable this) {}
+}
+
+function main() {
+    print("{}", test(i: 3));
+}
```
alimpfard, line 11: Please add a trailing newline here

mattisboeckle: Done

PR #278, opened 2022-05-23
Parser: Ensure `parse_function` doesn't enter infinite loops

### LINKED ISSUE #226: Parser infinite loop when parsing templated function with no body
For example:
```
function foo<Ts..>()
```

gdelazzari:
Fixes #226 and other instances where the parser was getting stuck
because an unexpected token was encountered, such as:

```
function foo<Ts..>(a: i32)
```
```
function foo<Ts, ..>(a: i32) 
```
```
function foo<Ts>(a: i32, ..)
```

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1319,6 +1319,8 @@ pub fn parse_function(
                                 "expected '('".to_string(),
                                 tokens[*index].span,
                             )));
+
+                            *index += 1;
                         }
                     }
                 } else {
@@ -1328,6 +1330,10 @@ pub fn parse_function(
                         "incomplete function".to_string(),
                         tokens[*index - 1].span,
                     )));
+                }
+
+                if error.is_some() {
+                    return (ParsedFunction::new(FunctionLinkage::Internal), error);
                 }
 
                 let mut params = Vec::new();
@@ -1402,6 +1408,8 @@ pub fn parse_function(
                                 "expected parameter".to_string(),
                                 tokens[*index].span,
                             )));
+
+                            *index += 1;
                         }
                     }
                 }
@@ -1413,6 +1421,10 @@ pub fn parse_function(
                         "incomplete function".to_string(),
                         tokens[*index - 1].span,
                     )));
+                }
+
+                if error.is_some() {
+                    return (ParsedFunction::new(FunctionLinkage::Internal), error);
                 }
 
                 let mut throws = false;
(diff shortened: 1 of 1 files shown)


src/parser.rs
```diff
@@ -1330,6 +1332,10 @@ pub fn parse_function(
                     )));
                 }
 
+                if error.is_some() {
```
jntrnr, line 1335: can you move these two if blocks up and into the previous else?

```
trace!("ERROR: incomplete function");
error = error.or(Some(JaktError::ParserError(
    "incomplete function".to_string(),
    tokens[*index - 1].span,
)));
return (ParsedFunction::new(FunctionLinkage::Internal), error);
```

src/parser.rs
```diff
@@ -1415,6 +1423,10 @@ pub fn parse_function(
...
+                    return (ParsedFunction::new(FunctionLinkage::Internal), error);
```
jntrnr, line 1427: second one here

src/parser.rs
jntrnr: If that's true, I'd probably rather we copy it so that the error is closer to the error occurring. When we have distance like this, it's possible that a later refactor will break the invariant that we know the kind of error that just occurred. If we move them so they're right beside when the error occurs, this is less likely to happen.

src/parser.rs
```diff
@@ -1319,6 +1319,8 @@ pub fn parse_function(
                                 "expected '('".to_string(),
                                 tokens[*index].span,
                             )));
+
+                            return (ParsedFunction::new(FunctionLinkage::Internal), error);
```
jntrnr, line 1323: We don't want to immediately return unless we're at the end of the input. Otherwise, we should tolerate errors and keep parsing.

gdelazzari: Uhm, okay. So the wanted behavior when `(` is missing is to go on and try parsing function parameters anyway?
What if the previous operation of parsing the generic parameters failed? If that fails, do we want to immediately return at line [1304](https://github.com/SerenityOS/jakt/pull/278/files/6cd18ef93cad3db5ca49339222519d4da6ea88e8#diff-4a04259da480a6b794a2e947e4cc03eff4d1aa9330836f5b91cac68c5398193fR1304)?
If not, I would imagine `index` to be left midway in the generic parameters section, how does it make sense to try and keep going? We should at least advance until we find the closing `>`, shouldn't we?
Also if `(` is missing, we should indeed fire an error, but if we don't want to return early then it would be good to advance tokens until we hopefully find `(`, and then go on from there, right?

Sorry, I'm not getting what the wanted behavior is at the moment.

AtkinsSJ:

PR #279, opened 2022-05-23
typechecker: Show nicer type names in compiler diagnostics

awesomekling:
- Fix a couple of situations where we were showing numeric type IDs in build errors instead of type names.
- Use `T?` shorthand for `Optional<T>` in diagnostics
- Use `weak T?` shorthand for `WeakPtr<T>` in diagnostics


linusg (commented):
> typechecker: Wrap type names in backticks (`) in diagnostics

I'm not convinced that putting markdown into error messages is the best idea, how about just regular `'`?

linusg (approved):
That was a suspiciously quick response :^)

PR #281, opened 2022-05-23
parser: Fix parsing one-line enum declarations 

MacDue:


tests/parser/one-line-enum-no-comma-ending.jakt
```diff
@@ -0,0 +1 @@
+enum Fruit { Apple }
```
jntrnr, line 1: this is valid Jakt - are you thinking we should make it invalid?

MacDue: It was an assumption based off the error that seemed to indicate that variants should end with a comma or newline.
I'll correct that.

MacDue: (Currently enums like `enum Fruit { Apple }` don't parse correctly, so I was not 100% sure on the correct parse)

MacDue:
I've changed the PR to now allow enum declarations like `enum Fruit { Apple }`, and parse them correctly.  

I've also added a new sample that failed to compile before these changes.

jntrnr:
thanks for the fix!

PR #282, opened 2022-05-24
typechecker+codegen: Reorder types in codegen to dependency order

alimpfard:
Had to also fix some parser and codegen boogs to get my test working :P


jntrnr:
yay!

PR #285, opened 2022-05-24
typechecker: Support for basic definite returns

mattisboeckle:
The typechecker will now detect blocks containing if statements with
literal boolean conditions as definitely returning if the respective
branch definitely returns.

We could probably check Variables, Unary- and BinaryOps as well, but that might be overkill.


jntrnr:
thanks!

PR #286, opened 2022-05-24
parser+typechecker: Add (A, B, C) type short-hand for Tuple<A, B, C>

mustafaquraish:
Seems like a nice quality-of-life feature, and was simple enough to add.


src/typechecker.rs
```diff
@@ -5294,6 +5294,23 @@ pub fn typecheck_typename(
...
+                error = error.or(err);
+                checked_types.push(type_id);
+            }
+
+            let tuple_struct_id = project
+                .find_struct_in_scope(0, "Tuple")
+                .expect("internal error: Tuple builtin definition not found");
+
+            let type_id =
+                project.find_or_add_type_id(Type::GenericInstance(tuple_struct_id, checked_types));
```
alimpfard, line 5310: Tuple isn't even generic right now...

mustafaquraish: Ah, well shit. I didn't actually look at the definition of `Tuple` in the prelude, but just tested the implementation with some examples and it seems to do the right thing, which is interesting. I'll leave this PR for now, I don't want to figure out how to add variadic templates 😛

mustafaquraish: I added a `FIXME` for now saying we'll probably have to look at this again when we add support for variadic generics or something else, but I think it works as intended the way it is for now, and seems nice to have in the mean-time,

jntrnr: Yeah for now just reaching for "Tuple" and hoping for the best is the right way. Once we get variadic generics, we'll go back over the uses of Tuple and move them to the new style. For now, your approach should be fine.

jntrnr:
thanks!

PR #287, opened 2022-05-24
parser: Add shorthand for `this.`

awesomekling:
After writing a couple hundred lines of Jakt, the main thing that
stood out was the large amount of `this.` spam throughout methods.

Let's try to mitigate that with a shorthand for refering to member
variables: `^foo` is short for `this.foo`


mustafaquraish:
for whatever my 2 cents are worth, `^` seems like a bit of a weird choice for `this`, makes it feel like it's some weird unary operator when looking at it. Just bikeshedding now, but I feel like perhaps `$` might be nicer? So something like `$foo` or `$.foo`. Maybe even `@`?

Or... maybe I'll just get used to `^`, but this is just my first impression

linusg:
`^` is usually XOR, I'm against this.

awesomekling:
Sure, let's roll with `$`, that was my initial choice as well. :)

Please note that this is an experiment to see if we can improve the code feel. If we end up disliking it, we can always change it.

mustafaquraish:
> Sure, let's roll with `$`, that was my initial choice as well. :)
> 
> Please note that this is an experiment to see if we can improve the code feel. If we end up disliking it, we can always change it.

Looks much nicer to me, but yeah lets see what it feels like!

PS: the test file is still called `this_caret.jakt`

src/parser.rs
```diff
@@ -2423,6 +2423,27 @@ pub fn parse_operand(tokens: &[Token], index: &mut usize) -> (ParsedExpression,
...
+                    *index += 1;
+                    ParsedExpression::IndexedStruct(
+                        Box::new(this_expr),
+                        name.clone(),
+                        tokens[*index].span,
+                    )
+                }
+                _ => {
+                    error = error.or(Some(JaktError::ParserError(
+                        "Missing member name after '^'".to_string(),
```
linusg, line 2440:
```suggestion
                        "Missing member name after '$'".to_string(),
```

linusg (changes requested):

src/parser.rs
linusg: Probably should have a test as well ;^)

README.md
```diff
@@ -178,6 +178,10 @@ let mutable foo = Foo(x: 3)
 foo.set(9)
 ```
 
+### Shorthand for accessing member variables
+
+To reduce repetitive `this.` spam in methods, the shorthand `^foo` expands to `this.foo`.
```
mustafaquraish, line 183: `$foo`

PR #288, opened 2022-05-24
documentation: Add coding-style document

AtkinsSJ:
This is largely based on the SerenityOS CodingStyle.md, with many
sections removed, with examples converted to Rust, and an additional
section on error messages. I barely understand Rust at all so there are
very likely to be some areas I have overlooked, or some issues with my
code examples, but this should be a good start. :^)

### DIFF
--- a/documentation/coding-style.md
+++ b/documentation/coding-style.md
@@ -1,0 +1,162 @@
+# Jakt coding style
+
+For low-level styling (spaces, parentheses, brace placement, etc), all code should be formatted with `cargo fmt`.
+
+This document describes the coding style for the Jakt compiler. All new code should conform to this style.
+
+This is a living document, and will change as we eventually rewrite the compiler in Jakt, and as the Jakt language develops.
+
+
+### Names
+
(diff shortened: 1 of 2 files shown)


documentation/coding-style.md
```diff
@@ -0,0 +1,162 @@
...
+
+```rust
+let ty = calculate_type(foo);
+```
+
+###### Wrong:
+
+```rust
+let type_ = calculate_type(foo);
+```
```
linusg, line 64: Swapped right and wrong examples here

documentation/coding-style.md
```diff
@@ -0,0 +1,162 @@
...
+Type names and code excerpts should be wrapped in ‘ `'LEFT SINGLE QUOTATION MARK' (U+2018)` and ’ `'RIGHT SINGLE QUOTATION MARK' (U+2019)`.
+
+Prefix any internal errors with "Internal error: ".
+
+##### Right:
+
+```
+"Type mismatch: expected ‘String’, but got ‘i64’"
+"Binary arithmetic operation between incompatible types (‘String’ and ‘i64’)"
+"Internal error: expression stack empty"
```
linusg, line 153: Can we have a guide for what follows after the `:` as well? I don't mind if the convention becomes lowercase, but it needs to be written down :)

AtkinsSJ: Sure, I meant to mention that and forgot. :man_facepalming:

PR #289, opened 2022-05-24
typechecker: Fix definite return of if statements that do not have 'true' as condition

nbvdkamp:
This regressed in #285. 
The test didn't catch this because it had true as its condition.


tests/typechecker/return_if_prongs.jakt
```diff
@@ -1,11 +1,11 @@
-function foo() -> i32 {
-    if true {
-        return 123
+function max(a: i32, b: i32) -> i32 {
```
linusg, line 1: Please add a separate test instead of complicating this one

linusg (changes requested):

tests/typechecker/return_if_prongs.jakt
nbvdkamp: Alright, I added another test too to ensure the definite returns still work because those didn't have a test yet.

PR #290, opened 2022-05-24
codegen: Don't rely on ScopeGuard for defer

juniorrantila:
Since the language is built from scratch, why depend on C++ for defer?
This is a first cut and there are probably still some kinks to iron out.

This patch removes the ScopeGuard dependency and instead makes a
built-in version of defer. Currently this works by creating a Vector of
Function at the start of each function. Whenever a defer block is
reached, the block is pushed onto the Vector. Whenever the function
would return it is instead directed to the end of the function, where
the defers are handled before finally returning.

The main benefit of this approach is we're able to call fallible
functions inside of defer blocks. Current semantics are that
deferred blocks run at the end of function scope, not local block
scope. First every defer block is run, and then if any block
failed, the error of the first one to fail is returned.

So for the code

```jakt
function main() {
    defer println("foo")
    return 0

    defer println("bar")
}
```

this would be the output:

```sh
foo
```

And for

```jakt
function fail() thows {
    println("fail")
    throw Error::from_string_literal("boom")
}

function no_fail() {
    println("no fail")
}

function main() {
    println("1")
    defer no_fail()

    println("2")
    defer fail()
}
```

this would be the output:

```sh
1
2
fail
no_fail
Runtime error: boom
```

And for

```jakt
function main() {
    for i in 0..3 {
        println("{}", i);
        defer println("deferred {}", i)
    }
}
```

this would be the output:

```sh
0
1
2
deferred 2
deferred 1
deferred 0
```

Not sure if the last one should be considered a bug or a feature.

### DIFF
--- a/src/codegen.rs
+++ b/src/codegen.rs
@@ -23,6 +23,12 @@ use std::collections::{HashMap, HashSet};
 
 const INDENT_SIZE: usize = 4;
 
+#[derive(Clone, Copy, PartialEq)]
+enum UseRawReturn {
+    No,
+    Yes
+}
+
 pub fn codegen(project: &Project, scope: &Scope) -> String {
     let mut output = String::new();
 
@@ -1126,6 +1132,16 @@ fn codegen_function_in_namespace(
     // Put the return type in scope.
     if is_main {
         output.push_str("using _JaktCurrentFunctionReturnType = ErrorOr<int>;\n");
+        output.push_str(&codegen_indent(INDENT_SIZE));
+        output.push_str("using _JaktCurrentFunctionDeferBlockReturnType = Optional<Error>;\n");
+        output.push_str(&codegen_indent(INDENT_SIZE));
+        output.push_str("auto _jakt_current_function_return_value = Optional<ErrorOr<int>>();\n");
+        output.push_str(&codegen_indent(INDENT_SIZE));
+        output.push_str("#define _JAKT_RETURN(_jakt_expression) \\\n");
+        output.push_str(&codegen_indent(INDENT_SIZE*2));
+        output.push_str("_jakt_current_function_return_value = static_cast<_JaktCurrentFunctionReturnType>(_jakt_expression); \\\n");
+        output.push_str(&codegen_indent(INDENT_SIZE*2));
+        output.push_str("goto _jakt_current_function_return_handler;\n");
     } else {
         if function.return_type_id == UNKNOWN_TYPE_ID {
             panic!("Function type unknown at codegen time in {}", function.name);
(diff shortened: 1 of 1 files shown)


BertalanD:
Will this work for early returns as well?

MacDue:
Does not seem like it:
```
function test(x: i32) {
    defer println("Bye!")
    if (x < 10) {
        return 1;
    } else {
        return 2;
    }
}

function main() {
    test(x: 5);
}
```
```cpp
static void test(const i32 x);

static void test(const i32 x)
{
    using _JaktCurrentFunctionReturnType = void;
    {
        if ((x < static_cast<i32>(10))) {
            return (static_cast<i64>(1LL));
        }
        else {
            return (static_cast<i64>(2LL));
        }
        outln(String("Bye!"));
    }
}

ErrorOr<int> JaktInternal::main(Array<String>)
{
    {
        test(static_cast<i32>(5));
    }
    return 0;
}
```

mustafaquraish:
Like others have pointed out, this does not work for functions with `return`, or with `break`/`continue` in loops. You need to be smarter about doing this, and I don't think the added complexity is worth it for now, since `ScopeGuard` already handles all this properly.

juniorrantila:
As I went for a walk I realized that there's a lot of things this does not handle (derp). In its current form it just puts the deferred code at the end of the scope, meaning it's only executed if control reaches that point. Meaning early returns -> defer never executed. I have some better ideas of how to fix this.

MacDue:
There's ways to fix it, though like Mustafa says it's probably not worth the complexity (in both the codegen + the generated code)

MacDue:
I'm not a fan of this solution. It a bunch of heap allocations to defers that were not there before, and now changes when deferred calls are made. Before a defer would be called at the end of a scope, not the end of the entire function. With this your deferred call list could keep growing if used in a loop. 

If you just want to allow fallible functions to work in `defer` statements and be able to at least propagate the first error that can be done by replacing the ScopeGuard with a slightly different class with a reference to some error:

Here's a sketch (I've not tested this): 

```cpp
template<typename DeferCallback>
struct Defer {
    Defer(DeferCallback cb, Optional<Error>& deferred_error)
        : cb{cb}, deferred_error{deferred_error} {}

    ~Defer() {
        auto result = cb();
        // Set the deferred error if not already set
        if (result.is_error() && !deferred_error_handle.has_value()) {
            deferred_error = result.release_error();
        }
    }

    DeferCallback cb;
    Optional<Error>& deferred_error;
};

ErrorOr<int> my_function() {
    Optional<Error> deferred_error;
    // Wrap everything in a scope so all the Defer's destructors are called before the function ends
    {
        // The functions deferred error (if any)
        // Corresponds to a `defer some_function_that_may_fail()`
        Defer some_defer{[&]{
            return some_function_that_may_fail();
        }, deferred_error} // <--- pass a reference to the deferred error
   }
   if (deferred_error.has_value()) {
        return *deferred_error;
   }
   return /* normal value */
}```

No extra allocations here

alimpfard:
Defer has to be scope based, we saw the nonsense that function-based defers create in Golang, what's so bad about using C++ when we're targeting C++ anyway?

awesomekling:
The `defer` keyword was specifically added to cover the `ScopeGuard` use case as-is in SerenityOS today. I don't think we need to complicate it beyond that, as long as we're transpiling to C++ and have the ability to simply use `ScopeGuard` anyway :^)

PR #292, opened 2022-05-24
lexer: Error out on compiling strings spanning multiple lines

mrkct:
Previously we allowed string spanning multiple line such as

    let x="Hello
    world"

These strings miscompiled since they were copy-pasted as is in C++,
which does not allow strings like these


jntrnr:
thanks!

PR #293, opened 2022-05-24
Everywhere: Implement floating point numbers

mattisboeckle:
Previously `f32` and `f64` were returned as `i64` by the lexer.
Now the typechecker and codegen can handle floats and the lexer returns
them.

The IntegerConstant enum in the typechecker is now a NumberConstant as
it contains floating numbers. The implementation functions have been
renamed accordingly.


mustafaquraish:
Would be nice to at least support floating point literals so we can add better tests

mattisboeckle:
Yeah, will look into it. This will directly conflict with the Tuple Index parsing as it is literally the same

Pythonen:
> This will directly conflict with the Tuple Index parsing as it is literally the same

I was experimenting with this. Would it be sufficient enough if you checked in the parser what type of token became before the `TokenContents::Dot` and acted accordingly? If it was a `TokenContents::Number` etc., shouldn't it then be a float and a float only?

mustafaquraish:
I would imagine that we would handle the dot in floating point literals at the lexer level. Shouldn't be too bad since we can't have variable names start with a numeric digit.

mattisboeckle:
Ok, so this has all the basics. Lexing, typechecking and codegening for float literals. 
What isn't working is stuff like `7.8 + 2` as the typechecker rejects it.

But I would rather fix that in another PR.

jntrnr:
This looks good enough to land and give a try

Issue #118, opened 2022-05-19
Optional chaining


awesomekling:
We should allow JavaScript-style optional chaining: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

```jakt
struct Bar {
    baz: i64
}

struct Foo {
    bar: Bar?
}

function main() {
    let mut foo = Foo(bar: None)
    let baz = foo?.bar?.baz
}
```

Implementation wise, this should become an expression that returns an `Optional<i64>`, since that's the type of the last `?.`
If we short-circuit on any of the `?.` operators, we should return an empty `Optional<i64>`.

brian-gavin:
i'm interested in picking this one up.

wlib:
[not sure where to ask this]

I'm wondering if there is an even more elegant approach syntax for these different types of accesses (normal `a.b`, optional `a?.b`, and infallible `a!.b`).
Typically, a chain of property accesses "degrades in certainty", so writing `a?.b!.c` would be ambiguous. Does it mean that `c` must exist if `b` does?

Also, the most common case would be something like `a!.b?.c?.d?.e?.f?.()`. Would it make more sense to "carry over" the degree of certainty unless overridden, like `a!.b?.c.d.e.f()`.

alimpfard:
Implemented in dc4e06920f30dc9b27f54bca04f5aa4b9d6cc867.

PR #298, opened 2022-05-24
Optional Chaining

lanmonster:
Implements #118

Currently only works for chains of length 1. Future commits to this PR will enable arbitrary length chains.


lanmonster:
> foo?.bar[i++]

I was able to get this working, but the syntax is `foo?.bar?.[i++]`. the result of `foo?.bar` is an optional, so we need to chain on that to do the index.

lanmonster:
@cybergsus is going to take this over.

PR #301, opened 2022-05-24
typechecker: Cache struct IDs for internally used classes

sin-ack:
This makes us avoid repeatedly searching in the project's structs for
internally used structs by caching their IDs the first time they're seen
in the global namespace.

### DIFF
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -163,6 +163,15 @@ pub struct Project {
 
     pub current_function_index: Option<usize>,
     pub inside_defer: bool,
+
+    pub cached_array_struct_id: Option<StructId>,
+    pub cached_dictionary_struct_id: Option<StructId>,
+    pub cached_error_struct_id: Option<StructId>,
+    pub cached_optional_struct_id: Option<StructId>,
+    pub cached_range_struct_id: Option<StructId>,
+    pub cached_set_struct_id: Option<StructId>,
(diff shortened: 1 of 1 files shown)


mustafaquraish:
I'm not really a big fan of the `.unwrap()` everywhere, can we just put those inside the actual helper methods so we're not unwrapping values everywhere else in the code?

jntrnr:
Seconding @mustafaquraish = `.unwrap()` is a bit of a "code smell"

sin-ack:
The intention was something like "if we ever get a chance to actually display this to the user beyond a panic, make it as simple as replacing those `unwrap`s with an `err = err.or(error)`", but I see your point. Will fix.

mustafaquraish:
> The intention was something like "if we ever get a chance to actually display this to the user beyond a panic, make it as simple as replacing those `unwrap`s with an `err = err.or(error)`", but I see your point. Will fix.

I don't think this is a case we necessarily want to handle as a compiler error. If everything has gone right we should never hit this error, otherwise something went wrong in including the prelude. Even previously we weren't treating this as a compiler error but just panicking the program with the `.expect(...)`

sin-ack:
Even gcc reports its internal errors with detailed information about what caused it tot crash, so I don't see why we shouldn't also give some info to the user to aid debugging. :^)

awesomekling:
This has a conflict. Also I agree the `unwrap()` looks silly. :)

Issue #276, opened 2022-05-23
jakt transpiles with incorrect c_char value resulting in clang error


demizer:
I was using c_char wrong but jakt eagerly attempted to compile:

```
$ cat main.jakt 
function main() {
    let str_c: c_char = '\n'
    print("{}", str_c)
}
```
clang error output:

```
$ ./build.sh && ./a.out
output.cpp:9:28: warning: missing terminating ' character [-Winvalid-pp-token]
        const char str_c = '\';
                           ^
output.cpp:9:28: error: expected expression
1 warning and 1 error generated.
```

mustafaquraish:
The issue here seems more to me like the escape characters are not handled. This example works fine with:
```
function main() {
    let str_c: c_char = 'n'
    print("{}", str_c)
}
```

mustafaquraish:
Fixed in 334191e

PR #303, opened 2022-05-24
Everywhere: fix compiler transpiling incorrect c_char value

surrealegg:
CharacterLiteral instead of storing a char it stores a string.
This way we can now store an escape sequence.

Solves #276


jntrnr:
Ahh, and then pass through the escaping? Sounds good. thanks!

PR #304, opened 2022-05-24
typechecker: Do not generate an implicit constructor for extern classes

sin-ack:
Extern classes and structs should not have any generated methods for
them, including constructors. This commit prevents implicit constructors
from being generated for them.

Depends on #302


jntrnr:
good catch

PR #306, opened 2022-05-24
typechecker: Introduce the ability to add a hint to a typechecker error and start making use of it

Lubrsi:
This allows you to show a hint on how to fix an error, for example:
```
Error: type 'String' does not match type 'i64' of previous values in array
----
 2 | function foo() throws {
 3 |     let p = [1, 2, "foo"]
                        ^- type 'String' does not match type 'i64' of previous values in array
 4 | }
----
Hint: array was inferred to store type 'i64' here
----
 2 | function foo() throws {
 3 |     let p = [1, 2, "foo"]
                  ^- array was inferred to store type 'i64' here
 4 | }
----
```
```
Error: redefinition of function foo
----
 3 |  
 4 | function foo() { 
              ^- redefinition of function foo
 5 |      
----
Hint: function foo was first defined here
----
 0 | function foo() { 
              ^- function foo was first defined here
 1 |  
----
```
```
Error: redefinition of variable a
----
 1 |     let a = 1; 
 2 |     let a = 2; 
             ^- redefinition of variable a
 3 | } 
----
Hint: variable a was first defined here
----
 0 | function foo() { 
 1 |     let a = 1; 
             ^- variable a was first defined here
 2 |     let a = 2; 
----
```

### DIFF
--- a/samples/arrays/array_bad.error
+++ b/samples/arrays/array_bad.error
@@ -1,1 +1,1 @@
-does not match type of previous values in vector
+type 'String' does not match type 'i64' of previous values in array
--- a/src/codegen.rs
+++ b/src/codegen.rs
@@ -57,7 +57,7 @@ fn codegen_namespace(project: &Project, scope: &Scope) -> String {
         }
     }
 
-    for (_, enum_id) in &scope.enums {
+    for (_, enum_id, _) in &scope.enums {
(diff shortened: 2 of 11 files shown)


jntrnr:
Niiiice, this will be a nice quality-of-life improvement!

PR #308, opened 2022-05-24
Parser: Fix infinite loop on arrays with multiple expressions

### LINKED ISSUE #300: Parser infinite loop when parsing certain array fill patterns
Examples:
```
function foo() {
    let a = [;]
}
```
```
function foo() {
    let a = [1, 2, 3;1]
}
```

intrnlerr:
This fix makes the parser not go into an infinite loop after
encountering an array item with multiple expressions.

Fixes #300


jntrnr:
Can you add a test case?

intrnlerr:
Sure.

jntrnr:
thanks!

PR #315, opened 2022-05-25
typechecker: Make dictionary index return non-optional value

### LINKED ISSUE #314: Typechecker does not reject unwrapping a non-optional type
This seems to specifically be an issue for dictionaries when using the indexing operator:

```
function main() {
    let a = [1:2, 3:4]
    println("{}", a[1]!)
}
```
Generates the following error from clang:
```
build/test.cpp:11:69: error: member reference base type 'long long' is not a structure or union
        outln(String("{}"),(((a).get(static_cast<i64>(1LL)).value()).value()));
                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
```

mustafaquraish:
Previously we were returning Optional<ValueType> even though the
generated code actually unwrapped the value.

Closes #314

### DIFF
--- a/samples/dictionaries/dict_index_unwrap.error
+++ b/samples/dictionaries/dict_index_unwrap.error
@@ -1,0 +1,1 @@
+Forced unwrap only works on Optional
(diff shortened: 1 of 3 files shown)


jntrnr:
tagging @awesomekling for thoughts

mustafaquraish:
(Just to be clear on the scope of the PR, this is only making the type-checking consistent with the actual generated code)

PR #319, opened 2022-05-25
typechecker: Add a (fairly primitive) match exhaustiveness check

alimpfard:
Now with fun hints and extra cute errors!


jntrnr:
Can you add a test of the error case when the match is incomplete?

alimpfard:
Added _three_ whole tests :P

PR #332, opened 2022-05-25
main: Support targetting g++

MoAlyousef:
Hello

Thanks for this inspired new programming language.

Currently jakt (by default) attempts to compile the generated C++ source using a C++ compiler, however it makes several assumptions:
- The compiler, if not passed thru arguments, is called `clang++`. It could be `clang++-<some-version>` or even g++ or `g++-<some-version>`.
- The extra arguments passed to the compiler are clang specific.

To address the issue:
- The CXX env variable is queried for the system compiler, which could be g++.
- Depending on whether the compiler name contains clang or not, the arguments passed to the compiler can be clang-specific or g++-specific.

The conditional can later accommodate to allow support for targetting the msvc cl.exe compiler, but it wasn't added since currently targetting cl.exe fails due to some unrelated include issues:
- Using compiler builtins: __builtin_fabs , __INT64_TYPE__ ..etc.
- search path conflicts due to window's filesystem insensitivity: runtime/AK/String.h will be used for C's string.h which is included in `<cstring>` etc.


src/main.rs
```diff
@@ -47,26 +47,40 @@ fn main() -> Result<(), JaktError> {
...
+
+                    let is_clang = cxx_compiler.to_string_lossy().contains("clang");
+
+                    let color_diags = if is_clang {
+                        "-fcolor-diagnostics"
+                    } else {
+                        "-fdiagnostics-color"
+                    };
+
+                    let def_literals = if is_clang {
```
MacDue, line 64: again `def_literals` is an unnecessary abbreviation, something like `disable_literal_warnings` would be fine.

MacDue (changes requested):
I think this looks good, just a few minor things:

src/main.rs
```diff
@@ -47,26 +47,40 @@ fn main() -> Result<(), JaktError> {
...
+                        "-Wno-user-defined-literals"
+                    } else {
+                        "-Wno-literal-suffix"
+                    };
+
+                    let clang_output = Command::new(cxx_compiler)
```
MacDue, line 70: Also this variable should now be `cxx_compiler_output` since it's not just clang.

MacDue:
Squash your `main: Apply suggested change` into `main: Support targetting g++` (fixups should not be their own commit)

src/main.rs
```diff
@@ -47,27 +47,41 @@ fn main() -> Result<(), JaktError> {
...
+
+                    let is_clang = cxx_compiler.to_string_lossy().contains("clang");
+
+                    let enable_color_diagnostics = if is_clang {
+                        "-fcolor-diagnostics"
+                    } else {
+                        "-fdiagnostics-color"
+                    };
+
+                    let disable_literal_warnings  = if is_clang {
```
MacDue, line 64: Tiny thing, run the rust code formatter over this (noticed this extra space here, which will likely fail the CI linting)

src/main.rs
```diff
@@ -47,27 +47,41 @@ fn main() -> Result<(), JaktError> {
...
-                        &input_cpp,
-                        "-o",
-                        &output_executable,
-                    ])
-                    .output()?;
-                    io::stderr().write_all(&clang_output.stderr)?;
+                    let cxx_compiler = if let Some(ref path) = arguments.cxx_compiler_path {
+                        path.to_owned()
+                    } else {
+                        PathBuf::from(env!("CXX", "No supported C++ compiler was found!"))
```
MacDue, line 53: Another thing, will this error with `"No supported C++ compiler was found!"` if `CXX` is not set and `cxx_compiler_path` is not provided? 

The help states the compiler defaults to clang++ https://github.com/SerenityOS/jakt/blob/ecc5d6755f1b46ed2c1b7ca82cadc4d5d989e1c1/src/main.rs#L128 

I think this probably should stay the case (so you could first check `cxx_compiler_path` then the `CXX` env, otherwise default to `clang++`).

MacDue: You could maybe check `clang++` exists and error otherwise?

MoAlyousef: I've changed this to use std::env::var("CXX").unwrap_or("clang++")

MacDue (approved):

MacDue:
@MoAlyousef do you want a hand re-basing this?

MoAlyousef:
Sure

MacDue:
I've done the re-basing here: https://github.com/MacDue/jakt/tree/gcc_support (on my gcc_support branch).
You can fork my repo and make a pull request from that branch, or I can make the pull request (you'll still get the credit for the commit either way)

MacDue:
Also I put your extensive PR description in the commit body, so it's not lost after merging.

MoAlyousef:
Yeah I prefer if you can make the pull request

MacDue:
Done: https://github.com/SerenityOS/jakt/pull/357

PR #334, opened 2022-05-25
parser: Add explicit error for fat arrow after return type specification

### LINKED ISSUE #331: Confusing error when defining => functions with a -> type
Functions defined using `=>` have their return type inferred automatically, and are not allowed to have it stated explicitly with `-> Foo`. However, the error you get is not at all clear. With this example code:

```jakt
class Limited {
    public function get_secret() -> String => "Shhhh! Don't tell anyone!"
}

function main() {
    Limited::get_secret();
}
```

...we get this error:

```
Error: expected field
----
 6 | } 
----
```

A better error would tell you that the method definition is invalid because of the `-> String`, and point to it.

EDIT: Alternatively, maybe explicit return types should be allowed and override the inferred one? Either way, this is a confusing error.

cybergsus:
Added an explicit match to catch situations in functions where the
return type for the function is specified and a fat arrow is matched
just after, to indicate the user that a fat arrow wasn't expected
in that context instead of erroring later on a different context.

Should close #331.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1564,6 +1564,26 @@ pub fn parse_function(
                     );
                 }
 
+                // Check for a fat arrow to tell the user that fat arrows aren't
+                // compatible with specifying the type
+                // TODO: add a hint explaining that fat arrows have their type deducted
+                if return_type != ParsedType::Empty
+                    && matches!(
+                        tokens[*index],
+                        Token {
+                            contents: TokenContents::FatArrow,
+                            ..
+                        }
+                    )
+                {
+                    trace!("ERROR: Fat arrow after explicit return type");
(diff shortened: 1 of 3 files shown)


mustafaquraish:
Just curious, why should explicit return types not be allowed for arrow functions? Feels weird to error out as opposed to just parsing and using it as a type hint for the decision.

cybergsus:
I  don't know the reasons, maybe it was initially like that to simplify the code. I don't see why this is an issue as well, but I think it's better if we ask in the [discord](https://discord.gg/serenityos)

linusg:
https://github.com/SerenityOS/jakt/issues/331#issuecomment-1137630544

Issue #218, opened 2022-05-22
Parser infinite loop when seeing unsupported operator in set while parsing return


Lubrsi:
This causes an infinite loop in the parser:
```jakt
function foo() {
    return {1:1}
}
```
Removing `return` makes it error out correctly:
```
-----
function foo() {
    {1:1}
}
 
-----
Error: ParserError("unsupported expression", Span { file_id: 1, start: 23, end: 24 
```

holychowders:
Looking into this.

lanmonster:
fixed as of 9012330

PR #335, opened 2022-05-25
Parser: Fix infinite loop in `parse_set()`

holychowders:
```
function main() {
    return {1:1}
}
```
Make code above error out instead of infinitely looping.


holychowders:
Resolves #218.

mustafaquraish:
This is not correct, we cannot just increment the index _before_ we parse the expression, we would be skipping the first token. This example does not work:

```
function main() {
    let x = {1}
    println("{}", x.size())
}
```

PR #336, opened 2022-05-25
Runtime: Supress Clang warning on Array resize call

### LINKED ISSUE #320: Clang warning on Array resize call
Here's the code:
```
function main() {
    let mutable x = [0]
    let y = x.pop()!
}
```

Here's the clang warning:
```
runtime/Builtins/Array.h:231:9: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
        resize(size() - 1);
        ^~~~~~ ~~~~~~~~~~
build/test.cpp:20:29: note: in instantiation of member function 'JaktInternal::Array<long long>::pop' requested here
        const i64 y = (((x).pop()).value());
```

surrealegg:
By casting void we tell the compiler that we are willingly discarding the value.

fixes #320

### DIFF
--- a/runtime/AK/StringBuilder.cpp
+++ b/runtime/AK/StringBuilder.cpp
@@ -77,7 +77,7 @@ StringView StringBuilder::string_view() const
 
 void StringBuilder::clear()
 {
-    m_buffer.resize(0);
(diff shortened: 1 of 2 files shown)


jntrnr:
thanks :)

PR #339, opened 2022-05-25
editor: Add single quote highlight for Vim

CrystalSage:
Add syntax highlighting for single quote character/byte literals. Highlight is similar to `String` type with double quotes.


editors/vim/syntax/jakt.vim
```diff
@@ -79,6 +79,7 @@ syntax region jaktCommentLine start="//" end="$"
 
 
 syntax region jaktString matchgroup=jaktStringDelimiter start=+"+ skip=+\\\\\|\\"+ end=+"+ oneline contains=jaktEscape
+syntax region jaktChar matchgroup=jaktCharDelimiter start=+'+ end=+'+ oneline
```
jntrnr, line 82: can this contain escapes also?

CrystalSage: If we need it, sure! I'll add another commit.

jntrnr:
@CrystalSage - thanks - can you squash the commits? then I think we're ready

CrystalSage:
@jntrnr I messed up while squashing. Can I open another PR? Getting confused.

PR #340, opened 2022-05-25
main: Silence "deprecated functions" warning on Windows

kleinesfilmroellchen:
This warning can happen for functions like fopen which Windows has deprecated but others not. Specifically, it will happen if clang uses the MSVC runtime and/or linker.


nico:
Or add that with a -D flag. Seema better than completely disabling this warning.

MacDue:
Yeah adding with the -D flag sounds the best

PR #342, opened 2022-05-25
parser+typechecker: Accept fat arrow with explicit return type

### LINKED ISSUE #331: Confusing error when defining => functions with a -> type

cybergsus:
Now Jakt accepts specifying a return type explicitly for functions using a fat arrow to declare their block as a single expression with
a return, and makes sure that the type of the expression matches with the explicit type:

```jakt
class S {
  function get_secret() -> i32 => "World"
}
```

would result in a `Type mismatch` error with `i32` and `String`, but

```jakt
function secret() -> i32 => 42
```

won't throw any errors.

Works for both methods and standalone functions.
Will close #331.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -157,6 +157,7 @@ pub struct ParsedFunction {
     pub block: ParsedBlock,
     pub throws: bool,
     pub return_type: ParsedType,
+    pub return_type_span: Option<Span>,
     pub linkage: FunctionLinkage,
     pub must_instantiate: bool,
 }
@@ -196,6 +197,7 @@ impl ParsedFunction {
             block: ParsedBlock::new(),
             throws: false,
             return_type: ParsedType::Empty,
+            return_type_span: None,
             linkage,
             must_instantiate: false,
         }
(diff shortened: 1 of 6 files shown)


tests/typechecker/function-return-type-with-fat-arrow.jakt
```diff
@@ -0,0 +1,3 @@
+class S {
+  public function get_value() -> i32 => "Hello"
```
linusg, line 2: 4 spaces please

src/parser.rs
```diff
@@ -1495,48 +1497,43 @@ pub fn parse_function(
...
+                    Some(Span {
+                        file_id,
+                        start,
+                        end: tokens[*index - 1].span.end,
+                    })
+                } else {
+                    None
+                };
+
+                // accept an (optional)  fat arrow
```
linusg, line 1524: double space in comment here

linusg (changes requested):

src/parser.rs
cybergsus: That code goes through automatic formatter. I can't enforce that manually.

cybergsus: Nevermind, I could enforce that manually.

alimpfard (commented):
Just a (repetitive) nitpick

src/parser.rs
```diff
@@ -1495,48 +1497,43 @@ pub fn parse_function(
...
+                    Some(Span {
+                        file_id,
+                        start,
+                        end: tokens[*index - 1].span.end,
+                    })
+                } else {
+                    None
+                };
+
+                //  Accept an (optional)  fat arrow
```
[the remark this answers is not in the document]
cybergsus: oh I thought the space was needed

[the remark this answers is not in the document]
AtkinsSJ: Linus was talking about the double space between "(optional)" and "fat", which shouldn't be there. :^)

awesomekling:
Seems like we could also have a test where the parse succeeds instead of fails :)

cybergsus:
I'll create a `.out` file as well with a sample output from a sample program that uses this syntax then

cybergsus:
Thanks! <3

PR #344, opened 2022-05-25
Remove `as truncated` and `as saturated` casts in favor of library functions

### LINKED ISSUE #220: Remove `as truncated` and `as saturated` from the syntax
I don't like this:
1. These two operators, by their semantics, only work on integer types, which treats them specially. Normal `as` works on all types and might have user-defined behavior in the future.
2. We're bloating the syntax with features that are non-essential and not super common. Sure, they are extremely good and helpful features, but does that warrant their own syntax?

My suggestion would be to move them to generic prelude functions `as_truncated<T>(anonymous a: T) throws T` and `as_saturated...`. Then they create almost no additional visual noise while reducing compiler complexity.

CC @awesomekling , you wrote the original section on casts in the README so I figured you care a whole bunch just as I do :^)

kleinesfilmroellchen:
See #220 for the rationale.

Fixes #220


jntrnr:
pinging @awesomekling for this one

linusg:
@jntrnr this was sanctioned :)

image

linusg (approved):

PR #346, opened 2022-05-25
Typechecker: Fix not erroring when checking Struct and GenericInstance

### LINKED ISSUE #326: Missing compiler error
The below compiles without error  in the Jakt stage, but should error out due to an (empty) array being passed to a parameter that is expecting a class type.

```

class B {
}


class A {
    b: B
}

function main() {
    let mutable a = A(b: [])
}
```

intrnlerr:
Fixes #326


jntrnr:
thanks :)

PR #347, opened 2022-05-26
Add type formatters for pretty-printing Jakt objects

awesomekling:
This PR adds code generation of `AK::Formatter` specializations for every emitted Jakt `class`/`struct`/`enum`.
This means you can now pretty-print Jakt objects by doing:
```
class Size {
    width: i64
    height: i64
}

class Point {
    x: i64
    y: i64
}

class Rect {
    location: Point
    size: Size
}

function main() {
    let rect = Rect(location: Point(x: 0, y: 0), size: Size(width: 15, height: 30))
    println("{}", rect)
}
```

The program above prints this:
```
Rect(location: Point(x: 0, y: 0), size: Size(width: 15, height: 30))
```

### DIFF
--- a/runtime/AK/NonnullRefPtr.h
+++ b/runtime/AK/NonnullRefPtr.h
@@ -231,14 +231,6 @@ inline NonnullRefPtr<T> adopt_ref(T& object)
     return NonnullRefPtr<T>(NonnullRefPtr<T>::Adopt, object);
 }
 
-template<typename T>
-struct Formatter<NonnullRefPtr<T>> : Formatter<const T*> {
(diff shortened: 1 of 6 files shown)


jntrnr:
I tried this one, though unfortunately wasn't able to get it to work:
```
enum Foo {
    Bar
    Baz(i64)
}

function main() {
    let foo = Foo::Baz(100)
    println("{}", foo)
}
```

awesomekling:
> enum Foo {
>     Bar
>     Baz(i64)
> }
> 
> function main() {
>     let foo = Foo::Baz(100)
>     println("{}", foo)
> }

Aha yes, that case was not covered. I've added support + test for it. :^)

PR #350, opened 2022-05-26
editor: Add single quote highlight for Vim

CrystalSage:
This is a duplicate of my earlier PR because I messed up there. Can @jntrnr have a look? :)

Add syntax highlighting for single quote character/byte literals. Highlight is similar to String type with double quotes.


jntrnr:
thanks!

PR #353, opened 2022-05-26
Parser: Allow an arbitrary number of newlines before code blocks

### LINKED ISSUE #330: Error with opening curly brace on newline
The following code doesn't compile
```
function main()
{
  print("Hello!")
}
```
with the compiler complaining
```
Error: expected complete block
-----
-----
```

I guess the parser expects `{` to be the next token after the function declaration, without ignoring newlines.

Is this intended as a way to enforce code style like Golang does? If not, maybe the error message could be better?

supex0fan:
Previously, the parser would throw an error if it encountered a newline
before the opening curly brace. This commit allows the parser to handle
any number of newlines before an opening curly brace.

Resolves #330.


awesomekling:
When fixing stuff, please add tests that fail before the change (but succeed after) :^)

PR #365, opened 2022-05-26
typechecker: Type check none coalescing (??) operations

awesomekling:
Verify that the LHS is Optional<T> and RHS is T.


mustafaquraish:
I'm not sure if it would actually be useful, but I would imagine it might also be nice to be able to do `??` on a weak reference? Maybe I'm just thinking that because we also use `None` for those

PR #367, opened 2022-05-26
Codegen: Generate debug descriptions for optional class members

lanmonster:
Previously, if a class definition had any optional members, clang would error out because there is no formatter for optional.


src/codegen.rs
```diff
@@ -978,26 +979,43 @@ fn codegen_debug_description_getter(structure: &CheckedStruct, project: &Project
...
 
-        output.push_str("builder.appendff(\"");
-        output.push_str(&field.name);
-        output.push_str(": {}");
-        if i != structure.fields.len() - 1 {
-            output.push_str(", ");
-        }
+        match project.types[field.type_id] {
+            Type::GenericInstance(struct_id, _) => {
+                if struct_id == project.get_optional_struct_id(Span::new(0, 0, 0)) {
```
mustafaquraish, line 984: Instead of passing `Span::new(0, 0, 0)` I think you should just pass along the span from `ParsedStruct` to `CheckedStruct`, and then just use that here.

lanmonster: fixed

mustafaquraish:
It seems you've messed something up with rebasing on the main branch. Please fix the commits associated with the PR, and merge the 2 commits you've made into one. Other than that, the change itself looks good to me.

(If you need some help with git, feel free to hit me up on discord)

alimpfard (commented):
Instead of doing this, just specialise AK::Formatter for `Optional<T>`.

awesomekling:
> Instead of doing this, just specialise AK::Formatter for `Optional<T>`.

Fixed by #389

PR #368, opened 2022-05-26
editors/vscode: Specify global vsce installation needed

u9g:
Specifies that in order to call `vsce` from terminal on windows, you have to install it globally. If you don't install it globally, you can only reference it via package.json scripts, which isn't done currently.


jntrnr:
thanks!

PR #372, opened 2022-05-26
vscode: Add placeholder VSCode extension icon

MacDue:
Inspired by #253 

This adds a placeholder icon for the VSCode extension, until an actual icon is created.

Not sure how people feel about this icon, but I think it looks better than nothing ¯\\\_(ツ)_/¯

image

(icon is CC-BY 4.0, like everything else, I made this without reusing other assets)


awesomekling:
Let's not add anything that could be interpreted as a logo at this stage :)

PR #376, opened 2022-05-27
codegen+lexer+typechecker+parser: Add NoneCoalescingAssign

u9g:
Adds `??=` syntax, since we already have a `??` operator, and most of the other operators have an equivalent for assignment too


tests/typechecker/none_coalescing_assign_to_immutable.err
```diff
@@ -0,0 +1,7 @@
+Error: Assignment between incompatible types (‘String?’ and ‘String’)
```
AtkinsSJ, line 1: Seems to me that the error here should be that you're trying to modify an immutable variable, not that the types don't match. (Because the types *do* match!)

This isn't because of your changes though as far as I can tell, just an observation.

src/codegen.rs
```diff
@@ -2818,6 +2818,12 @@ fn codegen_expr(indent: usize, expr: &CheckedExpression, project: &Project) -> S
                     output.push_str(&codegen_expr(indent, rhs, project));
                     output.push_str("; })");
                 }
+                BinaryOperator::NoneCoalescingAssign => {
+                    output.push_str(&codegen_expr(indent, lhs, project));
+                    output.push_str(".emplace(");
+                    output.push_str(&codegen_expr(indent, rhs, project));
```
awesomekling, line 2824: This looks like we will evaluate the RHS even if the LHS is `None`.
We'll want to short-circuit in that case, so that we don't get any side effects when assignment doesn't happen.

awesomekling (changes requested):

src/codegen.rs
u9g: I think you meant that if LHS has a value then we should short circuit, so I did that.

PR #378, opened 2022-05-27
Traits!

kleinesfilmroellchen:
yes, you read that right. This PR will not include generic traits or constraint stuff because :yaktangle:

- [x] parse traits and `implements`
- [x] typecheck traits
- [x] typecheck `implements`

Take a look at the new README or the examples to learn more.


src/typechecker.rs
```diff
@@ -2156,6 +2189,55 @@ fn typecheck_struct(
...
+        ));
+        if error.is_some() {
+            return error;
+        }
+        let function_id = project
+            .find_function_in_scope(trait_scope_id, &function.name)
+            .expect("internal error: function not added to scope");
+        let checked_function = &mut project.functions[function_id];
+        if checked_function.is_static() {
+            // Static functions are not allowed on traits right now, just use namespaces.
```
alimpfard, line 2224: Any reason why? I removed the check and it seemed to be working fine

kleinesfilmroellchen: The static function is still on the implementing type, that feels weird to me

alimpfard: the trait also expresses requirements on a type, I don't understand why this shouldn't be allowed:
```
trait ArgumentParser {
    function parse(anonymous args: SomeReferenceIterable) throws -> Self
}

struct Foo {
    implements(ArgumentParser) function parse(anonymous args: SomeReferenceIterable) throws -> Foo {
        ...
    }
}
```

PR #388, opened 2022-05-27
typechecker: Ensure function block typecheck runs in all cases

### LINKED ISSUE #380: typechecker: "Condition must be a boolean expression" when branching on return value from fat arrow function
This program fails to compile:
```
function moop() => true

struct Class {
    function foop(this) {
        if moop() {
        }
    }
}

function main() {
}
```
Error:
```
Error: Condition must be a boolean expression
----- sup.jakt:5:12
 3  |     function foop(this) { 
 4  |         if moop() { 
                 ^- Condition must be a boolean expression
 5  |         } 
-----
```

cybergsus:
Running typecheck on function blocks independently of generics enables typechecking
processes that come before said function is properly typechecked to know the return
type of the function, even if it was declared with a fat arrow and without specifying
an explicit return type.

Fixes #380.


jntrnr:
@cybergsus - we hit some subtle issues with the scope with this PR. I reverted it here: https://github.com/SerenityOS/jakt/pull/404 and added a test case to show the issue. That said, I think it's an important issue you found to fix, and I hope you'll give it another go.

cybergsus:
hmm, maybe I should have paid more attention before trying to resolve typevars in all cases. Saw the test, will check the code again asap

PR #392, opened 2022-05-27
Parser:  Mark main as throws automatically

### LINKED ISSUE #360: main function is not marked as throws by default
When I disallow storage types from being created inside non-throwing functions, it also disallows creating them in `main`, as we don't currently see `main` as throwing. For example, with this patch, you cannot create arrays in `main` without also marking it as `throws`:
```patch
From deddce99f344f07dedb5befac279b52aef7236e7 Mon Sep 17 00:00:00 2001
From: Luke Wilde <lukew@serenityos.org>
Date: Thu, 26 May 2022 16:28:56 +0100
Subject: [PATCH] disallow creating arrays in non-throwing functions

---
 src/typechecker.rs | 26 ++++++++++++++++++++++++++
 1 file changed, 26 insertions(+)

diff --git a/src/typechecker.rs b/src/typechecker.rs
index a249964..6c85ded 100644
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -3033,6 +3033,24 @@ pub fn try_promote_constant_expr_to_type(
     None
 }
 
+fn return_error_if_current_function_is_non_throwing(project: &Project, error_message: String, error_span: &Span) -> Option<JaktError> {
+    if let Some(current_function_index) = project.current_function_index {
+        let current_function = &project.functions[current_function_index];
+        if !current_function.throws {
+            Some(JaktError::TypecheckErrorWithHint(
+                error_message,
+                *error_span,
+                format!("consider marking function {} as 'throws'", current_function.name),
+                current_function.parsed_function.as_ref().expect("Typechecking inside non-parsed function").name_span,
+            ))
+        } else {
+            None
+        }
+    } else {
+        None
+    }
+}
+
 pub fn typecheck_match_body(
     body: &MatchBody,
     scope_id: ScopeId,
@@ -3456,6 +3474,14 @@ pub fn typecheck_expression(
             }
         }
         ParsedExpression::Array(vec, fill_size_expr, span) => {
+            error = error.or(
+                return_error_if_current_function_is_non_throwing(
+                    project,
+                    "arrays can only be created in throwing functions, because creating an array can throw an allocation failure".to_string(),
+                    span,
+                )
+            );
+
             let mut inner_type_id = UNKNOWN_TYPE_ID;
             let mut inner_type_span: Option<Span> = None;
             let mut output = Vec::new();
-- 
2.25.1


```

intrnlerr:
Fixes #360

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1615,6 +1615,10 @@ pub fn parse_function(
                     )));
                 }
 
+                if function_name == "main" {
+                    throws = true
+                }
+
                 if let FunctionLinkage::External = linkage {
                     return (
                         ParsedFunction {


Lubrsi:
The commit linter message is misleading, there's two spaces between the `:` and the starting `M`:
```
Parser:  Mark main as throws automatically (5773295f685e203c483c8b8dbd19be3066be59f6):
  First word of commit after the subsystem is not capitalized
```

intrnlerr:
Thank you, it's fixed now.

Issue #121, opened 2022-05-19
Imports/includes


linusg:
Being able to split code across files is crucial for writing maintainable and reusable software. No strong feelings regarding syntax and actual implementation.

awesomekling:
Let's prioritize finding an approach that optimizes for build time.

bbkane:
Also consider IDEs as clients - see https://rust-analyzer.github.io/blog/2020/07/20/three-architectures-for-responsive-ide.html

bbkane:
@lanmonster , could you drop a docs link about how included/imports work now or link to a commit?

PR #396, opened 2022-05-28
parser: Add very basic support for importing files

mustafaquraish:
This is currently very limited, but is a step towards having a much more robust system. For now:

1. We can only import files in the same directory
2. No recursive imports are handled
3. No aliasing for imports
4. Cannot import only specific functions / etc from a file

Some progress towards #121


jntrnr:
I like it. This is definitely enough to get us going.

PR #397, opened 2022-05-28
typechecker: Detect missing throws annotations

### LINKED ISSUE #153: Point out missing `throws` for functions that can fail

flostellbrink:
Keeps track of which scopes will handle throws. I.e. scopes in a try block and scopes in functions that are marked as throws.

If a throw happens in a scope that doesn't handle it, we show this message:

```
Throw statement needs to be in a try statement or a function marked as throws
```

If a function marked as throws is called in a scope that doesn't handle it, we have this:

```
Call to function that may throw needs to be in a try statement or a function marked as throws
```

~~Happy to undo the test changes when #392 get merged.~~

Fixes #153

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -96,7 +96,7 @@ impl Compiler {
             return Err(err);
         }
 
-        let scope = Scope::new(Some(0));
+        let scope = Scope::new(Some(0), false);
         project.scopes.push(scope);
 
         let file_scope_id = project.scopes.len() - 1;
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -2669,6 +2669,15 @@ pub fn typecheck_statement(
                 )));
             }
 
+            let scope = &project.scopes[scope_id];
+            if !scope.throws {
+                error = error.or(Some(JaktError::TypecheckError(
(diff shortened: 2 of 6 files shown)


flostellbrink:
Rebased and undid the changes to existing tests, as those are no longer necessary with #392 merged

flostellbrink:
Rebased and undid the changes to existing tests, as those are no longer necessary with #392 merged

jntrnr:
Looks like good progress. One thing that we may either want to be part of this PR or a follow-up PR: if you work with creating strings, recursive enums, or any of the other growable containers, we can also throw in the function. We'll need to detect this and give them an error if they don't mark the function as throws (otherwise codegen doesn't work properly)

flostellbrink:
Yes you're right, there are definitely still blindspots.
I'm sure they'll get addressed as we go.

PR #398, opened 2022-05-28
typechecker: Add "did you mean" hints

juniorrantila:
This patch adds "did you mean" hints for misspelled namespaces,
structs, classes, functions, enums and variables.

Before, we'd get an error stating the variable, type, etc does not
exist. Now it checks for a similarly named item and hints at the
misspelling.

```
Error: Unknown type
----- test.jakt:5:19
 3  |
 4  | function foo() -> NispeeledTypaad {
                        ^- Unknown type
 5  |     return MisspelledType()
-----
Hint: Did you mean 'MisspelledType'?
----- test.jakt:5:19
 3  |
 4  | function foo() -> NispeeledTypaad {
                        ^- Did you mean 'MisspelledType'?
 5  |     return MisspelledType()
-----
```

### DIFF
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -153,6 +153,46 @@ pub fn can_fit_number(type_id: TypeId, value: &NumberConstant) -> bool {
     }
 }
 
+// Uses optimal string alignment distance algorithm
+// (https://en.wikipedia.org/wiki/Damerau-Levenshtein_distance).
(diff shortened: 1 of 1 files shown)


awesomekling:
This looks really repetitive, can we make this generic so we don't have the same logic copy-pasted so many times? :)

PR #400, opened 2022-05-28
runtime: Add back AK::String::operator+

mustafaquraish:
This was removed from the String refactor. Currently, this fails
if String allocation is not successful, but this should be changed
once we can tell Jakt that `String + String` is fallible and get it
to wrap the operation in TRY()

### DIFF
--- a/runtime/AK/String.cpp
+++ b/runtime/AK/String.cpp
@@ -179,6 +179,14 @@ bool String::operator==(char const* c_string) const
     return view() == c_string;
 }
 
+String String::operator+(String const& other) const
+{
+    StringBuilder builder;
+    MUST(builder.try_append(*this));
+    MUST(builder.try_append(other));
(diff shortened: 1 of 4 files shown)


mustafaquraish:
(I'm not sure if this was removed on purpose, if so please feel free to close the PR)

awesomekling:
Both `a + b` and `a += b` fell out of the language because we couldn't propagate the `ErrorOr<String>` without additional smarts.
I'm not against the feature, but if we want to have it we should handle the error properly :)
(And also support `+=`)

mustafaquraish:
> Both `a + b` and `a += b` fell out of the language because we couldn't propagate the `ErrorOr<String>` without additional smarts. I'm not against the feature, but if we want to have it we should handle the error properly :) (And also support `+=`)

Sounds good to me! Closing for now.

PR #404, opened 2022-05-28
typechecker+tests: Revert "ensure function block typecheck runs in all cases" and add test

jntrnr:
Reverts SerenityOS/jakt#388

I've also added a test to check for proper scope id creation order in the presence of generic functions.


jntrnr:
(reviewed by @alimpfard)

PR #405, opened 2022-05-28
parser+typerchecker+codegen+tests: Tuple destruction

AndrewBregger:
Allow for tuple destruction during variable declaration.


creator1creeper1:
I think you mean destructuring and not destruction.

samples/tuples/bad_type_destruct.error
```diff
@@ -0,0 +1 @@
+cannot destruct i64 into tuple
```
mustafaquraish, line 1: `destruct` feels a bit weird, how do you feel about using `unpack` in the scope of this PR?

src/codegen.rs
```diff
@@ -3278,3 +3302,11 @@ fn postorder_traversal(
 
     output.push(type_id);
 }
+
+fn gen_temp_varaible() -> String {
+    static COUNTER: AtomicUsize = AtomicUsize::new(0);
+    format!(
+        "__temp{}",
+        COUNTER.fetch_add(1, std::sync::atomic::Ordering::SeqCst)
```
mustafaquraish, line 3310: I'm no expert in rust, but this feels rather unnecessary. There has to be an easier way to just get incrementing numbers, right? We don't really care about being atomic here.

src/parser.rs
```diff
@@ -12,10 +12,10 @@ use crate::lexer::{Span, Token, TokenContents};
 use crate::typechecker::NumericConstant;
 
 macro_rules! trace {
-    ($x: expr) => {
+    ($($x:tt)*) => {
         #[cfg(feature = "trace")]
         {
-            println!("{}", $x)
+            println!($($x)*)
```
mustafaquraish, line 18: The first argument to `println!()` should be a format string, like `clippy` is pointing out. Is there a reason to change this from what it was earlier? you could already do `trace!(format!(...))`

src/parser.rs
```diff
@@ -549,6 +579,48 @@ fn parse_import(
...
+        match &token.contents {
+            TokenContents::Name(name) => {
+                *index += 1;
+                (name.to_string(), token.span, None)
+            }
+            _ => (
+                String::new(),
+                tokens[*index - 1].span,
+                Some(JaktError::ParserError(
+                    format!("expecting name"),
```
mustafaquraish, line 593: The helper function here seems fine if we're using it a lot, but it might be nice to have more descriptive errors.

src/parser.rs
```diff
@@ -1440,22 +1511,28 @@ pub fn parse_struct(
...
 
+                            let var_decl = ParsedFieldDecl {
+                                name: parsed_var.name,
+                                parsed_type: parsed_var.parsed_type,
+                                mutable: false,
+                                span: parsed_var.span,
+                                visibility,
+                            };
+
+                            // Ignore immutable flag for now
```
mustafaquraish, line 1534: If this is meant to be something to fix in the future, please add a `FIXME:` so it's clear. Otherwise it's fine.

src/parser.rs
```diff
@@ -1646,27 +1723,26 @@ pub fn parse_function(
                         TokenContents::Name(..) => {
                             // Now lets parse a parameter
 
-                            let (var_decl, err) =
-                                parse_variable_declaration(tokens, index, Visibility::Public);
+                            // let (var_decl, err) =
+                            //     parse_variable_declaration(tokens, index, Visibility::Public);
```
mustafaquraish, line 1727: Please remove any commented code

src/parser.rs
```diff
@@ -4190,105 +4252,141 @@ pub fn parse_array(tokens: &[Token], index: &mut usize) -> (ParsedExpression, Op
...
+                error = error.or(Some(JaktError::ParserError(
+                    "expected name or '(' in variable declaration".to_string(),
+                    token.span,
+                )));
+                (DeclPattern::Empty, token.span)
+            }
+        }
+    } else {
+        return (
+            // why do we have to create a structure even when we fail...this feels like a bad design.
```
mustafaquraish, line 4341: feels like a weird comment to add in...

mustafaquraish (commented):
I've left some high-level comments below, I haven't actually looked into the details of the implementation for now.

jntrnr:
If you merge main, you should get the fix to the enum tests.

jntrnr:
If you merge main, you should get the fix to the enum tests.

src/parser.rs
AndrewBregger: Oops I meant to delete that one. I was frustrated with the parser.

src/parser.rs
AndrewBregger: I didn't add that. It was already there, but I should sure it is in the right place.

src/parser.rs
AndrewBregger: This would allow trace to be written like println! and format! 

But I will undo it since it breaks the current traces. I didn't realize it at the time.

src/codegen.rs
AndrewBregger: Rust doesn't allow mutable global variables without a crate. So it is necessary.

samples/tuples/bad_type_destruct.error
AndrewBregger: I agree that is better, will change.

mustafaquraish:
Please squash your commits. It's not ideal having 2 commits undoing / changing something you implemented in the first - the history becomes messier.

src/typechecker.rs
```diff
@@ -3246,46 +3250,118 @@ pub fn typecheck_statement(
...
+                            }
+                        }
 
-            (
-                CheckedStatement::VarDecl(checked_var_decl, checked_expression),
-                error,
-            )
+                        _ => {
+                            return (
+                                CheckedStatement::Garbage,
```
alimpfard, line 3340: This won't work with generic functions, it'll appear to have successfully typechecked, but it will generate no code in the function, leading to errors
Please either add an error type that's not ignored pre-generic instantiation, or produce _something_ that makes sense given the known types and data.

AndrewBregger: Do you have an example that I could try?

mustafaquraish:

PR #407, opened 2022-05-28
Parser: Don't spell out outer enum every time in precedence()

creator1creeper1:
We don't need to spell out that we want a ParsedExpression::Operator in
every case. This also revealed that
BinaryOperator::NoneCoalescingAssign was missing in this list, so this
patch also adds a precedence for it.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -436,38 +436,37 @@ impl ParsedExpression {
     // other
     pub fn precedence(&self) -> u64 {
         match self {
-            ParsedExpression::Operator(BinaryOperator::Multiply, _)
-            | ParsedExpression::Operator(BinaryOperator::Modulo, _)
-            | ParsedExpression::Operator(BinaryOperator::Divide, _) => 100,
-            ParsedExpression::Operator(BinaryOperator::Add, _)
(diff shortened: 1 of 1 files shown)


creator1creeper1:
CI failure is unrelated.

mustafaquraish:
If you want to remove more repetition, this might work too:

```rust
   pub fn precedence(&self) -> u64 {
        use BinaryOperator::*;
        match self {
            ParsedExpression::Operator(operator, _) => match operator {
                Multiply | Modulo | Divide => 100,
                Add | Subtract => 90,
                BitwiseLeftShift
                | BitwiseRightShift
                | ArithmeticLeftShift
                ...
}
```

Though I'm no expert in Rust best practices, so take it with a grain of salt 🤷‍♂️

jntrnr:
If you merge main, you should get the fix to the enum tests.

creator1creeper1:
@mustafaquraish I didn't really want to start using that. The problem is that for example `TokenContents` and `BinaryOperator` share a lot of enum names with each other. So I'd rather just consistently not use `use Enum::*;`.

PR #411, opened 2022-05-28
codegen: Generate code for generic classes

mmeeaallyynn:
Generics are mostly working for structs and enums, but generic classes were not considered in the code for the most part.  
This should fix them for the most part. The only thing missing is type inference when accessing members, but I feel like this is out of scope for this pull request.


jntrnr:
@mmeeaallyynn - thanks for the PR. Looks like you'll need to run it through `cargo fmt` to use the standard code formatting.

mustafaquraish:

Also, please squash your commits, we shouldn't be failing CI after any single commit in your PR.

Hendiadyoin1:
It seems like you have rebased/pulled `main` ontop of your changes instead of the other way around,
please rebase your changes onto `main` instead

mmeeaallyynn:
Thanks for the pointer, this did seem off to me after looking at other PRs.

AtkinsSJ:

PR #412, opened 2022-05-28
typechecker+tests: Ensure all function types are known before typechecking declarations

### LINKED ISSUE #380: typechecker: "Condition must be a boolean expression" when branching on return value from fat arrow function

cybergsus:
Fixes #380 (again). PR #388 introduced a bug where it would use the function's *parent* scope when the function was not generic, but now it was reverted and the typechecker does a separate pass after registering functions to infer types of fat arrow functions that don't have their type stated explicitly in the source code.

### DIFF
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -1729,8 +1729,8 @@ pub fn typecheck_namespace_imports(
                         error = error.or(err);
 
                         let current_module = project.current_module_index;
-                        let module =
-                            Module::new(ModuleId(num_modules), import.module_name.name.clone());
+                        let module_id = ModuleId(num_modules);
+                        let module = Module::new(module_id, import.module_name.name.clone());
                         compiler.loaded_files[file_id].module_id = module.id;
                         let scope = Scope::new(Some(PRELUDE_SCOPE_ID), false);
                         let file_scope_id = ScopeId(module.id, 0);
@@ -1738,8 +1738,13 @@ pub fn typecheck_namespace_imports(
                         project.current_module_index = num_modules;
                         project.current_module_mut().scopes.push(scope);
 
-                        let err =
-                            typecheck_module(&parsed_namespace, file_scope_id, project, compiler);
+                        let err = typecheck_module(
+                            &parsed_namespace,
+                            module_id,
+                            file_scope_id,
+                            project,
+                            compiler,
+                        );
                         error = error.or(err);
 
                         project.current_module_index = current_module;
@@ -1780,6 +1785,7 @@ pub fn typecheck_namespace_imports(
 
 pub fn typecheck_module(
     parsed_namespace: &ParsedNamespace,
+    module_id: ModuleId,
     scope_id: ScopeId,
     project: &mut Project,
     compiler: &mut Compiler,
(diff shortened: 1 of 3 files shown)


src/typechecker.rs
```diff
@@ -2611,38 +2603,29 @@ fn typecheck_function_predecl(
...
-        }
+        let _ = project.add_var_to_scope(check_scope, checked_variable, param.variable.span);
     }
 
     let (function_return_type_id, err) =
         typecheck_typename(&function.return_type, function_scope_id, project);
     error = error.or(err);
 
-    checked_function.return_type_id = function_return_type_id;
+    let (block, _) = typecheck_block(&function.block, check_scope, project, SafetyMode::Safe);
```
jntrnr, line 2613: We don't want to typecheck the block in the `predecl` version of the function, as we won't have all the types in scope yet. The `predecl` part currently is intended only to get the names and types ready for the full check which is done in a second part of the typechecking phase.

mustafaquraish:

src/typechecker.rs
cybergsus: Yeah, but typechecking the block is the only way to infer the type of the fat arrow function

jntrnr: I mean you could, but that feels like half a fix that will fail in surprising ways in practice. Unless you check in the order that respects the dependencies of functions and types on each other, you can't start checking the bodies before we know what is in scope.

Here's an idea: do all the predecls first. For any function that has an `Unknown` as its output, check those bodies first. Then do the full typecheck of everything else. Basically one pass for the predecl (like we do now), one pass specifically for inferring return types, and then one pass to finish the typecheck (like we do now)

cybergsus: oh okay

cybergsus: maybe I should just move the typechecking of the block inside the if expression where the return type is inferred from the block because it had no hints.

cybergsus:
Hmm, I'm going to change it to two different passes.

src/typechecker.rs
cybergsus: Okay, I made the block typecheck only run when the function return type is `UNKNOWN`, although what you're saying of a second pass makes more sense. On it!

cybergsus:
I reset the pull request so I can rework it from the beginning.

src/typechecker.rs
```diff
@@ -1583,6 +1583,30 @@ pub fn typecheck_namespace_predecl(
         ));
     }
 
+    // Runs a second pass through the functions to ensure that all functions get their
```
jntrnr, line 1586: This needs to be a pass between predecl and the full typecheck rather than something that's part of predecl. Each namespace is checked during the predecl, and we want to make sure they all get a chance to run first before we do the `typecheck_block` logic.

So, basically:

* all of predecl checks run
* then your return type inference runs
* then the rest of the typechecking runs

cybergsus: So I guess it would go after `typecheck_namespace_declarations` `src/typechecker:L146`?

cybergsus:
I've made merging this into a git nightmare. I'll start again, basing my branch off `main`.

cybergsus:
Merged with module system.

cybergsus:
Closing because Rust based compiler was discontinued, and typecheck passes are different in selfhost.

PR #413, opened 2022-05-28
Parser: Allow function parameters to be mutable

creator1creeper1:
The function parameter parser tries to keep track of whether the
keyword mutable has appeared before a variable declaration. However,
it previously only used this information when considering the this
parameter. When considering any other parameter, it would instead use
the mutability information of the variable declaration. However, since
the parameter parser had already parsed the mutable keyword, the
variable declaration parser did not see the mutable keyword and
reported the variable as immutable. This commit fixes that by always
using the mutability information the parameter parser has gathered
in addition to the information that the variable declaration parser
provides. It also adds an appropriate test case.


jntrnr:
I don't quite understand what you're changing here. You can already mark parameters as mutable:

```
function foo(x: mutable i64) {
    x = 3

    println("{}", x)
}

function main(args: [String]) {
    let mutable x = 9

    foo(x)
}
```

creator1creeper1:
@jntrnr Yes, your examples work. However the following example does not work:
```jakt
function foo(mutable x: i64) {
    x = 3

    println("{}", x)
}

function main(args: [String]) {
    let mutable x = 9

    foo(x)
}
```
In your example - `let mutable x = 9` - you write `mutable` before the variable and not as part of its type (that would be `let x: mutable i64 = 9`). Therefore, I think the same order should be legal in parameter lists - we should be able to write both `function foo(mutable x: i64)` (currently not possible) and `function foo(x: mutable i64)` (already possible).

jntrnr:
Why would we want two ways to say the variable is mutable?

creator1creeper1:
We already have two way to say that variables are mutable - `mutable x: i64` and `x: mutable i64`. I think that variable declarations in parameter lists should be consistent with normal variable declarations. Also, I think it would be pretty weird if one had to say `function foo(mutable this)` but was not allowed to say `function foo(mutable x: i64)`.

jntrnr:
This is because we already know the type of `this` so it doesn't make sense to say `this: mutable`.

For where Jakt is, it's best to try to fill the gaps where it's missing functionality. Adding additional syntax for functionality it already has doesn't buy us much.

creator1creeper1:
I mean, if you think that this adds no value, close it. But I think that it's important to have consistent syntax, which is why I very much dislike the inconsistency between normal variable declarations and parameter declarations as well as between the declaration of the `this` parameter and all other parameter declarations.

PR #416, opened 2022-05-28
tests+samples: Consistently use _ in filenames

MacDue:
Most test/samples use `_` in the filenames, some break this and use `-` instead.
This switches everything to use `_`.

It's a minor thing but it kinda bugged me :^)


MacDue:
This was done (mostly) automatically with: 
```sh
find ./tests -depth -name "*-*" -execdir rename 's/-/_/g' "{}" \; 
```

So it might be possible to lint this too?

PR #418, opened 2022-05-28
meta+ci: Lint sample filenames 

MacDue:
Since #416 was merged I think it would be good to keep things consistent. 

This is a simple check to make sure all samples follow using '_' rather than '-' in their names. 

It's possible to exclude files or directories from this check by adding the globs to `SAMPLE_IGNORE_LIST` in the script. 
I've added `"samples/apps/*"` to start off, as that may contain command-line utilities where '-' is sometimes preferred.

I've tested these actions in my fork, and it seemed to work well:

image
> **N.B.** I've fixed the typos in the above screenshot now

The script is written in Python to make sure it easily works cross platform.


MacDue:
Foiled by a linter...
Somehow ironic :laughing:

PR #419, opened 2022-05-28
compiler+typechecker: Improve name resolution around namespaces

jntrnr:
Previously, we did a full typecheck of child namespaces before we checked any definitions in the parent scope. This meant we were prematurely trying to figure out things like function bodies before we even know what all functions and types were available.

With this PR, we now do typecheck of the namespace in two phases:

* Phase 1: the "predecl" phase. This establishes what declarations will be available in the namespace before they're actually fully checked.
* Phase 2: the "full" check phase. Here we finish checking the bodies of the definitions

With this, we now at least have the knowledge of what is in scope. This prevents an internal compiler panic when we mix namespaces, externs, and functions together in the same sample (as shown by the new test)

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -12,7 +12,9 @@ use crate::{
     error::JaktError,
     lexer::{lex, Span},
     parser::{parse_namespace, ParsedNamespace},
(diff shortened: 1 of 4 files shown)


jntrnr:
(read by @alimpfard)

PR #421, opened 2022-05-28
Add function overloading

### LINKED ISSUE #116: Function overloading
We should allow function overloading by parameter count and parameter types.

```jakt
function dump(value: i64) => println("{}", a)
function dump(value: String) => println("{}", a)

function main() {
    dump(123)
    dump("Hello")
}
```

creator1creeper1:
Closes #116.


src/typechecker.rs
```diff
@@ -899,6 +911,20 @@ impl EqInProject for FunctionIdentifyingInformation {
...
+        } else {
+            let mut mangled_name = String::new();
+            mangled_name.push_str(&self.name);
+            mangled_name.push_str("_returns_");
+            mangled_name.push_str(&project.typename_for_type_id(self.return_type_id));
+            return mangled_name;
+        }
+    }
+}
+
```
MacDue, line 927: I feel like doing mangling like this could harm the readability of backtraces and stuff (also why does the typechecker need to mangle, surely it's only the codegen's concern). 

What about adding an extra tag parameter to calls to mark the return, e.g. 
```c++
template<typename T>
struct ReturnTypeTag { /* empty */ };
```
then this Jakt:
```
function foo() -> i32 {}
function foo() -> i64 {}

{
  let x: i32 = foo()
  let y: i64 = foo()
}
```
becomes this C++:
```cpp
i32 foo(ReturnTypeTag<i32>);
i64 foo(ReturnTypeTag<i64>);

auto x = foo(ReturnTypeTag<i32>{});
auto y = foo(ReturnTypeTag<i64>{});
```

(i.e. return overloading in Jakt becomes normal C++ overloading).

MacDue: (also would `typename_for_type_id` work for complex types using generics/templates and stuff anyway?)

AtkinsSJ:

PR #422, opened 2022-05-28
parser+typechecker: Add support for types inside namespaces

### LINKED ISSUE #401: parser: Types cannot contain namespace qualifiers
Minimal repro:

```
namespace Foo {
    struct Bar { a: i32 }
}

function get_bar() -> Foo::Bar {
    return Foo::Bar(a: 10)
}
```

Produces: 

```
Error: expected complete block
---- file.jakt:5:26
 3 |  
 4 | function get_bar() -> Foo::Bar { 
                              ^- expected complete block
 5 |     return Foo::Bar(a: 10) 
 6 | } 
----
```

Baitinq:
Hi, this is my first jakt and rust related pull request so I might be doing something stupid:)

I believe this pull requests add support for properly parsing types inside a namespace in a function's return type.

I believe this fixes #401 . Now we get another error:

```
Error: Unknown type
---- file.jakt:5:23
 3 |
 4 | function get_bar() -> Foo::Bar {
                           ^- Unknown type
 5 |     return Foo::Bar(a: 10)
 6 | }
----
```

:^)


MacDue:
Add a sample/test for this now you've added it

jntrnr:
The way to do this would probably be to parse the type in the parser the same way we parse out a list of namespaces when we parse namespacing calls in other parts of the parser. Then later, to resolve that type to the ultimate TypeId in the typechecker

Baitinq:
Update:

- Added tests
- Added support for types inside namespaces inside the typechecker

src/parser.rs
```diff
@@ -4168,63 +4168,85 @@ pub fn parse_typename(tokens: &[Token], index: &mut usize) -> (ParsedType, Optio
...
+                    //Type in namespace
+                    *index += 1;
+                    let (inner_type, err) = parse_typename(tokens, index);
+                    if err.is_some() {
+                        error = error.or(err);
+                        return (unchecked_type, error)
+                    }
+                    if let ParsedType::Name(inner_type_name, _) = inner_type {
+                        unchecked_type = ParsedType::Name(
+                            typename + "::" + &inner_type_name,
```
jntrnr, line 4239: Can we give ParsedType a namespace specifier like we have for calls? That way when we resolve it we can walk the namespace path and find the type

Baitinq: Sure, ill try and implement that today :)

Baitinq:
@jntrnr I believe I have implemented your idea it after some struggle, please let me know if this works :^)

linusg (changes requested):
`parser+formatter:`

alimpfard:
- This doesn't allow types like `Foo::Bar<Baz>` (namespaced types with generics, for simplicity, only the last name can have generic params).
- Seems like some tests are failing, might be an actual issue, or might need a rebase.

alimpfard:
Hey @Baitinq, I had to use your changes in my PR (#518), so I ended up moving your PR over there, could you take a look at that?
would it be okay to merge that change and close this PR?

Baitinq:
Hi @alimpfard, yeah that looks great! I'll close this now:)

PR #423, opened 2022-05-28
typechecker: Store ScopeId on CheckedBlock

mattco98:
Currently, blocks do not keep track of their own scope. This makes analysis of the produced type tree pretty much impossible. This PR adds a `ScopeId` field to `CheckedBlock` in order to allow analysis (for example, getting the declaration for a random identifier nested deeply inside of a function). 

I'm not really sure if some of these blocks even really need their own scopes (for example, generated constructors), so maybe instead of creating a new scope for those, the scope id field should be optional? Not sure. I only really need it for `typecheck_block`.


jntrnr:
sure, that looks helpful

PR #425, opened 2022-05-29
parser+typechecker+tests: Support import list imports

AndrewBregger:
Adds support for `from <file> import {<list of names>}` to import only a subset of the items declared.

### DIFF
--- a/samples/imports/foo.jakt
+++ b/samples/imports/foo.jakt
@@ -1,3 +1,11 @@
 function bar() {
     println("PASS")
 }
+
+function baz() {
+    println("PASS AS baz")
+}
+
+function other() {
+    println("OTHER")
+}
--- a/samples/imports/import_list_item_not_found.error
+++ b/samples/imports/import_list_item_not_found.error
@@ -1,0 +1,1 @@
+unable to find bat in namespace foo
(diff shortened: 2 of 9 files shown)


mustafaquraish:
I haven't looked at the actual PR yet, but please update the tests to follow the format from: 99a337b5734ac8f8c74ecdcb545dfe9cabde956d

jntrnr:
From the discord chat:

"Something I was trying to figure out how to do was to import from a file that isn't the main file. Say something like:
```
// main.jakt
from foo import {Foo}
from bar import {Bar}

// foo.jakt
from bar import {Bar}
```
so you don't have to be the main file to be able to import from other files
does your PR support this or maybe something we could add later?"

Current plan is to look into aliasing so that we can handle situations like the above without duplicating definitions in the codebase.

AndrewBregger:
@jntrnr Another thought is I can close this PR and instead add proper modules. It currently functions more like an include instead of modules.

mustafaquraish:
> @jntrnr Another thought is I can close this PR and instead add proper modules. It currently functions more like an include instead of modules. 

If you've got an idea for how to do this, go for it! I was looking into it, but I'm going to busy for the next week and a bit, so don't have time to work on it myself. The current system was very much a hack to get things rolling.

AndrewBregger:
Closing to go in a different direction to solve this issue.

PR #426, opened 2022-05-29
tests: Replace *.{out,error,error_out} with inline expect comments

alimpfard:
It was getting annoying to make two files for each test, so let's
use some new shiny inline comments instead.
A test file may start with an expect block that specifies the stdout,
stderr and compiler errors associated with compiling and running the
file; previous semantics about when things are checked still apply.

An expect block is a `/// Expect` comment, starting at the first line
of the test file and spanning all the subsequent `///` comments, and
may contain `/// -` comments expressing the output requirements as
double-quoted string literals.

For instance, the following expect block:

    /// Expect:
    /// - output: "Foo\n"
    /// - stderr: "Bar\n"

asserts that the output is "Foo" followed by a newline, and that the
standard error output is "Bar" followed by a newline.
As no "error" is given, the block asserts that there are no
compilation errors.

Additionally, a file may be marked as `Expect: Skip <reason>` to make
the test runner ignore that file.


tests/typechecker/none_coalescing_assign_to_immutable.jakt
```diff
@@ -1,5 +1,8 @@
+/// Expect: Skip: Typechecks, but it shouldn't!
```
alimpfard, line 1: This doesn't fail on main, so marked it as skipped for now.

PR #427, opened 2022-05-29
runtime: Expose Iterator classes globally

nbvdkamp:
Currently using any iterator outside `main` generates invalid cpp since the `...Iterator` classes are in the `JaktInternal` namespace which `main` is also in but other functions are not. 

This change exposes them in the global namespace.

The problem was not detected by the tests since they all use `main` only. One way we could prevent such issues in the future is moving `main` from `JaktInternal` to a separate namespace. Does that sound like a good option?

### DIFF
--- a/runtime/Builtins/Array.h
+++ b/runtime/Builtins/Array.h
@@ -326,3 +326,4 @@ private:
 }
 
 using JaktInternal::Array;
+using JaktInternal::ArrayIterator;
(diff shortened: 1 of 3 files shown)


mustafaquraish (commented):
Looks reasonable to me. Can you add some tests to make sure we don't regress?

nbvdkamp:
We could add tests for each type of iterator outside of `main` but that seems like duplicate work that will also be required for future classes in `JaktInternal`. It won't be necessary if we move `main` to a separate namespace since then the normal tests catch the problem.

PR #430, opened 2022-05-29
editor: Add syntax highlighting for Sublime Text

CrystalSage:
Add syntax highlighting for Sublime Text. It should cover most of the 
syntax and any that might be introduced in the future. 

Instructions included in README.

### DIFF
--- a/editors/sublime/README.md
+++ b/editors/sublime/README.md
@@ -1,0 +1,11 @@
+# Sublime Text syntax highlighting for Jakt
+
+**Installation**
+
+1. Press `CTRL+Shift+P` to open the command palette.
+2. Go to `Preferences: Browse packages`.
+3. Put the syntax file in the directory that opens.
+
+
+After installation, Jakt syntax highlighting should get applied. If not, 
+try restarting.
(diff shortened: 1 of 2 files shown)


AtkinsSJ (changes requested):
Very nice!

Been testing this out, and it looks pretty good. I browsed through a few of the files in `samples/` and found a few missing things though:

Only the first enum value is highlighted: (samples/enums/parse.jakt)
image

`throw`, `throws` and `class` aren't highlighted. (samples/classes/throwing_method.jakt)

`weak` isn't highlighted. (samples/weak/basic.jakt)

CrystalSage:
Ah. Noted! 

Can I fix those issues in subsequent PRs?  Making more commits in this PR means I'd have to squash them and I mess up a lot while squashing.

juniorrantila:
> Ah. Noted! 
> 
> 
> 
> Can I fix those issues in subsequent PRs?  Making more commits in this PR means I'd have to squash them and I mess up a lot while squashing.

Since this is only one commit, you can modify it by doing
```
git add the-files-that-changed
git commit --amend
git push --force
```

CrystalSage:
Done. Fixed everything @AtkinsSJ found to be broken. Thank you @juniorrantila!

PR #431, opened 2022-05-29
runtime: Move main to a separate namespace

nbvdkamp:
This ensures main has the same access to identifiers as other functions.

This serves to prevent cases where generated code is incorrect when something is used outside of main and prevent bugs like the one #427 fixed.


jntrnr:
thanks

PR #432, opened 2022-05-29
typechecker: Allow omitting bound variable name on struct-like matches

alimpfard:
Doing so just binds the named field with the field name into the match
scope.


jntrnr:
yay! excellent

PR #434, opened 2022-05-29
codegen: Try really hard to resolve inter-namespace dependencies

alimpfard:
Now we can have multiple namespaces referencing each others' types and
functions without causing explosions on the C++ side.


jntrnr:
looks good

Issue #428, opened 2022-05-29
Templated class compiles in Jakt, but the C++ stage fails


mnlrsn:
A templated struct is shown in samples/generics/generic_struct.jakt. It compiles and runs. However, if class is substituted for struct (see code below), it fails in the C++ stage. 
```
class A<T> {
    a: T
}

function main() {
    let b = A(a: 0f64)
}
```

mnlrsn:
Slightly updated test case:

```
class A<T> {
    public a: T
}

function main() {
    let b = A(a: 0f64)
    println("{}", b.a)
}
```
Error message:
```
build/jakt -cr ../jakt-mv/Test-class-template-1.jakt
build/Test-class-template-1.cpp:17:24: error: no member named 'a' in 'Jakt::NonnullRefPtr<Jakt::A<double> >'; did you mean to use '->' instead of '.'?
outln(String("{}"),((b).a));
                       ^
                       ->
1 error generated.
```
The suggestion from the error message seems to be correct. When manually changing the `.` to `->` it works.

Interestingly, it works correctly when running it in the REPL (with `--run`)

alimpfard:
> Interestingly, it works correctly when running it in the REPL

The interpreter makes little distinction between classes and structs beyond how they're copied, and it doesn't go through the codegen stage anyway, so it can't hit this bug :)

0GreenClover0:
Fixed for member variables in #1122 and for methods in #1126

PR #435, opened 2022-05-29
codegen: Fix incorrect type name when using template class

boki1:
Fix incorrect type name generation when using templated classes.

Addresses issue #428.

### DIFF
--- a/src/codegen.rs
+++ b/src/codegen.rs
@@ -1151,9 +1151,25 @@ fn codegen_struct(
 
     match structure.definition_type {
         DefinitionType::Class => {
+            let mut class_name_with_generics = String::new();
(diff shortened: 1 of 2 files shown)


samples/generics/generic_class.jakt
```diff
@@ -0,0 +1,7 @@
+class A<T> {
```
jntrnr, line 1: Can you turn this into a test? You can add some comments to the top to tell the test runner how to test it. The other examples have them, so you should be able to use a similar pattern.

boki1: Sure :)

jntrnr:
thanks

PR #442, opened 2022-05-30
all: rename 'ref enum' to 'recursive enum'

jntrnr:
Renames `ref enum` to `recursive enum` for better readability


mustafaquraish:
Already much better! I was looking at the code earlier and had no idea what `ref enum` meant 😅

jntrnr:
Looks like we're going to call it `boxed enum` for now instead, since it may not always be `recursive`

PR #443, opened 2022-05-30
parser+tests: Show an error if an enum is empty

0GreenClover0:
After parsing something like:

`enum Foo { }`

a C++ error was visible. Now we check if an enum is empty and show a Jakt error with a hint.
Also add a test for this behaviour.

### DIFF
--- a/samples/enums/empty_enum.jakt
+++ b/samples/enums/empty_enum.jakt
@@ -1,0 +1,8 @@
+/// Expect:
+/// - error: "Empty enums are not allowed"
+
+enum Foo {
+    function oops(mutable this) { }
+}
+
+function main() { }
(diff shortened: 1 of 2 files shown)


src/parser.rs
```diff
@@ -872,6 +872,23 @@ pub fn parse_enum(
         // - Ident(name)
 
         skip_newlines(tokens, index);
+
+        // If the next token after an opening bracket is a closing bracket, then an enum is empty
```
alimpfard, line 876: not necessarily:
```
enum Foo {
    function oops(mutable this) { }
}
```

0GreenClover0: Good catch

cybergsus (approved):

jntrnr:
thanks

PR #446, opened 2022-05-30
parser+typechecker+codegen: add support for early returns in functions that return void

### LINKED ISSUE #436: parser: Return statements must always have a value
Early returns in functions which do not return a value are current impossible, as the parser will always attempt to parse an expression after a `return` keyword.

dipeshkaphle:
I believe this fixes #436 by making Return statements optionally have an expression. I have made changes to the typechecker and codegen as well.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -220,7 +220,7 @@ pub enum ParsedStatement {
     For((String, Span), ParsedExpression, ParsedBlock),
     Break,
     Continue,
-    Return(ParsedExpression, Span),
+    Return(Option<ParsedExpression>, Span),
     Throw(ParsedExpression),
     Try(Box<ParsedStatement>, String, Span, ParsedBlock),
     InlineCpp(ParsedBlock, Span),
@@ -1776,7 +1776,9 @@ pub fn parse_function(
                 let (block, err) = match fat_arrow_expr {
                     Some(expr) => {
                         let mut block = ParsedBlock::new();
-                        block.stmts.push(ParsedStatement::Return(expr, name_span));
+                        block
+                            .stmts
+                            .push(ParsedStatement::Return(Some(expr), name_span));
                         (block, None)
                     }
                     None => parse_block(tokens, index),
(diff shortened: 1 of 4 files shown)


src/parser.rs
```diff
@@ -2040,22 +2042,49 @@ pub fn parse_statement(
...
-                        start: start.start,
-                        end: tokens[(*index - 1)].span.end,
-                    },
-                ),
-                error,
-            )
+            if *index < tokens.len() {
+                let tok = &tokens[*index].contents;
+                match tok {
+                    TokenContents::Eol => (
```
sagering, line 2048: Don't you have to consume `TokenContents::Eol` token by advancing the index?

src/typechecker.rs
```diff
@@ -3405,27 +3405,30 @@ pub fn typecheck_statement(
...
+                        "‘return’ is not allowed inside ‘defer’".to_string(),
+                        *span,
+                    )));
+                }
 
-            (CheckedStatement::Return(output), error)
-        }
+                (CheckedStatement::Return(Some(output)), error)
+            }
+            None => (CheckedStatement::Return(None), error),
```
sagering, line 3430: `return` should also not be allowed in `defer` as above?

src/parser.rs
dipeshkaphle: Yeah. I'll add it.

src/typechecker.rs
dipeshkaphle: I'll add that check

src/typechecker.rs
```diff
@@ -3405,30 +3405,31 @@ pub fn typecheck_statement(
...
-                        .map(|i| project.functions[i].return_type_id),
-                );
-                error = error.or(err);
-
-                if project.inside_defer {
-                    error = error.or(Some(JaktError::TypecheckError(
-                        "‘return’ is not allowed inside ‘defer’".to_string(),
-                        *span,
-                    )));
+        ParsedStatement::Return(expr, span) => {
```
ADKaster, line 3408: This second commit touches the exact same lines as the first one. So, it should be fixed up into the first one.

> Squash your commits when making revisions after a patch review.

Also, the project doesn't use conventional commits, the commit category is supposed to refer to the files, classes, or subsystems you touched in the commit, rather than be `fix: ` or `feat: `

dipeshkaphle: I'll get this done and change commit to show the files touched as well

awesomekling:
Superseded by c213b581ce544d2882d7b13a0e48ae40ad00aeb8

PR #447, opened 2022-05-30
Everywhere: Implement from block expressions

MacDue:
This implements `var from { ... }` expressions that allow blocks to to be used as expressions. Where the expression evaluates to the variable `var` from the following block.

This maps directly on the the GCC block expression extension `({ ...; var; })`. This can be useful for match cases and for initialising stuff that you may want to bail out from.

This was one of the syntaxes mentioned yesterday in the `#jakt` channel on Discord, that I was rather fond of, so I wanted to try implementing a little proof of concept. I'm not sure this will survive the court of public option though :sweat_smile:

### DIFF
--- a/samples/block_exprs/block_expression_with_return.jakt
+++ b/samples/block_exprs/block_expression_with_return.jakt
@@ -1,0 +1,15 @@
+/// Expect:
+/// - output: "100\n"
+
+function mayFail() throws {
+    return 10;
+}
+
+function main() {
+    let x = y from {
+        // may return here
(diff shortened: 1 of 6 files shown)


MacDue:
An alternate syntax for this could also be `{ ... } as var` (though I think the from may be more explicit)

jntrnr:
We were chatting about this on discord the other day as another approach:

```
{
   out 3 + 4
}
```
so the `out` tells us what value to treat the block as. We could restrict it to the last line of the block.

PR #450, opened 2022-05-30
More idiomatic cat.jakt

jmalcolm137:
More idiomatic: avoid heavy use of anonymous


ADKaster:
The functions in question have since been moved into the "jakt::" standard library headers, so this PR seems to be overcome by events.

PR #459, opened 2022-05-31
Vscode: Add icon support

Lenart12:
This commit adds to new icons to vscode language support, one for dark and one for light colored themes. I tried to keep the colors and simplisity inline with the default vs-code icons feel.

Also added out folders to .gitignore


DanielNoord (commented):
Note that extensions icons are ideally 256x256, as per: https://code.visualstudio.com/api/references/extension-manifest

linusg:
- https://github.com/SerenityOS/jakt/pull/253
- https://github.com/SerenityOS/jakt/pull/372

PR #464, opened 2022-05-31
samples: Add apps/pipes.jakt

Lenart12:
Simple project using inline cpp and system calls to
demonstrate pipe, fork and dup2.

### DIFF
--- a/samples/apps/pipes.jakt
+++ b/samples/apps/pipes.jakt
@@ -1,0 +1,159 @@
+enum Stream : c_int {
+    STDIN  = 0,
+    STDOUT = 1,
+    STDERR = 2
+}
+
+struct Pipe {
+    reading_end: c_int
+    writing_end: c_int
+
+    function close(this) throws {
+        System::close(fd: .reading_end)
+        System::close(fd: .writing_end)
+    }
+}
+
+class System {
+    public function syscall_error_number() -> c_int {
+        unsafe { 
+            cpp {
+                "return errno;"
+            }
+        }
+        // Unreachable code
+        return -1
+    }
+
+    public function pipe() throws -> Pipe {
+        let mutable pipe_ret : c_int = 0;
+        let mutable fds : [c_int] = [0 as! c_int, 0 as! c_int]
+
+        unsafe {
+            cpp {
+                "pipe_ret = ::pipe(fds.unsafe_data());"
+            }
+        }
+
+        if pipe_ret == 0 {
+            return Pipe(reading_end: fds[0], writing_end: fds[1])
+        } else {
+            throw Error::from_errno(syscall_error_number() as! i32)
+        }
+
+        // Unreachable code
+        return Pipe(reading_end: 0, writing_end: 1)
+    }
+
+    public function dup2(replace: c_int, with: c_int) throws -> c_int {
+        let mutable dup2_ret : c_int = 0;
+
+        unsafe {
+            cpp {
+                "dup2_ret = ::dup2(with, replace);"
+            }
+        }
+
+        if dup2_ret >= 0 {
+            return dup2_ret
+        } else {
+            throw Error::from_errno(syscall_error_number() as! i32)
+        }
+
+        // Unreachable code
+        return -1
+    }
+
+    public function fork() throws -> c_int {
+        let mutable fork_ret : c_int = 0;
+
+        unsafe {
+            cpp {
+                "fork_ret = ::fork();"
+            }
+        }
+
+        if fork_ret >= 0 {
+            return fork_ret
+        } else {
+            throw Error::from_errno(syscall_error_number() as! i32)
+        }
+
+        // Unreachable code
+        return -1
+    }
+
+    public function read(fd: c_int, count: usize) throws -> String {
+        let mutable read_ret : c_int = 0;
+        let mutable buffer : String = String::repeated(character: 0x00, count: count);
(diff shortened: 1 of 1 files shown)


linusg (changes requested):
https://github.com/SerenityOS/jakt/blob/main/CONTRIBUTING.md#code-submission-policy

samples/apps/pipes.jakt
```diff
@@ -0,0 +1,183 @@
...
+    STDERR = 2
+}
+
+struct Pipe {
+    reading_end: c_int
+    writing_end: c_int
+}
+
+class System {
+    public function getErrno() -> c_int {
```
linusg, line 14: Jakt follows a similar coding style as serenity, and serenity uses `snake_case` for methods and omits `get_` from getters.

samples/apps/pipes.jakt
```diff
@@ -0,0 +1,183 @@
...
+        unsafe { cpp {
+            "return errno;"
+        }}
```
linusg, line 17:
```suggestion
        unsafe {
            cpp {
                "return errno;"
            }
        }
```

samples/apps/pipes.jakt
```diff
@@ -0,0 +1,183 @@
...
+        // Unreachable code
+        return -1
+    }
+
+    public function pipe() throws -> Pipe  {
+        unsafe{
```
linusg, line 23: Not nearly all of this needs to be wrapped in unsafe. In fact, it looks like *only* the cpp block needs it.

samples/apps/pipes.jakt
```diff
@@ -0,0 +1,183 @@
...
+                throw Error::from_errno(getErrno() as! i32)
+            }
+
+            // Unreachable code
+            return Pipe(reading_end: 0, writing_end: 1)
+        }
+    }
+
+    public function dup2(replace: c_int, with: c_int) throws -> c_int {
+        unsafe{
```
linusg, line 42: Likewise

samples/apps/pipes.jakt
linusg: Also missing a space between the keyword and {

samples/apps/pipes.jakt
```diff
@@ -0,0 +1,183 @@
...
+                throw Error::from_errno(getErrno() as! i32)
+            }
+
+            // Unreachable code
+            return -1
+        }
+    }
+
+    public function fork() throws -> c_int {
+        unsafe{
```
linusg, line 61: Likewise

samples/apps/pipes.jakt
```diff
@@ -0,0 +1,183 @@
...
+            let mutable fork_ret : c_int = 0;
+
+            cpp {
+                "fork_ret = ::fork();"
+            }
+
+            if fork_ret as! i64 == -1 {
```
linusg, line 68: Do we have to cast this? `dup2_ret >= 0` works...

samples/apps/pipes.jakt
```diff
@@ -0,0 +1,183 @@
...
+                return fork_ret
+            }
+
+            // Unreachable code
+            return -1
+        }
+    }
+
+    public function wait() throws -> c_int {
+        unsafe{
```
linusg, line 80: Likewise

linusg: I'll stop pointing these out now, please find the rest yourself :)

samples/apps/pipes.jakt
```diff
@@ -0,0 +1,183 @@
+
```
linusg, line 1: Weird newline here

samples/apps/pipes.jakt
```diff
@@ -0,0 +1,183 @@
...
+            let c_exit = exit_code as! c_int 
+            cpp {
+            "::_exit(c_exit);";
+            }
+        }
+    }
+}
+
+function main() {
+
```
linusg, line 161: Weird newline here

linusg (changes requested):
Code style comments:

samples/apps/pipes.jakt
Lenart12: Yes, for now (#465)

samples/apps/pipes.jakt
Lenart12: Function name cant be just `errno` since its defined as a macro. Will rename to `syscall_error_number`

Lenart12:
Changed commit message and implemented changes mentioned in the review. Some weird arithmetic which I don't like was needed because negated c_int needs to be fixed (to avoid things like `not (close_ret >= 0)`, `close_ret == -1` would be better!).

linusg (changes requested):
`Adds` is not imperative mood

samples/apps/pipes.jakt
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        // Unreachable code
+        return -1
+    }
+
+    public function read(fd: c_int, count: usize) throws -> String {
+        let mutable read_ret : c_int = 0;
+
+        unsafe {
```
linusg, line 109: not all of this needs to be in `unsafe`

Lenart12:
Sorry this is taking so long, english is not my strong suit :(
Unsafe was done like that before because if I had two unsafe blocks the char buffer wouldnt be defined in the second block but I have refactored it so there is only one now.

PR #466, opened 2022-05-31
typechecker: Added cint type id to negate

### LINKED ISSUE #465: Parser: c_int type cast error
Compiling
```
function main() {
    let array: [c_int] = [-1 as! c_int, -1 as! c_int]
}
```
gives following error
```
Error: negate on non-numeric value
```
Agni on [discord](https://discord.com/channels/830522505605283862/976984180313452604/981255634076966962):
> that looks like a parsing mistake :yakthonk: 
> unary `-` should bind more tightly than `as` and friends

Lenart12:
Before we couldn't typecast negative constants to c_int.
This fix makes it so now we can do things like `-1 as! c_int`.
Fixes #465.

### DIFF
--- a/tests/typechecker/array_negative_c_int.jakt
+++ b/tests/typechecker/array_negative_c_int.jakt
@@ -1,0 +1,9 @@
+/// Expect:
+/// - output: "-1\n2\n"
+
+function main() {
+    let array: [c_int] = [-1 as! c_int, 2 as! c_int]
+
+    println("{}", array[0])
+    println("{}", array[1])
+}
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -5394,7 +5394,8 @@ pub fn typecheck_unary_operation(
                 | crate::compiler::U32_TYPE_ID
                 | crate::compiler::U64_TYPE_ID
                 | crate::compiler::F32_TYPE_ID
-                | crate::compiler::F64_TYPE_ID => {
+                | crate::compiler::F64_TYPE_ID
+                | crate::compiler::CINT_TYPE_ID => {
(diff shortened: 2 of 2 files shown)


tests/typechecker/array_negative_c_int.jakt
```diff
@@ -0,0 +1,10 @@
+/// Expect:
+/// - output: "-1\n-1\n2\n"
+
+function main() {
+    let array: [c_int] = [-1 as! c_int, -1 as! c_int, 2 as! c_int]
```
linusg, line 5: What's the point of doing the same test twice here?

linusg (changes requested):
https://github.com/SerenityOS/jakt/blob/main/CONTRIBUTING.md#code-submission-policy

Lenart12:
Changed the commit message and changed up the test.
`if array[0] == -1 as! c_int` still feels weird, but atleast its castable now.
Would be cool if we could do just `if array[0] == -1` since casting -1 to c_int would never fail.

tests/typechecker/array_negative_c_int.jakt
```diff
@@ -0,0 +1,11 @@
+/// Expect:
+/// - output: "-1\n2\n"
+
+function main() {
+    let array: [c_int] = [-1 as! c_int, 2 as! c_int]
+
+    if array[0] == -1 as! c_int {
```
linusg, line 7: I don't understand why this is needed. Just print `array[0]` unconditionally...?

linusg (changes requested):
`Adds` is not imperative mood

linusg (approved):
Thanks!

PR #467, opened 2022-05-31
vscode+compiler: Better vscode errors

jntrnr:
Improves error reporting for VSCode (and improves experience for other editors):

* Adds a `check-only` mode to the compiler, so we aren't generated C++ we don't need if we only care about type errors
* Adds a JSON error mode to make parsing errors much easier for editors
* Updates the VSCode extension to use the JSON errors


jntrnr:
(checked by @awesomekling)

PR #468, opened 2022-05-31
vscode: Fix enum and function that `throws` highlighting

0GreenClover0:
Image for reference: https://imgur.com/a/WE9i53U
This makes underlined words from section 1 "variable.field.enum.jakt". Words from section 2 are "entity.name.type.enum.jakt" (no change here).
The only thing I'm not sure is, if underlined words in sections 1 and 2 should be both "variable.field.enum.jakt" or "entity.name.type.enum.jakt", or the way it is on the screenshot, or maybe even something else.
Marking them both as a type makes some sense, but I don't like (a bit) how it looks after that: https://imgur.com/a/Bi7fLcx

Apart from enum highlighting, this also fixes
`function Foo() throws {`
where `throws` was not highlighted but when we had something like:
`function Foo() throws -> String {`
it was highlighted.

Before: https://imgur.com/a/GWTI2V5
After: https://imgur.com/a/jdW8Fu7

### DIFF
--- a/editors/vscode/syntaxes/jakt.tmLanguage.json
+++ b/editors/vscode/syntaxes/jakt.tmLanguage.json
@@ -243,37 +243,16 @@
     "enum-field": {
       "patterns": [
         {
-          "name": "meta.field.enum.jakt",
-          "begin": "\\b((?:\\w|_)(?:\\w|_|[0-9])*)\\s*(\\{)",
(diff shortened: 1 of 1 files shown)


0GreenClover0:
Did it count a space at the end of a commit message? I had exactly 72 characters, not my lucky day :')

PR #470, opened 2022-05-31
vscode: fix comment highlighting in enum block

ryangjchandler:
AK noticed this whilst working on the JSON video, straight-forward fix.


jntrnr:
thanks

0GreenClover0:
I've tried doing that but I didn't like that you get different highlighting when there is a comma after a variant and also a different one when you specify the type :P Although it might only be my editor? I'm not an expert (or maybe even that is a desirable effect)

[https://imgur.com/FvAdUGi](https://imgur.com/FvAdUGi)

PR #476, opened 2022-06-01
codegen: implement match with switch instead of Variant::visit

maddanio:
this replaces the Variant::visit based implementation of match by one based on a switch statement.
on my machine that reduces the total build time of self hosted from 2s to 1.2s.


awesomekling:
Awesome :^)

PR #478, opened 2022-06-01
vscode: Switch compilation to esbuild and clean up language server

Lenart12:
This change allows us to make bundled extensions, reducing file size.
Also adds placeholder license to stop warnings.

Add various changes to remove errors or warnings in language server
caused by eslint and remove unused code.


jntrnr:
Thanks for fixing this up!

I was wondering - can you leave the unused other LSP features for now instead of deleting them? We're thinking of adding things like completions, and it'd be nice to leave the template in place so we know where it goes.

Lenart12:
> Thanks for fixing this up!
> 
> I was wondering - can you leave the unused other LSP features for now instead of deleting them? We're thinking of adding things like completions, and it'd be nice to leave the template in place so we know where it goes.

I reverted deleting completion skeleton. Hope this is okay now.

jntrnr:
@Lenart12 - this seems to have broken running "debug" from vscode inside the vscode folder. It used to spawn a new vscode instance with the extension running, but now it seems to spin forever on the watch.

PR #482, opened 2022-06-01
main+vscode: Json array output

Lenart12:
Using `-j` flag now outputs all errors inside an array instead of line by line.

Like so:
```
[{"message": "variable 'inputt' not found", "severity": "Error", "span": {"start": 7381, "end": 7387}}, {"message": "Did you mean 'input'?", "severity": "Hint", "span": {"start": 7381, "end": 7387}}]

### DIFF
--- a/editors/vscode/server/src/server.ts
+++ b/editors/vscode/server/src/server.ts
@@ -201,23 +201,21 @@ async function validateTextDocument(textDocument: TextDocument): Promise<void> {
 
 	const diagnostics: Diagnostic[] = [];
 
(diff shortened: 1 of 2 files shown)


Lenart12:
> @Lenart12 - just curious, does this fix an issue with the current system? Having json per line is also valid (like json object files)

No "issue" but I think it's cleaner to output it as a single json array (so its easier to pipe output to other programs if necesary). If this PR gets rejected I understand.

jntrnr:
Let's keep the current system in case we want to stream errors out in the future. That way we aren't waiting and bundling them up, but instead can handle them as the compiler outputs them.

PR #487, opened 2022-06-02
vscode: Separate function and field autocompletions

u9g:
Simple yet effective way to get field and function completions separately:
abc


jntrnr:
nice touch

PR #502, opened 2022-06-03
editors/vscode: Cache line break positions for faster span conversion

awesomekling:
This whole situation is a mess, since LSP requires that we provide
file offsets and column indices in UTF-16 "coordinates". This patch
doesn't improve the architecture, but instead makes a cache of line
break locations in the document text so we don't have to scan from the
start of the document every time we want to convert between systems.

This significantly reduces UI lag caused by the type inlay hints.

The *nice* solution to all this would be that vscode adds support for
UTF-8 offsets in the future. There's a new LSP extension for this,
but so far they only advertise UTF-16 support.

### DIFF
--- a/editors/vscode/server/src/server.ts
+++ b/editors/vscode/server/src/server.ts
@@ -111,6 +111,7 @@ connection.onDefinition(async (request) => {
 	const text = document?.getText();
 
 	if (typeof text == "string") {
+		const lineBreaks = findLineBreaks(text);
 		// console.log("request: ");
 		// console.log(request);
 		// console.log("index: " + convertPosition(request.position, text));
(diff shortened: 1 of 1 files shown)


editors/vscode/server/src/server.ts
```diff
@@ -283,24 +285,21 @@ documents.onDidChangeContent((
...
-		}
-
-		if (buffer.at(i) == 0x0A) {
-			line++;
-			character = 0;
-		} else {
-			character++;
+function convertSpan(utf8_offset: number, utf16_text: string, lineBreaks: Array<number>): Position {
+	let line = undefined;
+	for (let n = 0; n < lineBreaks.length; ++n) {
```
Lenart12, line 290: Here instead of a linear search we could do a binary one since line breaks are in order.

awesomekling: We could absolutely do that :)
I'll leave that for our future selves to do.

PR #504, opened 2022-06-03
selfhost: Parse fat arrow in function definition

lanmonster:
Parse the fat arrow token, and the following expression.
Make a new block and set the parsed expression as the only
statement so that return type inference can operate on the block.

### DIFF
--- a/selfhost/main.jakt
+++ b/selfhost/main.jakt
@@ -1366,17 +1366,25 @@ struct Parser {
                 return_type = .parse_typename()
                 return_type_span = JaktSpan(start, end: .index)
             }
-            else => {
-                .error("Expected ->", .current().span())
-            }
+            else => {}
         }
 
-        // FIXME: Accept an (optional) fat arrow ('=>')
-
-        parsed_function.block = .parse_block()
+        parsed_function.block = match .current() {
+            FatArrow => .parse_fat_arrow()
+            else => .parse_block()
+        }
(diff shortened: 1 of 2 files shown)


AtkinsSJ:
Nit-pick: The category should be `selfhost` not `parser`, I believe.

awesomekling:
We should also support explicit return type together with fat arrow.
`function foo() -> i64 => 5`

selfhost/main.jakt
```diff
@@ -1366,19 +1366,30 @@ struct Parser {
...
-                .error("Expected ->", .current().span())
-            }
+            else => {}
         }
 
-        // FIXME: Accept an (optional) fat arrow ('=>')
-
-        parsed_function.block = .parse_block()
+        
+        parsed_function.block = .parse_fat_arrow()
```
awesomekling, line 1373: This looks a bit strange. If there's no fat arrow present, the function block is still parsed by `parse_fat_arrow()`.
Please rearrange the code so that `parse_fat_arrow()` isn't responsible for the non-fat-arrow case. :)

awesomekling (changes requested):

selfhost/main.jakt
lanmonster: I was trying to avoid having two `match .current() {` right next to each other, but I will rearrange.

PR #505, opened 2022-06-03
vscode: Optimize convertSpan

Lenart12:
Use binary search to find the correct line index and remove unnecesary loop to get the index of the character in the line.

### DIFF
--- a/editors/vscode/server/src/server.ts
+++ b/editors/vscode/server/src/server.ts
@@ -126,8 +126,13 @@ connection.onDefinition(async (request) => {
 			// console.log(obj);
 
 			console.timeEnd('onDefinition');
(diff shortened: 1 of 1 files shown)


editors/vscode/server/src/server.ts
```diff
@@ -276,34 +287,42 @@ function throttle(fn: (...args: any[]) => void, delay: number) {
...
 
+function lowerBoundBinarySearch(arr: number[], num: number): number {
+	let low = 0;
+	let mid = 0;
+	let high = arr.length - 1;
+
+	if(num >= arr[high]) return high;
+
+	while (low < high) {
+		// Bitshift to avoid floating point division
```
awesomekling, line 306: This sounds like broscience but sure :)

PR #506, opened 2022-06-03
selfhost: Parse structs and classes

AtkinsSJ:
Some work towards struct and class support. This does *not* include parsing member access, so most of the samples/tests for structs and classes will still fail to compile. But, I figured it was better to get this submitted sooner rather than later to not hold anyone else up.

This is my first contribution to the self-hosted compiler/using Jakt, so I'm 100% sure I've done some things wrong/weirdly, so don't hesitate to tell me. :^)


awesomekling:
Looks super tidy :)

PR #518, opened 2022-06-04
Everywhere: Don't say hello to C++ interop, have some bugfixes instead

alimpfard:
- Pulls https://github.com/SerenityOS/jakt/pull/422 and adds some of my own changes
- Puts everything in namespace Jakt, this should at least avoid having ODR violations with the rest of the universe
- Typechecks some more (`raw T` was never resolved)
- that's it, no interop in this PR.


alimpfard:
With this, I call this PR ready to go:
```rs
import extern "LibJS/Interpreter.h" as anon {
    namespace AK {
        extern struct NonnullRefPtr<T> {
            function ptr(this) -> raw T
        }
        extern struct NonnullOwnPtr<T> {
            function ptr(this) -> raw T
        }
        extern struct StringView {
            function StringView(anon s: raw c_char) -> StringView
        }
        extern struct Vector<T> {
            function size(this) -> usize
            function at(this, anon index: usize) -> T
        }
        extern struct Result<T, E> {
            function is_error(this) -> bool
            function release_value(mutable this) -> T
        }
    }
    namespace JS {
        namespace Parser {
            extern struct Error {
            }
        }
        extern struct GlobalObject {
        }
        extern struct Value {
            function as_i32(this) -> i32
        }
        extern struct VM {
            // This is a lie, this takes default parameters.
            function create() -> AK::NonnullRefPtr<JS::VM>
        }
        extern struct ThrowCompletionOr<T> {
            function is_error(this) -> bool
            function release_value(mutable this) -> T
        }
        extern struct Realm {
            function create(anon vm: JS::VM) -> raw JS::Realm
        }
        extern struct Script {
            function parse(source: AK::StringView, anon realm: JS::Realm) -> AK::Result<AK::NonnullRefPtr<JS::Script>, AK::Vector<JS::Parser::Error> >
        }
        extern struct Interpreter {
            function create<GlobalObjectType>(anon vm: JS::VM) -> AK::NonnullOwnPtr<JS::Interpreter>
            function run(mutable this, anon script: JS::Script) -> JS::ThrowCompletionOr<JS::Value>
        }
    }
}

function main() {
    unsafe {
        let mutable vm = JS::VM::create();
        let mutable realm = JS::Realm::create(*vm.ptr());
        let src = "42 + 27;"
        let mutable result = JS::Script::parse(source: AK::StringView(src.c_string()), *realm);
        if result.is_error() {
            return 1
        }

        let mutable script = result.release_value();
        let mutable interpreter = JS::Interpreter::create<JS::GlobalObject>(*vm.ptr());
        let mutable exec_result = (*interpreter.ptr()).run(*script.ptr());
        if exec_result.is_error() {
            return 2
        }

        let value = exec_result.release_value();
        println("{}", value.as_i32())
    }
}
```

src/compiler.rs
```diff
@@ -35,9 +35,8 @@ pub const F32_TYPE_ID: usize = 11;
 pub const F64_TYPE_ID: usize = 12;
 pub const CCHAR_TYPE_ID: usize = 13;
 pub const CINT_TYPE_ID: usize = 14;
+// Note: keep USIZE_TYPE_ID last as it is how we know how many slots to pre-fill
 pub const USIZE_TYPE_ID: usize = 15;
```
ADKaster, line 39: So... could we replace this with `pub const LAST_BUILTIN_TYPE_ID: usize = 15;` ?

cxboog: We could, but FOO_TYPE_IDs are treated as a valid unique type id, your suggestion kinda aliases the usize one
I don't mind either

cxboog: Dammit, wrong account _again_?

alimpfard: Done.

runtime/Jakt/CharacterTypes.h
```diff
@@ -172,30 +172,30 @@ constexpr u32 to_ascii_base36_digit(u32 digit)
...
-using AK::is_unicode_surrogate;
-using AK::parse_ascii_base36_digit;
-using AK::parse_ascii_digit;
-using AK::parse_ascii_hex_digit;
-using AK::to_ascii_base36_digit;
-using AK::to_ascii_lowercase;
-using AK::to_ascii_uppercase;
+
+
+
```
ADKaster, line 177: Replacing these using declarations with newlines is definitely a statement, but did you intend to delete the newlines after? 😬

ADKaster: (extraneous newlines show up in other files too)

alimpfard: I decided I'd be lazy and remove them later, later never came. will fix

ADKaster (approved):
Looks good from the AK/Prelude side. I can't speak to the rust code though 🙂. Less ODR collisions when linking against serenity libs (even LibC!!) Is pretty sweet.

PR #520, opened 2022-06-04
typechecker: Check whether each match arm returns

### LINKED ISSUE #394: typechecker: Block return detection ignores returns in `match` expressions
```
enum Foo {
    Id(i64)
    Name(String)
}
function test(x: Foo) -> String {
    match this {
        Name(name) => {
            return name
        }
        Id(id) => {
            return format("anon:{}", id)
        }
    }
}
```

Errors with "control reaches end of non-void function", but `match` must always be exhaustive, and all its arms return from the function.

F35idk:
This fixes #394.

### DIFF
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -1358,6 +1358,25 @@ impl CheckedExpression {
             _ => false,
         }
     }
+
+    pub fn definitely_returns(&self) -> bool {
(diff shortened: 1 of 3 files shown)


AtkinsSJ:
Can we have a test for the error condition too?

F35idk:
Done.

PR #523, opened 2022-06-04
vscode: Add Temporary Icons

IrishBruse:
Add temporary icons to vscode extension.


awesomekling:
We're not adding anything that might be interpreted as an icon at the moment.

PR #524, opened 2022-06-04
editors/vscode: Add syntax highlighting for namespace declarations

Maxssho13:
Previously the keyword `namespace` would be gray in namespace declarations, now it is treated as a keyword in a similar way to struct definitions.

Before this commit:
Screen Shot 2022-06-04 at 2 07 24 PM

Now:
Screen Shot 2022-06-04 at 2 04 00 PM


jntrnr:
Thanks!

PR #528, opened 2022-06-04
all: Rename anonymous to anon globally

jntrnr:
We've gone back and forth on this one. I propose we try `anon` again. Some trade-offs:

* `anonymous` is more clear but much harder to type (I regularly mistype it as `anonymouse` 🐭 )
* `anon` is shorter though perhaps a little vague? I think it should be obvious with a little time, but we should pay attention to if code gets harder to read or stays the same

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ rect.set_size(width: 640, height: 480)
 
 There are two exceptions to this:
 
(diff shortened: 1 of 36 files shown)


awesomekling:
I'm fine with this. `anonymous` is a mouthful.

PR #529, opened 2022-06-04
editors: Add syntax highlighting for emacs

veera-sivarajan:
This PR adds an emacs major mode for highlighting syntax and indentation. I've tried my best to highlight all keywords by looking at jakt extensions for other editors. The indentation offset defaults to 4 and can be changed by the user if need be.


jntrnr:
thanks!

PR #534, opened 2022-06-05
parser+typechecker+codegen+tests: Basic module support

AndrewBregger:
Adds basic module support to typecheck and codegen.

The project is now split into modules and every id now has the module it comes from. Codegen generates the code for each module and wraps it in a namespace. The only module that doesn't get a namespace is prelude (which is its own module).

### DIFF
--- a/tests/main.rs
+++ b/tests/main.rs
@@ -446,6 +446,11 @@ fn test_imports() -> Result<(), JaktError> {
 }
 
 #[test]
+fn test_modules() -> Result<(), JaktError> {
+    test_samples("samples/modules")
+}
+
+#[test]
 fn test_selfhost() -> Result<(), JaktError> {
     test_samples("selfhost")
 }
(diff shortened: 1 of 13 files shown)


tests/main.rs
```diff
@@ -254,7 +254,7 @@ fn test_samples(path: &str) -> Result<(), JaktError> {
                 let mut exe_name = temp_dir();
                 exe_name.push(format!("output{}", uuid));
 
-                let status = Command::new("clang++")
+                let status = Command::new("clang++-11")
```
jntrnr, line 257: we should probably leave these as `clang++` as not all systems will have `clang++11` (mine doesn't, for example)

jntrnr (commented):
This is an interesting approach. It's definitely worth a try as it could simplify how we handle definitions later on, too.

Can you add some sample tests? You could add a directory for just the module tests in the `samples` directory

tests/main.rs
AndrewBregger: That is a local change I need to do to get the testing to work. Will undo.

ADKaster: If you're using ubuntu, I'm a fan of this script to keep things up to date as versions bump up :D
https://gist.github.com/junkdog/70231d6953592cd6f27def59fe19e50d

AndrewBregger:
@jntrnr Tests have been added.

jntrnr:
Thanks for putting this together! We'll land and then continue improving as needed.

One thing this lets us do is to move the selfhost compiler to being multi-file. This is one of the last pieces that it.

PR #547, opened 2022-06-05
all: Rename `let mutable` to `mut` globally

awesomekling:
After writing a couple thousand lines of Jakt, the main thing that
annoys me (other than compiler bugs) is definitely `let mutable`.
It takes up an inordinate amount of horizontal space, and I find
myself frequently "pausing" while my eyes search for the actual
identifier part of variable declarations.

To reduce this friction, this patch shortens `let mutable` to `mut`.

Likewise, `mutable` modifiers on function parameters become `mut`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,18 +163,18 @@ let x = Foo()
 x.func()
 ```
 
-**Mutating member functions** require an object to be called, and may modify the object. The first parameter is `mutable this`.
+**Mutating member functions** require an object to be called, and may modify the object. The first parameter is `mut this`.
 ```jakt
 class Foo {
     x: i64
 
-    function set(mutable this, anon x: i64) {
+    function set(mut this, anon x: i64) {
         this.x = x
     }
 }
 
-// Foo::set() can only be called on a mutable Foo:
-let mutable foo = Foo(x: 3)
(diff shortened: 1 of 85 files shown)


AtkinsSJ:
How about just using `mutable` instead of `let mutable`? I like that it's an actual word.

awesomekling:
After using it for a few hours, I really enjoy how `let` and `mut` declarations align vertically.
But it's a living design, if we end up hating `mut` we'll change it again. :)

kud1ing:
> I find myself frequently "pausing" while my eyes search for the actual
identifier part of variable declarations.

I tend to agree. Keywords don't need to be that descriptive, since you know them by heart: there are only a handful and you have used them all the time. The longer the keywords are, the more they distract from the important stuff, the identifiers in the project. For example:

```
mut interrupts ...
mut file_system ...
```

```
mutable interrupts ...
mutable file_system ...
```

```
let mutable interrupts ...
let mutable file_system ...
```

I prefer the first.

kud1ing:
Or with functions/methods:

```
pub fn wait_until_coredump_is_ready(this) ...
```

```
public function wait_until_coredump_is_ready(this) ...
```

PR #551, opened 2022-06-05
compiler+typechecker+codegen: Root module created properly

AndrewBregger:
Fixed a bug where root module was not being created properly, resulting in the scopes being incorrect.

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -187,21 +187,38 @@ impl Compiler {
             .expect("Cannot find parent directory of file");
         self.include_paths.push(parent_dir.to_path_buf());
 
-        // should always be 0
-        let file_scope_id = ScopeId(project.current_module().id, 0);
-
         let file_id = match self.find_or_load_file(fname) {
             Ok(file_id) => file_id,
-            Err(e) => return (file_scope_id, Some(e)),
+            Err(e) => return (ScopeId(ModuleId(project.current_module_index), 0), Some(e)),
         };
 
         let (parsed_namespace, err) = self.parse_file(file_id);
         error = error.or(err);
 
         let scope = Scope::new(Some(PRELUDE_SCOPE_ID), false);
+
+        let module_name = fname
+            .file_stem()
+            .expect("expected a filename for input file")
+            .to_ascii_lowercase()
+            .to_str()
+            .expect("filename should be valid utf-8")
+            .to_string();
+
+        let module_id = project.add_module(module_name);
+
+        // should always be 0
+        let file_scope_id = ScopeId(module_id, 0);
+        project.current_module_index = module_id.0;
+
+        assert_eq!(module_id.0, 1);
+
         project.current_module_mut().scopes.push(scope);
+        // setting the input module as root, meaning this should have main.
+        project.set_current_module_root();
 
         let err = typecheck_module(&parsed_namespace, file_scope_id, project, self);
+
         error = error.or(err);
 
         (file_scope_id, error)
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -262,6 +262,8 @@ pub struct Module {
     pub enums: Vec<CheckedEnum>,
     pub scopes: Vec<Scope>,
     pub types: Vec<Type>,
+
+    pub is_root: bool,
 }
 
 impl Module {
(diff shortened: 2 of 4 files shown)


src/typechecker.rs
```diff
@@ -7255,9 +7277,11 @@ pub fn check_types_for_compat(
             if rhs_type_id != lhs_type_id {
                 error = error.or(Some(JaktError::TypecheckError(
                     format!(
-                        "Type mismatch: expected ‘{}’, but got ‘{}’",
+                        "Type mismatch: expected ‘{}’, but got ‘{}’ {:?}  {:?}",
                         project.typename_for_type_id(lhs_type_id),
                         project.typename_for_type_id(rhs_type_id),
+                        lhs_type_id,
+                        rhs_type_id
```
jntrnr, line 7284: This looks like debug info?

src/compiler.rs
```diff
@@ -187,21 +187,38 @@ impl Compiler {
             .expect("Cannot find parent directory of file");
         self.include_paths.push(parent_dir.to_path_buf());
 
-        // should always be 0
-        let file_scope_id = ScopeId(project.current_module().id, 0);
-
         let file_id = match self.find_or_load_file(fname) {
             Ok(file_id) => file_id,
-            Err(e) => return (file_scope_id, Some(e)),
+            Err(e) => return (ScopeId(ModuleId(project.current_module_index), 0), Some(e)),
```
jntrnr, line 192: Does the `file_id` here point to the prelude instead of the user's file (since you create the module for the user's file after)?

AndrewBregger: It will point to the prelude. Technically this should point to the module that is importing the new module. Since the Rust compiler is deprecated, the refactor isn't necessary.

src/typechecker.rs
AndrewBregger: It is.

src/compiler.rs
jntrnr: We'll still need to get the Rust compiler to where it can handle multi-file support since without it would mean putting everything in one file for the selfhost compiler.

jntrnr:
You'll also want to update ide.rs so that it uses the same logic. Without it, the IDE support will break when this lands.

src/compiler.rs
AndrewBregger: I have a hack to add qualified names. I can add it to this PR. Which should fix using struct, classes, and enums from imported modules. That should be the last think that I know of that keeps multi-file support from working.

AndrewBregger:
> You'll also want to update ide.rs so that it uses the same logic. Without it, the IDE support will break when this lands.

I'm not sure what needs to be changed here.

src/compiler.rs
```diff
@@ -187,21 +187,38 @@ impl Compiler {
...
 
         let (parsed_namespace, err) = self.parse_file(file_id);
         error = error.or(err);
 
         let scope = Scope::new(Some(PRELUDE_SCOPE_ID), false);
+
+        let module_name = fname
+            .file_stem()
+            .expect("expected a filename for input file")
+            .to_ascii_lowercase()
```
Lenart12, line 203: Why does it need to be lowercase?

AndrewBregger: Lowercase imports make the most sense to me but it really doesn't matter.

jntrnr:
> > You'll also want to update ide.rs so that it uses the same logic. Without it, the IDE support will break when this lands.
> 
> I'm not sure what needs to be changed here.

If you try your branch in the IDE, you won't be able to get any of the IDE features to work. I suspect that's because you're now putting things in the second module instead of the first. If you update the support to start looking in the second module, I think it should work again.

AndrewBregger:
```rust
pub fn find_span_in_project(project: &Project, span: Span) -> Option<Usage> {
    // lol, for now since vscode is going to send us a url instead
    // of a filepath, let's just assume we are going to look at the file
    // at hand.

    for module in project.modules.iter() {
        let scope = project.get_scope(ScopeId(module.id, 0));

        return find_span_in_scope(project, scope, span);
    }

    None
}
```
It searches the entire project for the span...

jntrnr:
I think you need to not return on the first one, as it will return after looking in the prelude.

src/compiler.rs
jntrnr: That'd be great if you have that. I think once we have it landed and can start moving to multi-file projects, we can do that with selfhost and really move everyone over to contributing to selfhost instead of the rust-based compiler.

jntrnr:
Great, thanks for putting this together :)

PR #553, opened 2022-06-06
selfhost+runtime: Small progress on self-hosting

awesomekling:
The runtime patch makes Array capacity grow in larger increments, which makes the self-hosted compiler lex itself in milliseconds instead of 10+ seconds 😅 
The other patch adds character and byte literal parsing to selfhost.


awesomekling:
Dropped the character and byte literal stuff in favor of #554 :)

PR #555, opened 2022-06-06
selfhost: Parse `PreIncrement` operator

Maxssho13:


selfhost/main.jakt
```diff
@@ -1983,6 +1983,22 @@ struct Parser {
...
+
+                let expr = .parse_operand()
+                let span = merge_spans(start, expr.span())
+                return ParsedExpression::UnaryOp(expr, op: UnaryOperator::PreIncrement, span)
+            }
+            And => {
+                .index++
+                return ParsedExpression::Operator(op: BinaryOperator::LogicalAnd, span)
+            }
+            Or => {
```
jntrnr, line 1998: I think `and` and `or` should be handled in `parse_operator`?

Maxssho13: In the rust compiler in `parser.rs`, it handles them in both `parse_operator` and `parse_operand` so I'm not really sure in what case `parse_operand` would be called with `And` or `Or` but it seems to think it needs to so I just trusted that.

jntrnr: The Rust-based compiler was written very quickly, so it may not use the best techniques in some places. Binary operators should probably all get parsed by `parse_operator` so that we have one place where we can understand the operator and find the proper precedence during precedence parsing.

Maxssho13: Ah makes sense, thanks!

PR #556, opened 2022-06-06
all: Add import lists to compiler and selfhost

jntrnr:
This adds import lists to imports, so that you can import specific symbols from an import.

Current syntax comes from the proof-of-concept:

```
import foo with Bar, Baz, Qux
```

This imports foo.jakt in the import path and then aliases the given imports in the import list in the current scope. Codegen will skip over these as their ModuleId is not the current file's ModuleId.

With that, I split the selfhost into two files. We'll need to do a bit more work to split further, but this should give us a start.


src/typechecker.rs
```diff
@@ -1783,7 +1783,37 @@ pub fn typecheck_namespace_imports(
...
+                                }
+                            }
+                            for (name, id, _) in &module_scope.functions {
+                                if name == &imported_item.name {
+                                    scope
+                                        .functions
+                                        .push((name.clone(), *id, imported_item.span))
+                                }
+                            }
+                        }
```
AndrewBregger, line 1816: Might be overkill but this seemed to work without copying of the module.
```rust
let import_scope_id = ScopeId(imported_module_id, 0);

      for imported_name in &import.import_list {
          if let Some(id) = project.find_function_in_scope(
              import_scope_id,
              imported_name.name.as_str(),
          ) {
              if let Err(e) = project.add_function_to_scope(
                  scope_id,
                  imported_name.name.clone(),
                  id,
                  imported_name.span,
              ) {
                  return Some(e);
              }
          } else if let Some(id) = project
              .find_enum_in_scope(import_scope_id, imported_name.name.as_str())
          {
              if let Err(e) = project.add_enum_to_scope(
                  scope_id,
                  imported_name.name.clone(),
                  id,
                  imported_name.span,
              ) {
                  return Some(e);
              }
          } else if let Some(id) = project
              .find_type_in_scope(import_scope_id, imported_name.name.as_str())
          {
              if let Err(e) = project.add_type_to_scope(
                  scope_id,
                  imported_name.name.clone(),
                  id,
                  imported_name.span,
              ) {
                  return Some(e);
              }
          } else if let Some(id) = project
              .find_struct_in_scope(import_scope_id, imported_name.name.as_str())
          {
              if let Err(e) = project.add_struct_to_scope(
                  scope_id,
                  imported_name.name.clone(),
                  id,
                  imported_name.span,
              ) {
                  return Some(e);
              }
          }
      }
  }
```

PR #557, opened 2022-06-06
selfhost: Prevent infinite looping on error reporting

mateusauler:
Increment the token index when finding an unexpected token on operator, struct/class and function parsing.

To prevent spamming error messages, we now have a flag to know whether  error has already been reported in the current state.
This is a hack to mimic the way the bootstrap compiler works and isn't implemented everywhere it should.

This patch also changes some class/struct error messages to display the correct declaration type.

Also, should we report every error we encounter when running in "language server" mode?

### DIFF
--- a/selfhost/parser.jakt
+++ b/selfhost/parser.jakt
@@ -773,6 +773,9 @@ struct Parser {
         mut last_visibility: Visibility? = None
         mut last_visibility_span: Span? = None
 
+        // Have we already found an error?
+        mut error = false;
+
         while not .eof() {
             let token = .current()
             match token {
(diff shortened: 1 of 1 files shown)


selfhost/parser.jakt
```diff
@@ -1389,6 +1407,9 @@ struct Parser {
 
         let lhs = .parse_operand()
         expr_stack.push(lhs)
+        if lhs is Garbage {
+            .error("Unsupported expression", lhs.span())
+        }
```
jntrnr, line 1412: I think we may just want to give an error in `parse_operand_base` when we first create the garbage rather than detecting Garbage later and giving an error then.

selfhost/parser.jakt
```diff
@@ -1402,6 +1423,9 @@ struct Parser {
             .skip_newlines()
 
             let rhs = .parse_operand()
+            if rhs is Garbage {
```
jntrnr, line 1426: same here too

selfhost/parser.jakt
mateusauler: Yeah, I agree it makes more sense. Fixed.
Maybe the error could bubble up and be handled more gracefully by the caller, who then could provide more information?

jntrnr:
thanks

mateusauler:
Done! :^)

PR #559, opened 2022-06-06
selfhost: lexer: add token for `namespace` keyword

charliemirabile:
This token did not exist, so the `namespace` keword in
samples/namespaces/hello_namespace.jakt was parsed as an identifier.
Adding this token and adding it to the list of keywords fixes this.


charliemirabile:
Fixed! Sorry about the mixup with naming

PR #568, opened 2022-06-06
selfhost: Resolve compiler hang when parsing type

### LINKED ISSUE #561: Selfhosted compiler hangs on fat-arrow function
This makes the compiler hang:
```
class C {
    function f() =>  1u32 << 1 
}
```
Removing the `u32` part makes it pass.

mattisboeckle:
Resolves #561

### DIFF
--- a/selfhost/main.jakt
+++ b/selfhost/main.jakt
@@ -1371,7 +1371,10 @@ struct Parser {
                     index: index_expr,
                     span: merge_spans(start, .current().span()))
             }
-            else => expr
+            else => {
+                .index++
+                return expr
+            }
         }
     }
 


awesomekling:
Identifiers are not postfix operators, so something with this patch isn't right :)

mattisboeckle:
Right... I was checking the implementation in the rust parser and probably looked at the wrong lines.
This should be better

awesomekling:
If there is no postfix operator, that doesn't mean we should just advance the parser. It just means that there's no postfix operator.
The real problem in #561 is that the selfhost lexer doesn't yet recognize numeric literal suffixes like `u32`.

PR #574, opened 2022-06-06
editors/vscode: Add working file parent directory to include path

154852:
Adds the path of the working file to the include path when invoking
the compiler. Without this the compiler fails to resolve imports as
the file it is invoked on is in a temporary directory, without any
of the other source files present.


jntrnr:
thanks!

PR #585, opened 2022-06-07
selfhost: Add basic CLI to selfhosted compiler

charliemirabile:
You can now re-enable dumping the tokens from the lexer, using the `-l`
flag, and for consistency's sake the dumping of the final parsed
namespace is also toggled by a flag (`-p`).

If neither option is passed, then the compiler simply prints nothing and
returns zero if it was successful, or prints the errors and returns one if
there were issues with the code.


jntrnr:
thanks :)

PR #592, opened 2022-06-07
selfhost: Clean up some matches to use simpler syntax

jntrnr:
Cleans up a few places in the parser and typechecker to use less right-ward drift and the simpler enum match syntax


alimpfard:
Nice!

PR #595, opened 2022-06-07
selfhost/parser: Support comma in import list

cybergsus:
Quick change to make parser recognize a comma after an import list name.


cybergsus:
Was already implemented :ninyak:

PR #598, opened 2022-06-08
editor/vscode: Fix collition of generic argument opening and any `<`

horenso:
image

This small change fixes a syntax highlighting issue in vscode. As shown in the image, the second function keyword previously didn't get highlighted correctly. This happened because with the rule `(?<=function.*?)(<)` the gerneric-aguments list started at the operator `<`, since we match everything with the `.` up until this point. There is no `>`, so we also never exit the (non-existent) generic argument list.

### DIFF
--- a/editors/vscode/syntaxes/jakt.tmLanguage.json
+++ b/editors/vscode/syntaxes/jakt.tmLanguage.json
@@ -492,7 +492,7 @@
         },
         {
           "name": "meta.function.generic-arguments.jakt",
-          "begin": "(?<=function.*?)(<)",
+          "begin": "(?<=function\\s+(\\w)*?)(<)",
           "beginCaptures": {
             "1": {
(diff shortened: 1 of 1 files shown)


editors/vscode/syntaxes/jakt.tmLanguage.json
```diff
@@ -492,7 +492,7 @@
         },
         {
           "name": "meta.function.generic-arguments.jakt",
-          "begin": "(?<=function.*?)(<)",
+          "begin": "(?<=function\\s+(\\w|_)*?)(<)",
```
alimpfard, line 495: `\w` includes `_`.

alimpfard:
Also commit message typo: "collition" -> "collision"

horenso:
Thanks, good catch. I fixed the typo and removed the `_`.

PR #600, opened 2022-06-08
selfhost/codegen: Start fleshing out some of the codegen logic

awesomekling:


selfhost/codegen.jakt
```diff
@@ -0,0 +1,177 @@
...
+            Throw(expression) => {
+                return "return " + .codegen_expression(expression) + ";"
+            }   
+            Continue => {
+                if .control_flow_state.passes_through_match {
+                    return "return JaktInternal::LoopContinue{};"
+                }
+                return "continue;"
+            }
+            Continue => {
```
jntrnr, line 115: You've got `Continue` twice

alimpfard: this should definitely make the typechecker get mad at you, how did I manage to miss making this an error...?

awesomekling: That I do!

PR #602, opened 2022-06-08
selfhost: Add `error.jakt` file and move error code there

charliemirabile:
Similar to `error.rs` in the boostrap compiler `error.jakt` contains
the code for the `JaktError` enum and for printing error messages.

This code was cluttering up the top of `lexer.jakt` so it is much
nicer to have a proper home for it.

Note: this did require moving the `Span` struct out of the lexer.
Since the error code and lexer both need the span, and the lexer
also needs the error code, `Span` was moved into `utility.jakt`
at least for the time being until we are able to handle circular
imports. Not completely ideal, but I think the improvements to
`lexer.jakt` are worth it.

### DIFF
--- a/selfhost/error.jakt
+++ b/selfhost/error.jakt
@@ -1,0 +1,124 @@
+import utility { Span }
+
+enum JaktError {
+    Message(message: String, span: Span)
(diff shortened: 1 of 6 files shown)


jntrnr:
looks good!

PR #604, opened 2022-06-08
selfhost/lexer: Clarify `next` skip whitespace loop

charliemirabile:
Adding a small helper function to check for whitespace characters
allows making the whitespace skipping loop in the `next` function
extremely clear in function / very idiomatic to read.


selfhost/lexer.jakt
```diff
@@ -17,6 +17,7 @@ function is_ascii_alpha(anon c: u8) -> bool => (c >= b'a' and c <= b'z') or (c >
 function is_ascii_digit(anon c: u8) -> bool => (c >= b'0' and c <= b'9')
 function is_ascii_hexdigit(anon c: u8) -> bool => (c >= b'0' and c <= b'9') or (c >= b'a' and c <= b'f') or (c >= b'A' and c <= b'F')
 function is_ascii_alphanumeric(anon c: u8) -> bool => is_ascii_alpha(c) or is_ascii_digit(c)
+function is_ascii_whitespace(anon c: u8) -> bool => (c == b' ' or c == b'\t' or c == b'\r')
```
awesomekling, line 20: I don't think this name is appropriate. `\n` is also ASCII whitespace, but has special significance in the lexer. `\f` and `\v` are typically also considered whitespace.
Ultimately I think we're better off without this change.

PR #607, opened 2022-06-08
selfhost/parser: Unify classes structs and enums as `ParsedRecord` type

charliemirabile:
Since classes structs and enums all have a `name`, a `name_span`,
`generic_parameters`, `methods`, etc. and all represent user defined
types it makes sense for them to be combined into a single type to
reduce code duplication.

This change gets rid of `ParsedEnum` and `ParsedStruct` and replaces
both of them with:
```
struct ParsedRecord {
    name: String
    name_span: Span
    generic_parameters: [[String:Span]]
    definition_linkage: DefinitionLinkage
    methods: [ParsedMethod]
    record_type: RecordType
}
```
where `record_type` is one of these:
```
enum RecordType {
    Struct(fields: [ParsedField])
    Class(fields: [ParsedField], super_class: ParsedType?)
    ValueEnum(underlying_type: ParsedType, variants:[ValueEnumVariant])
    SumEnum(is_boxed: bool, variants: [SumEnumVariant])
    Garbage
}
```
Splitting structs and classes here allows for including space for a
base class for inheritence on class definition instances, and splitting
enums into their two main kinds (value enums with an underlying type
ala C/C++ and sum-type enums ala OCaml) at parse time facilitates
better error messages for the user and clearer code in the parser.

### DIFF
--- a/selfhost/parser.jakt
+++ b/selfhost/parser.jakt
@@ -15,11 +15,6 @@ enum DefinitionLinkage {
     External
 }
 
-enum DefinitionType {
-    Class
-    Struct
-}
-
 struct ImportName {
     name: String
     span: Span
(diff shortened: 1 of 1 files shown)


charliemirabile:
looks like one of the lines in the code block is 73 characters. I could force push and change that, but it probably would be worse for readability.

charliemirabile:
good catch / suggestion. fixed

PR #610, opened 2022-06-08
selfhost+parser: Connect typechecker to selfhost

jntrnr:
This connects the typechecker to the selfhost (with a couple workarounds for C++ codegen limitations)

It also fixes a couple infinite loops in the rust parser. As well as fixing codegen so that it was deterministic (using BTree* containers instead of Hash* containers)

Closes #613
(another attempt at fixing codegen to be deterministic)


awesomekling:
Could you put the rust-based compiler fixes for infinite looping in a separate commit? :^)

PR #620, opened 2022-06-09
selfhost/lexer: Support escape character in byte literal

lanmonster:
This patch allows for `b'\n'` to be lexed as SingleQuoteByteString(quote: "\n", ...).
Previously, it was reading only the backslash and skipping the rest.

### DIFF
--- a/selfhost/lexer.jakt
+++ b/selfhost/lexer.jakt
@@ -351,6 +351,10 @@ struct Lexer {
         // Everything but the quotes
         mut builder = StringBuilder::create()
         builder.append(.input[start + 1])
+        if .input[start + 1] == b'\\' {
+            builder.append(.input[start + 2])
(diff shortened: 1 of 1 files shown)


selfhost/lexer.jakt
```diff
@@ -350,10 +350,14 @@ struct Lexer {
 
         // Everything but the quotes
         mut builder = StringBuilder::create()
-        builder.append(.input[start + 1])
+        mut i = 1uz
+        while .input[start + i] != b'\'' {
```
awesomekling, line 354: This looks unbounded. What if we're at the end of the input?

lanmonster: decided against the loop because byte literals should be either 1 character, or 1 backslash and 1 character.

PR #627, opened 2022-06-10
samples: Greatly improve json sample and read_all sample app

u9g:
read_all sample is a reference as to how to get a file as a string that I made while using the function for reading a json file for testing


samples/apps/read_all.jakt
```diff
@@ -0,0 +1,17 @@
...
+    mut builder = StringBuilder::create()
+    for b in file.read_all().iterator() {
+        builder.append(b)
+    }
+    return builder.to_string()
+}
+
+function main(args: [String]) throws {
+    // replace "mystery.txt" with args[1] to be able to run "./readall mystery.txt"
+    let file_data = parse_to_string(file_path: "mystery.txt")
```
AtkinsSJ, line 15: You accidentally named the data file `mystory.txt` instead of `mystery.txt`, so this will always fail. You probably want to run your tests before submitting a PR.

u9g: fixed

samples/apps/json.jakt
```diff
@@ -223,37 +245,95 @@ class JsonParser {
         return JsonValue::Object(values)
     }
 
+    function char_to_f64(anon num: u8) throws -> f64 {
+        // FIXME 1: Shouldn't need this function at all
```
mjz19910, line 249: The first fixme can be put before the function definition, the second one can be put at the return statement that shouldn't be necssary

samples/apps/json.jakt
```diff
@@ -223,37 +245,95 @@ class JsonParser {
...
             }
-            value *= 10
-            value += (ch - b'0') as! i64
+
+            if not decimal_start_index.has_value() {
+                value *= 10.0
+                value += char_to_f64(ch - b'0')
+            } else {
+                mut num = char_to_f64(ch - b'0')
+                // FIXME: This should really be: `value += pow(10, -decimal_place)*num`, but: there's no pow function and you can't multiply float by usize
```
mjz19910, line 295: This is a Fixme: no pow() fn, and can't convert to float (or maybe you can)

mjz19910, line 295: FIXME: use pow function when implemented

mjz19910: If you can't cast usize to f64 then add another fixme `FIXME: can't cast to float from usize`

Issue #452, opened 2022-05-31
Swift-style `guard`


thislooksfun:
Swift has an (in my opinion) excellent feature called `guard` (can't link directly to it, but it's on [this page](https://docs.swift.org/swift-book/ReferenceManual/Statements.html)), which is implemented as an `if not` block that ***must*** exit the current scope (compiler error if you forget). It improves readability since you can define what you *do* want the state to be, rather than what you *don't*, and it prevents bugs by not letting you forget to `return` (or `throw`, or `break`, or `continue`). The code @awesomekling wrote in the recent Jakt-native JSON parser would be (imo) a lot cleaner with the use of a `guard`-like syntax.

awesomekling:
Yes, I think the `guard` thing in Swift is neat and we should totally have it.

PR #628, opened 2022-06-10
Unless keyword

lanmonster:
`unless` is syntax sugar for `if not` borrowed from ruby. `else` is disallowed for `unless` blocks because it is a double negative, and `if ... else` is easier to understand in that case. Ruby allows it, but general sentiment from [this stackoverflow question](https://stackoverflow.com/questions/38627935/do-we-use-else-with-unless-statement) seems to be that it should be avoided.

### DIFF
--- a/editors/emacs/jakt_mode.el
+++ b/editors/emacs/jakt_mode.el
@@ -39,7 +39,7 @@
 
 (setq jakt-font-lock-defaults
       (let* (
-             (jakt-conditional '("if" "else" "match"))
+             (jakt-conditional '("if" "else" "match" "unless"))
              (jakt-repeat '("while" "for" "loop"))
              (jakt-execution '("return" "break" "continue" "throw" "yield"))
              (jakt-boolean '("true" "false"))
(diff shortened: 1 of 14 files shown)


jntrnr:
@lanmonster - thanks for putting the work into this PR, as it's clear from your thoroughness. Before we start landing new features, we still need to finish selfhost. It's not yet ready for any experimentation until we know it's able to compile itself and other jakt files.

We'd also like to take some time to use jakt as a language on Serenity projects to make sure that we understand the trade-offs of the language in its current forms, and see where there are gaps in the language for writing Serenity apps.

After that point is a good time to try to fix those gaps with proposed new features.

We'd be more than happy to have you help with getting selfhost to ready. There's still a ton of porting of the Rust codebase to the selfhost codebase that needs to be done.

charliemirabile:
see also issue #452. `guard` is a very similar syntax sugar, but I think more clear in purpose (for checking preconditions, or other early fail return conditions) because of its requirement to exit scope. Given the comments there, it seems like it already has more traction, and I doubt we need both.

PR #645, opened 2022-06-13
selfhost/codegen: Implement codegen_namespace_path

mattisboeckle:


selfhost/codegen.jakt
```diff
@@ -339,8 +339,25 @@ struct CodeGenerator {
     }
 
     function codegen_namespace_path(this, call: CheckedCall) throws -> String {
-        // FIXME: Implement
-        return ""
+        mut output = ""
+
+        for namespace in call.namespace_.iterator() {
+            output += namespace.name
```
AtkinsSJ, line 345: This is missing the check for constructors that's included in the bootstrap compiler. We're still outputting C++ so it still needs to be included for now.

AtkinsSJ (changes requested):

selfhost/codegen.jakt
mattisboeckle: Sure. I pushed the updated version

PR #657, opened 2022-06-15
selfhost/typechecker: Reverse logic for is_static

lanmonster:
A function is considered static if the first parameter is not "this"
or "mut this". The previous logic was inverted. Oops!


awesomekling:
Oopsie!

PR #660, opened 2022-06-16
Some IDE things, Some selfhost, and a little bit of return; sauce

alimpfard:
The `else` thing:
image

The rest are just niceties to make the last commit possible :P


selfhost/codegen.jakt
```diff
@@ -182,6 +182,33 @@ struct CodeGenerator {
...
+                | U64(val)
+                | F32(val)
+                | F64(val)
+                | I64(val)
+                | I16(val)
+                | U16(val)
+                | USize(val)
+                | U32(val)
+                => format("static_cast<{}>({}{})", type_name, val, suffix)
+            }
```
alimpfard, line 210: Rust: "what does this even _mean_?"

PR #661, opened 2022-06-16
selfhost/typechecker: Expand type checking of variable declarations

greytdepression:
I expanded the existing type checking of variable declarations in the selfhosted compiler to get closer to functionality of the rust compiler.
The checking of builtin types is still missing and will require additional functions.

This is my first pull request, so I'd appreciate all the feedback I can get :^)


awesomekling:
Looks solid! :^)

PR #662, opened 2022-06-16
Fix helper function error

lanmonster:
`CheckedFunction`'s method `is_static` did not know about the members of `CheckedParameter` because `CheckedParameter` was defined after `CheckedFunction`. Ideally the order would not matter, but this will get the error in the editor to go away for now.


awesomekling:
Let's the compiler instead. :)
Doing that in #664

PR #670, opened 2022-06-16
Check and generate If statements

lanmonster:
With this change, we pass all of the if samples! :^)


awesomekling:
Dropped the second commit since it overlapped with Grey's PR

PR #674, opened 2022-06-17
selfhost/typechecker: Implement try_to_promote_constant_expr_to_type

greytdepression:


selfhost/typechecker.jakt
```diff
@@ -282,11 +282,7 @@ boxed enum Type {
         I32 => -2147483648
         // FIXME: this generates a warning in the C++ code
         I64 => -9223372036854775808
-        U8 => 0
-        U16 => 0
-        U32 => 0
-        U64 => 0
-        Usize => 0
+        U8 | U16 | U32 | U64 | Usize => 0
```
FireFox317, line 285: You could squash this commit into the first commit, since you could just as well do this in the first commit :)

greytdepression: Yeah that would be neater, but I honestly have no idea how to do that ^^

AtkinsSJ: https://www.youtube.com/watch?v=ElRzTuYln0M is a good guide to that. :^) If you're anything like me you'll make a big mess a couple of times when you're learning it, but it does get easier.

FireFox317: There is a helpful video that explains the `git` magic (It's really not magic tho) that you have to do: https://youtu.be/ElRzTuYln0M

Basically you wan't to do an interactive rebase, and then you can move commits around and choose to squash one commit into another. After you have done that, you can force push to the branch and it will be updated :)

greytdepression: Okay, I think I got it! Thanks, @FireFox317 @AtkinsSJ :^)

greytdepression:
I linked [issue 667](https://github.com/SerenityOS/jakt/issues/677) in the FIXME for the promote() function

PR #678, opened 2022-06-17
meta: Add an LL1 description of the language

Hendiadyoin1:
This description might not be 100% accurate yet, but it is a start to
have some sort of formal documentation for the syntax.

The file name, and location dont seem ideal to me, suggestions welcome.
If I got anything wrong or forgot something, feel free to point that out

### DIFF
--- a/meta/language.LL1
+++ b/meta/language.LL1
@@ -1,0 +1,185 @@
+// About this document
+// This is a EBNF *like* description of the Jakt Programming Language
+//
+// EBNF/LL1 dialect:
+// * Blocks are defined by an identifier and an equals, followed by the definition
+// * Literal Characters are surrounded by quotation-marks, preferably double-quotes
+// * Regex-literals matches are surrounded by forward-slashes
+// * Optional groups are surrounded by square-brackets
+// * Repeated groups are surrounded by curly-braces
+// * Groupings are surrounded by round-parentheses
+// * Comments are denoted by a double-slash in-front of them
+// * Multiple possible matches are separated by a vertical-bar/or-sign
+// * FIXME: This currently relies on left-side-recursion, which is not valid in LL(k)
+//
+// The most outer scope checks for `Declaration`s
+//
+// This is not yet as fine grained as it could be and does not differentiate
+// between some groups, so this is not (yet) a full guide to tokenization
+// Also the order of definitions and notation is currently not fully consistent
+
+// Stolen from https://webidl.spec.whatwg.org/#prod-i and adjusted to match Jakt more closely
+integer     = /-?([1-9][0-9]*|0[Xx][0-9A-Fa-f]+|0[Oo][0-7]*|0)([ui](8|16|32|64)|uz)?/
+float       = /-?(([0-9]+\.[0-9]*|[0-9]*\.[0-9]+)([Ee][+-]?[0-9]+)?|[0-9]+[Ee][+-]?[0-9]+)(f(64|32))?/
+            | /-?([1-9][0-9]*|0)f(64|32)/
+identifier  = /_?[A-Za-z][0-9A-Z_a-z-]*/
+
+String      = '"' {EscapedChar} '"'
+ByteLiteral = "b'" EscapedChar "'"
+CChar       = "'" EscapedChar "'"
+
+EscapedChar = /([^\\]|\\.)/
+
+boolean_op = /[<>]|[<>=!]=/
+
+op  = /[+-*\/%|^&]/
+    | /[&|+-*\/]=/
+    | boolean_op
+    | /\?\?=?/ 
+lop = /++|--?|[*~]|/
+    | "not"
+    | "&raw"
+rop = /++|--/
+    | ("[" Expression "]")
+
+Constant         = integer
+                 | float
+                 | String
+                 | ByteLiteral
+                 | CChar
+                 | "None"
+                 | "true" | "false"
+
+BuiltinType      = /f(32|64)/
+                 | /[iu](8|16|32|64)/
+                 | "usize"
+                 | /c_(int|char)/
+                 | "bool"
+                 | "String"
+                 | "void"
+
+TypeExpression   = BuiltinType ["?"]
+                 | {identifier [GenericSpecifier] "::" } identifier [GenericSpecifier] ["?"]
+                 | "[" TypeExpression "]" ["?"]
+                 | "[" TypeExpression ":" TypeExpression "]" ["?"]
+                 | "{ TypeExpression "}" ["?"]
+                 | "raw" TypeExpression  ["?"]
+                 | "weak" TypeExpression "?"
+
+GenericSpecifier = "<" TypeExpression {"," TypeExpression} ">"
+
+ArrayLiteral     = "[" [ArrayItem [ArrayItems]] "]"
+                 | "[" Expression ";" integer "]"
+ArrayItems       = "," ArrayItem [ArrayItems]
+ArrayItem        = Expression
+
+MapLiteral       = "[:]"
+                 | "[" MapItem [MapItems] "]"
+MapItems         = "," MapItem [MapItems]
+MapItem          = Expression ":" Expression
+
+SetLiteral       = "{" [SetItem [SetItems]] "}"
+SetItems         = "," SetItem [SetItems]
+SetItem          = Expression
+
+Range            = Expression ".." Expression
+
+MatchExpression  = "match" Expression "{" {MatchCase} "}"
+MatchCase        = MatchKey "=>" Statement
+MatchKey         = MatchKey "|" MatchKey
+                 | {identifier "::"} identifier ["(" {identifier [":" identifier]} ")"]
+                 | Constant
+                 | "else"
+
+// FIXME: The second operand can be an Enum-Variant as well, but that is coincidentally
+//        also handled by the  TypeExpression
+IsExpression     = Expression "is" TypeExpression
+CastExpression   = Expression /as[!?]/ TypeExpression
+
+// FIXME: Generics are currently only allowed on the last part of the Expression
+Expression       = Expression op Expression
+                 | lop Expression
+                 | Expression rop
+                 | "(" Expression ")"
+                 | Constant
+                 | ArrayLiteral
+                 | MapLiteral
+                 | SetLiteral
+                 | Range
+                 | MatchExpression
+                 | IsExpression
+                 | CastExpression
+                 | identifier [GenericSpecifier] {"::" identifier [GenericSpecifier]}
+                 | [Expression] /!?\./ identifier [GenericSpecifier]
+                 | Expression "(" ArgumentList ")"
+
+Statement        = Expression  [";"]
+                 | DeclarationStatement
+                 | AssignmentStatement
+                 | ReturnStatement
+                 | YieldStatement
+                 | BlockStatement
+                 | ThrowStatement
+                 | IfStatement
+                 | TryStatement
+                 | ForStatement
+                 | LoopStatement
+                 | WhileStatement
+                 | "defer" Statement
+                 | "unsafe" BlockStatement
+                 | "cpp" "{" {String} "}"
+                 | "continue"
+                 | "break"
+
+DeclarationStatement = /let|mut/ identifier [":" TypeExpression] "=" Expression [";"]
+AssignmentStatement  = identifier "=" Expression [";"]
+ReturnStatement      = "return" [Expression] [";"]
+YieldStatement       = "yield" [Expression] [";"]
+ThrowStatement       = "throw" [Expression] [";"]
+
+BlockStatement       = "{" Statement "}"
+
+IfStatement          = "if" BooleanExpression BlockStatement ["else" BlockStatement]
+TryStatement         = "try" (Expression | BlockStatement) "catch" identifier BlockStatement
+ForStatement         = "for" identifier "in" Expression BlockStatement
+LoopStatement        = "loop" BlockStatement
+WhileStatement       = "while" Expression BlockStatement
+
+// NOTE: `mut this` is only allowed in ClassFunctions
+ParameterList   = ["mut"] "this" [Parameters]
+                | Parameter [Parameters]
+                |
+Parameters      = "," Parameter [Parameters]
+Parameter       = ["anon"] ["mut"] identifier ":" TypeExpression
+
+ArgumentList  = [Argument [Arguments]]^^
+Arguments     = "," Argument [Arguments]
+Argument      = [identifier ":"] Expression
+
+Declaration    = Comment
+               | Namespace
+               | Import
+               | Function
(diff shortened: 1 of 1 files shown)


meta/language.LL1
```diff
@@ -0,0 +1,147 @@
...
+// * Multiple possible matches are separated by a vertical-bar/or-sign
+//
+// The most outer scope checks for `Declaration`s
+//
+// This is not yet as fine grained as it could be and does not differentiate
+// between some groups, so this is not (yet) a full guide to tokenization
+// Also the order of definitions and notation is currently not fully consistent
+
+// Stolen from https://webidl.spec.whatwg.org/#prod-i and adjusted to match Jakt more closely
+integer     = /-?([1-9][0-9]*|0[Xx][0-9A-Fa-f]+|0[Oo][0-7]*|0)([ui](8|16|32|64))?/
```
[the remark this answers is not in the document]
Hendiadyoin1: is `z` valid as well?

meta/language.LL1
```diff
@@ -0,0 +1,147 @@
...
+Comment        = /\/\/.*\n/
+Import         = "import" identifier ["{" identifier {"," identifier} "}"]
+Function       = ["extern"] "function" identifier [GenericSpecifier] "(" ArgumentList ")" ["->" TypeExpression] ["throws"] [("{" {Statement} "}") | ("=>" Expression)]
+
+// NOTE: This describes Classes and Structs
+Class          = ["extern"] ("class"|"struct") identifier "{" {ClassMembers} "}"
+ClassMembers   = ClassFunction | ClassMember
+ClassMember    = [AccessModifier ["mut"]] identifier ":" TypeExpression
+ClassFunction  = [AccessModifier] "function" identifier [GenericSpecifier] "(" ArgumentList ")" ["->" TypeExpression] ["throws"] (("{" {Statement} "}") | ("=>" Expression))
+AccessModifier = "public"
```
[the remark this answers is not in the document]
Hendiadyoin1: Sure missed that

meta/language.LL1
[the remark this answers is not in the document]
alimpfard: I don't think so, `uz` is `usize`, I don't think we have `size`.

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+// The most outer scope checks for `Declaration`s
+//
+// This is not yet as fine grained as it could be and does not differentiate
+// between some groups, so this is not (yet) a full guide to tokenization
+// Also the order of definitions and notation is currently not fully consistent
+
+// Stolen from https://webidl.spec.whatwg.org/#prod-i and adjusted to match Jakt more closely
+integer     = /-?([1-9][0-9]*|0[Xx][0-9A-Fa-f]+|0[Oo][0-7]*|0)([ui](8|16|32|64)|uz)?/
+float       = /-?(([0-9]+\.[0-9]*|[0-9]*\.[0-9]+)([Ee][+-]?[0-9]+)?|[0-9]+[Ee][+-]?[0-9]+)(f64)?/
+            | /-?([1-9][0-9]*|0[Xx][0-9A-Fa-f]+|0[Oo][0-7]*|0)f64/
```
AtkinsSJ, line 23: f32 is also a suffix: https://github.com/SerenityOS/jakt/blob/a0fd770771c134149f8b11605d295280207eb519/src/lexer.rs#L583

I think you asked about `z`, and that's not listed. (I thought it was one but I don't have the best memory.)

Hendiadyoin1: Thanks
Added that aswell

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+// NOTE: This describes Classes and Structs
+Class          = ["extern"] ("class"|"struct") identifier "{" {ClassMembers} "}"
+ClassMembers   = ClassFunction | ClassMember
+ClassMember    = [AccessModifier ["mut"]] identifier ":" TypeExpression
+ClassFunction  = [AccessModifier] "function" identifier [GenericSpecifier] "(" ArgumentList ")" ["->" TypeExpression] ["throws"] (("{" {Statement} "}") | ("=>" Expression))
+AccessModifier = "public"
+               | "private"
+               | "restricted" "(" identifier [GenericSpecifier] {"," identifier [GenericSpecifier]} ")"
+
+Enum           = ["boxed"] "enum" identifier [GenericSpecifier] [":" TypeExpression] "{" EnumMember {EnumMember} "}"
```
alimpfard, line 175: `GenericSpecifier` and `":" TypeExpression` are mutually exclusive (and so is `"boxed"` with `":" TypeExpression`).

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+EnumMember     = ClassFunction
+               | identifier ["(" [identifier ":"] TypeExpression {"," [identifier ":"] TypeExpression} ")"]
```
alimpfard, line 177: Given an enum with an underlying type, the members can also be `identifier ["=" Expression]`.

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+Parameter      = [identifier ":"] Expression
+
+Declaration    = Comment
+               | Namespace
+               | Import
+               | Function
+               | Class
+               | Enum
+
+Comment        = /\/\/.*\n/
```
alimpfard, line 161: This won't match comments at the end of the file.

```suggestion
Comment        = /\/\/.*$/
```

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+
+IfStatement          = "if" BooleanExpression BlockStatement ["else" BlockStatement]
+TryStatement         = "try" (Expression | BlockStatement) "catch" identifier BlockStatement
+ForStatement         = "for" identifier "in" Expression BlockStatement
+LoopStatement        = "loop" BlockStatement
+WhileStatement       = "while" Expression BlockStatement
+
+// NOTE: `mut this` is only allowed in ClassFunctions
+ArgumentList   = [[["mut"] "this"] [Arguments]]
+Arguments      = "," Argument [Arguments]
```
alimpfard, line 147: This allows `foo(, bar: T)`

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+Argument       = ["anon"] ["mut"] identifier [":" TypeExpression]
```
alimpfard, line 148: I'm pretty sure the type annotation is required here

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+BlockStatement       = "{" Statement "}"
+
+IfStatement          = "if" BooleanExpression BlockStatement ["else" BlockStatement]
+TryStatement         = "try" (Expression | BlockStatement) "catch" identifier BlockStatement
+ForStatement         = "for" identifier "in" Expression BlockStatement
+LoopStatement        = "loop" BlockStatement
+WhileStatement       = "while" Expression BlockStatement
+
+// NOTE: `mut this` is only allowed in ClassFunctions
+ArgumentList   = [[["mut"] "this"] [Arguments]]
```
alimpfard, line 146: This is a Parameter...

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+Arguments      = "," Argument [Arguments]
+Argument       = ["anon"] ["mut"] identifier [":" TypeExpression]
+
+ParameterList  = [Parameter [Parameters]]
```
alimpfard, line 150: ...and this is an Argument.

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+// Also the order of definitions and notation is currently not fully consistent
+
+// Stolen from https://webidl.spec.whatwg.org/#prod-i and adjusted to match Jakt more closely
+integer     = /-?([1-9][0-9]*|0[Xx][0-9A-Fa-f]+|0[Oo][0-7]*|0)([ui](8|16|32|64)|uz)?/
+float       = /-?(([0-9]+\.[0-9]*|[0-9]*\.[0-9]+)([Ee][+-]?[0-9]+)?|[0-9]+[Ee][+-]?[0-9]+)(f(64|32))?/
+            | /-?([1-9][0-9]*|0)f(64|32)/
+identifier  = /_?[A-Za-z][0-9A-Z_a-z-]*/
+
+String      = '"' EscapedContent '"'
+ByteString = "b'" EscapedContent "'"
```
alimpfard, line 27: You mean a ByteLiteral? this thing can only contain a single byte in the ASCII range afaik

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+MatchCase        = MatchKey "=>" Statement
+MatchKey         = MatchKey "|" MatchKey
+                 | {identifier "::"} identifier ["(" {identifier [":" identifier]} ")"]
+                 | Constant
+                 | "else"
+
+IsExpression     = Expression "is" TypeExpression
+CastExpression   = Expression "as!" TypeExpression
+
+Expression       = Expression op Expression
```
alimpfard, line 97: `Expression "is" Expression` is not a valid construct (op => boolean_op => "is").

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+                 | lop Expression
+                 | Expression rop
```
alimpfard, line 99: Left recursion isn't valid in LL(k).

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+                 | "(" Expression ")"
+                 | Constant
+                 | ArrayLiteral
+                 | MapLiteral
+                 | SetLiteral
+                 | Range
```
alimpfard, line 105: This technically can't be here as it causes left recursion (there's tonnes of this, I won't bother with the rest).

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+
+MatchExpression  = "match" Expression "{" {MatchCase} "}"
+MatchCase        = MatchKey "=>" Statement
+MatchKey         = MatchKey "|" MatchKey
+                 | {identifier "::"} identifier ["(" {identifier [":" identifier]} ")"]
+                 | Constant
+                 | "else"
+
+IsExpression     = Expression "is" TypeExpression
+CastExpression   = Expression "as!" TypeExpression
```
alimpfard, line 95: There's also `as?`.

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+Comment        = /\/\/.*\n/
+Namespace      = "namespace" identifier "{" {Declaration} "}"
+Import         = "import" identifier ["{" identifier {"," identifier} "}"]
+Function       = ["extern"] "function" identifier [GenericSpecifier] "(" ArgumentList ")" ["->" TypeExpression] ["throws"] [("{" {Statement} "}") | ("=>" Expression)]
+
+// NOTE: This describes Classes and Structs
+Class          = ["extern"] ("class"|"struct") identifier "{" {ClassMembers} "}"
+ClassMembers   = ClassFunction | ClassMember
+ClassMember    = [AccessModifier ["mut"]] identifier ":" TypeExpression
+ClassFunction  = [AccessModifier] "function" identifier [GenericSpecifier] "(" ArgumentList ")" ["->" TypeExpression] ["throws"] (("{" {Statement} "}") | ("=>" Expression))
```
alimpfard, line 170: A function (class or not) is allowed to have no body.

meta/language.LL1
```diff
@@ -0,0 +1,178 @@
...
+AccessModifier = "public"
+               | "private"
+               | "restricted" "(" identifier [GenericSpecifier] {"," identifier [GenericSpecifier]} ")"
+
+Enum           = ["boxed"] "enum" identifier [GenericSpecifier] [":" TypeExpression] "{" EnumMember {EnumMember} "}"
+EnumMember     = ClassFunction
```
alimpfard, line 176: This currently doesn't accept `restricted(...)` (but it should in the future).

alimpfard (commented):
Sorry I took a while to review this ^^'
The left recursions can be left alone since this isn't really trying to conform to either ebnf or ll(k), but I figured I'd keep the comments.

meta/language.LL1
Hendiadyoin1: Ah ok
Wa confused with the variant-is, but the IsExpression should magically handle that one aswell

meta/language.LL1
Hendiadyoin1: I could not find a user of that in the current code base, so I was not sure if that exists

meta/language.LL1
Hendiadyoin1: wdym, an ellipsis in the restriction?

meta/language.LL1
Hendiadyoin1: Ok I currently dont really know how to fix that one, but I will add a note to the introduction

meta/language.LL1
alimpfard: No, I mean restricted (anything)
Currently only public/private is accepted there iirc

meta/language.LL1
```diff
@@ -0,0 +1,181 @@
...
+                 | ArrayLiteral
+                 | MapLiteral
+                 | SetLiteral
+                 | Range
+                 | MatchExpression
+                 | IsExpression
+                 | CastExpression
+                 | identifier {"::" identifier}
+                 | [Expression] /!?\./ identifier
+                 | Expression "(" ArgumentList ")"
```
Hendiadyoin1, line 111: I feel there are some missing `[GenericSpecifier]`s missing here
@alimpfard can you confirm that any identifier here can take a GenericSpecifier?

alimpfard: Only the call target can atm
```
Expression ["<" {Typename} ">"] "(" ... ")"
```

Hendiadyoin1: So something like `my_enum<i64>::myvalue` is not allowed currently?
Same for any of
```
MyClass<T>::static_function(...)
MyClass::static_function<T>(...)
generic_constant<T>
```
My guess would be that the GenericSpecifier belongs to the identifier and not to the Expression

alimpfard: Yep, that's currently not valid.

cybergsus:
You can't iterate over a Range<Range<_>> but, even if it sounds goofy, the syntax allows it: `(0..2)..(8..3)`. This makes no logical sense, but the parser can be made simpler and then reject any iteration that goes over that type

mnlrsn:
> 

> even if it sounds goofy, the syntax allows it

Sure, that was my point, too. The question is more if it *should* allow it...

cybergsus:
> Sure, that was my point, too. The question is more if it *should* allow it...

I don't feel like prohibiting this explicitly syntax-wise, since it's nature is more specific to the **type** *inside* of the range. That's OK to be parsed, a range is an expression. It will be rejected by typechecker. Disallowing this pattern explicitly in syntax only adds noise to the language spec and to the parser.

PR #685, opened 2022-06-18
selfhosted/lexer: Add integer literal type suffixes

### LINKED ISSUE #634: selfhosted: parse error in fat-arrow function
This snippet of code produces an error message:
```
function f() => 1u32 << 1 
```
The error message: 
```
1  | function f() => 1u32 << 1
                       ^- Unrecognized token in namespace (probably not implemented yet)
```
In comparison, this implementation parses without  error message:
```
function g() {
    return 1u32 << 1 
}
```

onsah:
Adds integer literal type support to lexer.

Also fixes #634

### DIFF
--- a/selfhost/lexer.jakt
+++ b/selfhost/lexer.jakt
@@ -393,11 +393,11 @@ struct Lexer {
 
         mut total: i64 = 0
 
-        if (.eof()) {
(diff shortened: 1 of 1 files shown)


onsah:
Seems this is already implemented in latest jakt

PR #689, opened 2022-06-18
README: Update usage instructions

Vantadaga:
Update usage instructions to improve readability


Vantadaga:
Is there anyone that could review this PR for me and give feedback if needed?

jntrnr:
@Vantadaga - we're moving away from the Rust compiler, so we may want to start from using the jakt (selfhost) instruction instead.

PR #693, opened 2022-06-19
tests: Make skip explicit

### LINKED ISSUE #666: Meta: Check that tests have valid test comments in CI
#665 fixes a couple of tests that were never being run because their comments didn't follow the correct format. It would be good to add a check to CI that makes sure that the format is followed, probably for all jakt files in `tests/` and `samples/`.

flostellbrink:
Fixes #666 

This addresses the problem where tests are implicitly skipped when they don't have expectations.
Tests will now fail if they are not marked as skipped and don't have any expectations.

We also test the selfhosted compiler. In that case we only want to test `main.jakt`.
Instead of marking all other selfhost files as `skip` I've changed the test to target `main.jakt` instead.

### DIFF
--- a/tests/main.rs
+++ b/tests/main.rs
@@ -180,12 +180,134 @@ fn find_results(
     }
 
     if output.is_none() && error.is_none() && stderr.is_none() {
-        return Err(JaktError::StringError(
-            "Expected output, error, or stderr".into(),
-        ));
+        panic!(
+            "\r\nTest: {:?}\r\nNot marked as skip and does not have expectations\r\n",
+            path
+        );
     }
 
     Ok((output, error, stderr))
+}
+
+fn test_sample(path: &PathBuf) -> Result<(), JaktError> {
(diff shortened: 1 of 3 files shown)


tests/main.rs
```diff
@@ -180,143 +180,145 @@ fn find_results(
...
+
+fn test_samples(path: &str) -> Result<(), JaktError> {
+    for sample in std::fs::read_dir(path).expect("if this fails, the repo or harness is broken") {
+        let sample = sample.expect("if this fails, the filesytem is having issues");
+
+        let path = sample.path();
+
+        if let Some(ext) = path.extension() {
+            if ext == "jakt" {
+                test_sample(&path);
```
AtkinsSJ, line 321: We're getting warnings on this line that the result is being ignored.

flostellbrink: Thanks for the teminder forgot to handle that, will update the PR later today.

PR #694, opened 2022-06-19
selfhost/typechecker: Sanitize root scope

greytdepression:
Make sure the scope ScopeId(0, 0) is created when
Typechecker::typecheck() is called and that it does not have a parent.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -65,6 +65,10 @@ struct InferenceId {
 struct ScopeId {
     module: ModuleId
     id: usize
+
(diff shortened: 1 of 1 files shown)


awesomekling:
This has build errors that were not visible before #695
Please rebase and fix them :)

greytdepression:
It should work now :^) @awesomekling

awesomekling:
Great, but please squash fixups instead of adding them as separate commits.

greytdepression:
Oops, sorry ^^
Still getting used to the git workflow

PR #701, opened 2022-06-20
selfhost/typechecker: Implement Check Function TODOs but 1

TophEvich:
Implement all the TODOs in typecheck_function but 1:
* resolve concrete types
* second return type pass to resolve generics
* infer return type (if necessary)
* report error if control reaches end of non-void function

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -1445,19 +1445,63 @@ struct Typechecker {
             .add_var_to_scope(scope_id: function_scope_id, name: variable.name, var_id, span: variable.definition_span)
         }
 
-        // TODO: Resolve concreete types
+        // Resolve concrete types
+        mut function_return_type_id = .typecheck_typename(
+            parsed_type: parsed_function.return_type
+            scope_id: function_scope_id
+        )
+        checked_function.return_type_id = function_return_type_id
 
         // TODO: Typecheck function block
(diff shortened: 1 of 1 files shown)


jntrnr:
@TophEvich - thanks for the PR

I was playing around with it a bit and noticed that it seems to add some type information that will break the unit tests. Here's one example:

```
Error: Control reaches end of non-void function
----- samples/variables/blocks_and_scopes.jakt:4:10
 3 |
 4 | function main() {
               ^- Control reaches end of non-void function
 5 |     let x = 10
-----
```

I'm guessing it might be coming from `main` getting an inferred return type for codegen but that return type being required even though main doesn't require a return.

jntrnr:
@TophEvich - I created that function with the hope we could do more sophisticated type inference. Having said that, I never really finished it. Probably the easiest way forward for now would be to port over the logic we have from the Rust compiler and ignore that helper function for now.

We can go back later and come up with better inference, but for now just getting the selfhost compiler up and able to compile the tests we have is probably more important.

TophEvich:
@jntrnr OK, so remove everything "Inference Type" and replace with TypecheckError and UNKNOWN_TYPE ?

TophEvich:
@jntrnr I created #725 to do just that

TophEvich:
I guess this one should now be good to go in?

PR #703, opened 2022-06-20
typechecker: Fix bug in typecheck_namespace_declarations()

greytdepression:
Due to a typo in this function we never checked unnamed child
namespaces.

### DIFF
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -2077,16 +2077,14 @@ pub fn typecheck_namespace_declarations(
     let mut error = None;
 
     for namespace in parsed_namespace.namespaces.iter() {
-        if let Some(namespace_name) = &namespace.name {
-            // Finish typecheck of the named namespaces
+        let namespace_scope_id = if let Some(namespace_name) = &namespace.name {
             let (namespace_scope_id, _) = project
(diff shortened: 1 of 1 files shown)


jntrnr:
@greytdepression are there any tests that fail before but pass now with this? If not, would you mind adding a test case so we can ensure we don't make a similar mistake in the future?

greytdepression:
@jntrnr Currently all tests pass. I'll try to make one that fails because of this

greytdepression:
@jntrnr So I tried making a failing test, but for that I needed an unnamed namespace that is the child of another namespace. And I don't think that's possible (at least not right now). I went through the parser and it seems like there is no way we could instantiate a child namespace without a name.
So this PR seems unnecessary, but also we might want to panic the program if we encounter a nameless namespace then?

greytdepression:
Okay, sorry for going back and forth on this so much. I don't know why but this entire thing kinda broke my brain :^)
Anyway, the following snippet produces a stack overflow, because it falls into infinite recursion due to the bug.
```jakt
namespace Foo {
    namespace {
        enum Bar {
            A
        }
    }
}

function main() {
    println("Well hello friend!")
}
```
Unfortunately, while the fixed version does not recurse infinitely, it also does not compile yet for a different reason.

PR #711, opened 2022-06-20
selfhost/typechecker: Typecheck structs

greytdepression:
I've been working on porting this for a while now :^)

I can't really check if everything is working as it is supposed to, as struct codegen is not implemented yet. But in order to work on codegen we need to typecheck first anyway.
This PR will conflict with #701 as both implement `resolve_type_var()`.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -2777,6 +2777,10 @@ struct Typechecker {
                 }
 
                 for i in 0..call.args.size() {
+                    if i == callee.params.size() {
+                        break
+                    }
+
                     let parsed_label_and_arg = call.args[i]
                     let param = callee.params[i]
 
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -2,7 +2,8 @@ import error { JaktError }
 import parser { BinaryOperator, DefinitionLinkage, DefinitionType,
                 FunctionLinkage, FunctionType, ParsedBlock, ParsedCall,
                 ParsedExpression, ParsedFunction, ParsedNamespace,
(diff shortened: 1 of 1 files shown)


greytdepression:
@jntrnr Yes, I noticed that. I didn't mention it in my comment, but I took a look at all the tests that pass with the current version of main and fail in my branch. Here are the 8 tests my branch fails that main passes
```
[ PASS ] samples/enums/empty_enum.jakt
[ PASS ] samples/enums/parse.jakt
[ PASS ] samples/weak/question_mark_required.jakt
[ PASS ] samples/weak/reject_immutable_weak.jakt
[ PASS ] tests/codegen/out_of_order_types.jakt
[ PASS ] tests/parser/enum_correct_typename_in_typed_variant.jakt
[ PASS ] tests/parser/unclosed_brace_in_enum.jakt
[ PASS ] tests/parser/unclosed_brace_in_ref_enum.jakt
```
and here's the 1 test my branch passes that main fails
```
[ PASS ] samples/namespaces/hello_namespace.jakt
```
The reason they pass in main is that they do not expect an output and since main does not try to typecheck structs or classes, it just acts like they aren't even there. So all such tests that expect no output think that main passes them, even though it does not correctly typecheck them. Whereas in my branch those tests will run into `todo(...)` calls in the typechecker and fail. :^)

jntrnr:
@greytdepression - ah okay, so they'll break but they're breaking because more is getting checked rather than ignored.

Sounds like we're good to land?

For the clash, do you want #723 to land first and then rebase this PR?

greytdepression:
@jntrnr Sure, sounds good! I'll just wait for #723 then

jntrnr:
should be in now

greytdepression:
@jntrnr Rebased and passing checks :^)

PR #716, opened 2022-06-21
selfhost/typechecker: Implement most of typecheck_block

TophEvich:
This Implements the last big TODO left over by #701 and adds most of the type-checking to typecheck_block that exists in the rust version of the compiler, only one of the support functions has an optimization TODO.

It also adds the following on the way
* the Match expression
* a function to check if an expression definitely returns
* a function to check if statement definitely returns
* a function to check types for their compatibility
* a function to substitute type-vars in types 

**Before:**
```
75 passed
257 failed
7 skipped
```

**After:**
```
76 passed
256 failed
7 skipped
```

**This PR and #701**
```
81 passed
251 failed
7 skipped
```

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -2012,6 +2012,37 @@ struct Typechecker {
         }
     }
 
+    function statement_definitely_returns(this, anon statement: CheckedStatement) -> bool => match statement{
+        Return(expr) => true
+        // FIXME: CheckedStatement::If branches ugly implementation due to current compiler limitations
+        If(guard, then_block, else_statement) => {
+            mut ret = false
+            let guard_result = match guard {
+                Boolean(val) => {
+                    if val and then_block.definitely_returns {
+                        ret = true
+                    }
+                    yield ret
+                }
+                else => false
+            }
+            if not guard_result and else_statement.has_value(){
+                if then_block.definitely_returns and .statement_definitely_returns(else_statement.value()) {
+                    ret = true
+                }
+            }
+            yield ret
+        }
+        Block(block) => block.definitely_returns
+        Loop(block) => block.definitely_returns
+        While(guard, block) => block.definitely_returns
+        Expression(match_expr) => match match_expr {
+            Match() => checked_expression_definitely_returns(match_expr)
+            else => false
+        }
+        else => false
+    }
+
     function typecheck_block(mut this, anon parsed_block: ParsedBlock, parent_scope_id: ScopeId, safety_mode: SafetyMode) throws -> CheckedBlock {
         let parent_throws = .get_scope(parent_scope_id).can_throw
         let block_scope_id = .create_scope(parent_scope_id, can_throw: parent_throws)
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -2399,26 +2399,54 @@ struct Typechecker {
             definitely_returns: false
             yielded_type: none_type_id
         )
+        // FIXME: Use [TypeId: TypeID] without TypeId.to_string()/TypeId::from_string() workaround
+        mut generic_inferences: [String: String] = [:]
 
         for parsed_statement in parsed_block.stmts.iterator() {
-            let statement = .typecheck_statement(parsed_statement, scope_id: block_scope_id, safety_mode)
-            
-            match statement {
-                Return(expr) => {
-                    checked_block.definitely_returns = true
-                    checked_block.yielded_type = match expr.has_value() {
-                        true => .expression_type(expr!)
-                        else => builtin(BuiltinType::Void)
-                    }
+
+            let checked_statement = .typecheck_statement(
+                statement: parsed_statement
+                scope_id: block_scope_id
+                safety_mode
+            )
+
+            // check that statment definitly returns
+            checked_block.definitely_returns = .statement_definitely_returns(checked_statement)
+
+            let none_yield: Span? = None // FIXME: Can't yield None directly
+            let yield_span: Span? = match parsed_statement {
+                ParsedStatement::Yield(expr) => Some(expr.span())
+                else => none_yield
+            }
+            let  none_expr: CheckedExpression? = None // FIXME: Can't yield None directly
+            let checked_yield_expression: CheckedExpression? = match checked_statement {
+                CheckedStatement::Yield(expr) => Some(expr)
+                else => none_expr
+            }
+            if yield_span.has_value() and checked_yield_expression.has_value() {
+                let type_var_type_id = .expression_type(checked_yield_expression.value())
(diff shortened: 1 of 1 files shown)


cybergsus:
Those errors come from either a compiler bug or a parse error that wasn't registered. A quick `dbg!` insertion in the Rust version reveals:
```
[src/typechecker.rs:5781] &structure = CheckedStruct {
    name: "CheckedBlock",
    generic_parameters: [],
    fields: [],
    (... rest of struct fields truncated)
}
```

selfhost/typechecker.jakt
```diff
@@ -1460,6 +1548,334 @@ struct Typechecker {
...
+                if seen_type_id_string.has_value() {
+                    if seen_type_id_string.value() != rhs_type_id_string {
+                        .error(
+                            format(
+                                "Type mismatch: expected `{}` but got `{}`"
+                                .type_name(TypeId::from_string(seen_type_id_string.value()))
+                                .type_name(rhs_type_id)
+                            )
+                            span
+                        )
```
greytdepression, line 1610: I won't write this under every `.error(...)`, but I think you have to explicitly `return false` or set `ret = false` after each error (personally I would remove `ret` and just `return true` at the end if the checks did not fail along the way).
The original Rust code used return type `Option<JaktError>`, so it was enough to emit an error anywhere a check failed, but since the Jakt function here returns a boolean, you have to return false in addition to emitting the error.

selfhost/typechecker.jakt
```diff
@@ -1460,6 +1548,334 @@ struct Typechecker {
...
+        }
+        return result
+    }
+
+    // FIXME: Use [TypeId: TypeID] without TypeId.to_string()/TypeId::from_string() workaround
+    function substitute_typevars_in_type_helper(mut this, type_id: TypeId , generic_inferences: [String: String]) throws -> TypeId {
+        let type_ = .get_type(type_id)
+
+        match type_ {
+            TypeVariable() => {
```
greytdepression, line 1825: You can just use `TypeVariable => {`

selfhost/typechecker.jakt
```diff
@@ -1460,6 +1548,334 @@ struct Typechecker {
...
+                for arg in new_args.iterator() {
+                    arg = .substitute_typevars_in_type(type_id: arg, generic_inferences)
+                }
+
+                return .find_or_add_type_id(Type::GenericInstance(id, args: new_args))
+            }
+            GenericEnumInstance(id, args) => {
+                mut new_args = args
+                for arg in new_args.iterator() {
+                    arg = .substitute_typevars_in_type(type_id: arg, generic_inferences)
```
greytdepression, line 1842: This does not actually change `args`. Try running this test:
```
struct TypeId {
    id: u64
}

function main() {
    mut args = [TypeId(id: 0); 2]

    for arg in args.iterator() {
        arg.id = 2
    }

    println("{}", args)
}
```
You'll see that `args` is still the same as before even though we changed `arg`. You need to index into `new_args` and assign to that. So something like `new_args[index] = .substitute_typevars_in_type(...)`.
The same holds applies to the `Struct` and `Enum` cases below :^)

selfhost/typechecker.jakt
```diff
@@ -1460,6 +1548,334 @@ struct Typechecker {
...
+                }
+            }
+            RawPtr(rawptr_type_id) => {
+                let rawptr_type = Type::RawPtr(
+                    .substitute_typevars_in_type(type_id: rawptr_type_id, generic_inferences)
+                )
+                return .find_or_add_type_id(rawptr_type)
+            }
+            else => {
+                return type_id
```
greytdepression, line 1873: You could just leave that blank (`else => {}`) as you'd return `type_id` immediately after anyway

greytdepression (commented):
I know this PR is still a draft, but I took a look already :)

selfhost/typechecker.jakt
```diff
@@ -1055,6 +1141,7 @@ struct Typechecker {
         OptionalNone(type_id) => type_id
         OptionalSome(type_id) => type_id
         ForcedUnwrap(type_id) => type_id
+        Match(expr, match_cases, span, type_id, all_variants_constant) => type_id
```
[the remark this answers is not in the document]
TophEvich: Hm, interesting I tried that previously but had problems with that where if the signature was not 100% the same (same naming an all params) it could not match.
I'll check, maybe I had a different error and I miss-interpreted it.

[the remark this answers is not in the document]
greytdepression: I just tried it out and you can even put the parameters in a different order and it will assign them properly as long as you use the exact same names. :^)
But yes, it should always be possible to fewer parameters than there are.

selfhost/typechecker.jakt
```diff
@@ -640,6 +661,17 @@ enum CheckedUnaryOperator {
     IsEnumVariant(String)
 }
 
+enum CheckedMatchBody {
+    Expression(CheckedExpression)
+    Block(CheckedBlock)
+}
+
+enum CheckedMatchCases {
+    EnumVariant(name: String, args: [(String?, String, String)], subject_type_id: TypeId, index: usize, scope_id: ScopeId, body: CheckedMatchBody, marker_span: Span)
```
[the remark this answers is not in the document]
TophEvich: good catch

selfhost/typechecker.jakt
```diff
@@ -1460,6 +1548,334 @@ struct Typechecker {
...
+        let optional_struct_id = .find_struct_in_prelude("Optional")
+        let weakptr_struct_id = .find_struct_in_prelude("WeakPtr")
+
+        // TODO: Skip the type compatibility check if assigning a T to a T? or to a
+        //       weak T? without going through `Some`.
+
+        match lhs_type {
+            TypeVariable(name) => {
+                let seen_type_id_string = generic_inferences.get(lhs_type_id_string)
+                if seen_type_id_string.has_value() {
```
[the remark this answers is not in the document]
TophEvich: Good idea

selfhost/typechecker.jakt
TophEvich: good catch, as we talked this would change `args` by reference, so I made new emtpy new_args with ensure capacity, and push the results to the new one.

TophEvich:
Rebased on the latest, and incorporated review feedback. (Thanks @cybergsus and @greytdepression)

selfhost/typechecker.jakt
```diff
@@ -2012,6 +2012,37 @@ struct Typechecker {
         }
     }
 
+    function statement_definitely_returns(this, anon statement: CheckedStatement) -> bool => match statement{
+        Return(expr) => true
+        // FIXME: CheckedStatement::If branches ugly implementation due to current compiler limitations
```
awesomekling, line 2038: It would be nice if this comment told us *which* limitation is making the code ugly, so that we'll know when to fix it :)

PR #718, opened 2022-06-21
selfhost/typechecker: Fix find_namespace_in_scope() bug

greytdepression:
There was a bug in find_namespace_in_scope() where we returned the
parent scope of the scope we were looking for instead of the scope
itself.


greytdepression:
Compare to https://github.com/SerenityOS/jakt/blob/16bbb96f9aada87e2245dadc7d71f68da45c7839/src/typechecker.rs#L611

PR #722, opened 2022-06-21
selfhost/typechecker: Impl `typecheck_for`

cg-jl:
Implements desugaring of the `For` loop exactly like the Rust
version. With this, `typecheck_statement` is 100% implemented, since
there are no match cases left to fill.


jntrnr:
thanks

PR #724, opened 2022-06-22
selfhost/typechecker: Fix panicking Optional unwrap in `typecheck_if`

cybergsus:
In `selfhost/main.jakt:54` there are these lines:
```jakt
if not file_name.has_value() {
  eprintln(...)
  eprintln(...)
  return 1
}
```

That `return 1` makes the block have a yielded type **without having
a yield span**. I've changed `typecheck_if` to find the *actual*
yield span and unwrap it when producing the error once we know it has
a value.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -1876,8 +1876,9 @@ struct Typechecker {
         }
 
         let checked_block = .typecheck_block(then_block, parent_scope_id: scope_id, safety_mode)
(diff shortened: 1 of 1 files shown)


awesomekling:
Hmm, wait, there's more code that assumes `find_yield_span()` will always succeed if there's a yielded type.
It seems wrong that `return` statements affect the yielded type in the first place.
cc @jntrnr @alimpfard

PR #726, opened 2022-06-22
Jakttest: Check compiler output and errors on SuccessTests

greytdepression:
I noticed that sometimes jakttest passes a test just because the expected output is empty, even when the compiler throws an error. Now we check what the compiler output is during SuccessTests to avoid this issue and get more accurate test results.


greytdepression:
before
```
82 passed
250 failed
7 skipped
```
and after
```
78 passed
254 failed
7 skipped
```
And this should become more relevant once #711 is merged

PR #727, opened 2022-06-22
selfhost/typechecker: Fix return error in non-void functions

mattisboeckle:
The typechecker would abort if the return type of a function was
non-void. This patch also requires the function to not definitely return
before aborting.

Fixes 4 tests.


mattisboeckle:
Nevermind, this is included in #701.

PR #731, opened 2022-06-22
selfhost/typechecker: Unify `UNKNOWN_TYPE_ID`

cybergsus:
Now you can get UNKNOWN_TYPE_ID from a single source of truth,
`unknown_type_id()`.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -771,6 +771,7 @@ struct CheckedCall {
     callee_throws: bool
 }
 
+function unknown_type_id() -> TypeId => TypeId(module: ModuleId(id: 0), id: 0)
 
 function builtin(anon builtin: BuiltinType) -> TypeId {
     return TypeId(module: ModuleId(id: 0), id: builtin as! usize)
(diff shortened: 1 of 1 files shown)


jntrnr:
You may need to give this a return type as return type inference is incomplete

selfhost/typechecker.jakt
```diff
@@ -771,6 +771,7 @@ struct CheckedCall {
     callee_throws: bool
 }
 
+function unknown_type_id() => TypeId(module: ModuleId(id: 0), id: 0)
```
TophEvich, line 774: This needs a return type (atm)
```suggestion
function unknown_type_id() -> TypeId => TypeId(module: ModuleId(id: 0), id: 0)
```

TophEvich:
@cybergsus I've been meaning to do the same.
Just a nitpicky question, why did you leave in the UNKNOWN_TYPE_ID declarations, and not just use it in place?

cybergsus:
I don't think a refactor of those uses is needed, it's perfectly fine to save it in a variable

cybergsus:
But I can inline the uses, no problem ;)

PR #732, opened 2022-06-23
selfhost/parser: Remove check for eof while parsing struct

mattisboeckle:
`parse_struct_class_body` was already handling the closing RCurly, this
caused structs to have no methods and RecordType::Garbage as the
function was returning prematurely.

### DIFF
--- a/selfhost/parser.jakt
+++ b/selfhost/parser.jakt
@@ -985,7 +985,7 @@ struct Parser {
         return (fields, methods)
     }
 
-    public function parse_struct(mut this, anon definition_linkage: DefinitionLinkage) throws -> ParsedRecord{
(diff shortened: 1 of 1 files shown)


selfhost/parser.jakt
```diff
@@ -1030,12 +1030,7 @@ struct Parser {
             return parsed_struct
         }
 
-	let fields_methods = .parse_struct_class_body(definition_linkage, default_visibility: Visibility::Public)
-
-	if .eof() {
-	    .error("Incomplete struct definition, expected `}`", .current().span())
-	    return parsed_struct
-	}
+	    let fields_methods = .parse_struct_class_body(definition_linkage, default_visibility: Visibility::Public)
```
awesomekling, line 1033: Please use 4 spaces instead of tab characters.

PR #739, opened 2022-06-23
parser+typechecker+compiler+tests: Fix unnamed namespaces

greytdepression:
After looking deeper into the issue of unnamed namespaces that I brought up in #703, there were some more issues that had to be fixed in order to make a test for this bug.
Importantly I added a `uid: u64` field to parsed namespaces and checked namespaces in order to identify unnamed namespaces in the typechecker. Before we did not even try and instead created a new scope each time we encountered the namespaces again in the typechecker (which naturally caused some bugs).
Thanks to @jntrnr for making me look deeper into this issue. I didn't really think about it as much more than fixing a typo when I opened #703.

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -52,6 +52,10 @@ pub const STRING_TYPE_ID: TypeId = TypeId(ModuleId(0), 16);
 
 pub const PRELUDE_SCOPE_ID: ScopeId = ScopeId(ModuleId(0), 0);
 
+pub struct CompilerMetaData {
+    pub namespace_uid_counter: u64,
+}
+
 pub struct SourceInfo {
     pub module_id: ModuleId,
     pub file_id: FileId,
@@ -62,6 +66,7 @@ pub struct SourceInfo {
 pub struct Compiler {
     include_paths: Vec<PathBuf>,
     pub loaded_files: Vec<SourceInfo>,
+    pub meta_data: CompilerMetaData,
 }
 
 impl Compiler {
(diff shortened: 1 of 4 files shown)


greytdepression:
It should be noted that while unnamed namespaces now are successfully typechecked, there is AFAIK no way to actually access any of its members from outside with the current state of Jakt. So I was not able to test if and how that might work.

greytdepression:
Also if this PR gets approved and merged, we should port over the fixes to the selfhosted compiler as it will have the same issues.

jntrnr:
@greytdepression - nice catch to see if we can normalise this thing a bit better. I looked through the code and had a couple questions:

Is uid the same as ScopeId? It feels like one way we went wrong with the rust-based compiler is that we have a lot of similar concept. Namespaces, scopes, modules. Makes me wonder if we should maybe tease apart the scopes from the namespace/module thing and then make their relationship a bit more clear...

We should definitely be clear about when we're creating a new scope - like you're saying. If we aren't doing that right we're going to get some weird behaviour for sure.

I wonder. Should we maybe just get rid of namespaces and just use modules? They already have ModuleIds we could use as the unique identifier. This also means we wouldn't have to do any additional work when we encounter the same one a second time as the work we did before would be interned like the other data structures.

greytdepression:
@jntrnr Technically, no. The namespace uid is not the same as the ScopeId. ScopeIds are a concept within the typechecker, whereas the namespace uid is introduced by the parser and the typechecker only makes use of it to differentiate them later.
But yes, it is a bit of a hack to get things to work in the current framework of namespaces, modules, and scopes.

I agree it might be a good idea to rework the entire concept and unify namespaces and modules into one. But that might be big enough of an issue that we might want to discuss that publicly first.

AtkinsSJ:

PR #744, opened 2022-06-24
jakttest: Script enhancements

lanmonster:
I got burned by running the tests on an old version of the selfhost.


awesomekling:
I don't love the idea of adding 15+ seconds to the startup time of `jakttest` which is already slow, but I guess it *is* better than running it on an outdated copy of the compiler.

PR #747, opened 2022-06-24
selfhost/typechecker: Fix infinite recursion in typecheck_typename

TophEvich:
Fix infinite recursion bug for GenericType in typecheck_typename.
Add a FIXME in place as the given data from the parser is also not what the typechecker would expect.
The existing parser FIXME at the corresponding spot has a third opinion.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -2629,10 +2629,13 @@ struct Typechecker {
             GenericType(name, generic_parameters, span) => {
                 mut checked_inner_types: [TypeId] = []
 
+                // FIXME: This expects the gerneric_paramters to be parsed as [ParseType] not [(String:Span)]
+                //        as in the rust version
                 for inner_type in generic_parameters.iterator() {
-                    let inner_type_id = .typecheck_typename(parsed_type, scope_id)
-                    checked_inner_types.push(inner_type_id)
+                    //let inner_type_id = .typecheck_typename(inner_type, scope_id)
+                    //checked_inner_types.push(inner_type_id)
                 }
+
                 // FIXME: add support for "GenericResolvedType", though the Rust version
                 // puts this in the parser even though it should be in the typechecker
                 // as the parser doesn't have a concept of type ids


selfhost/typechecker.jakt
```diff
@@ -2629,10 +2629,13 @@ struct Typechecker {
             GenericType(name, generic_parameters, span) => {
                 mut checked_inner_types: [TypeId] = []
 
+                // FIXME: This expects the gerneric_paramters to be parsed as [ParseType] not [(String:Span)]
```
awesomekling, line 2632: Typo: gerneric_paramters -> generic_parameters

selfhost/typechecker.jakt
```diff
@@ -2629,10 +2629,13 @@ struct Typechecker {
...
+                //        as in the rust version
                 for inner_type in generic_parameters.iterator() {
-                    let inner_type_id = .typecheck_typename(parsed_type, scope_id)
-                    checked_inner_types.push(inner_type_id)
+                    //let inner_type_id = .typecheck_typename(inner_type, scope_id)
+                    //checked_inner_types.push(inner_type_id)
```
awesomekling, line 2636: Please don't comment out code. If the code is wrong, remove it or replace it with correct code. :)

awesomekling (changes requested):

selfhost/typechecker.jakt
TophEvich: I see the rule of not commenting out code in general as a good one, and I'll honor it in the future. Just so you know, I didn't do it lightly.

I thought I'd make the correct change for typechecker (would work if parser "correct" parser), so it's clear what the issue is, but it would necessitate the correct implementation in parser, for which I didn't have the time to sort out, so I commented it out, to not have compile errors. So no runtime stackoverflow anymore, and a hint what in parser should change. As the parser FIXME comment has a third opinion (current implementation, rust implementation, parser FIXME suggestion), the solution someone wanted to have was also not clear adding to the decision on leaving parser as it is.

Issue #483, opened 2022-06-01
Returning enum named `c_int` causes C++ error


gltchitm:
This program causes the C++ compiler to throw an error:
```
enum c_int: u64 {
    Abc = 1
}

function main() {
    return c_int::Abc
}
```

Pablo96:
There is a mismatch in main's return value. The expected result here is for the error to be handled by Jakt's front-end instead of the C++ back-end.

PR #748, opened 2022-06-25
Fix: main function does not check inferred return type nor namespacevar type

Pablo96:
# What does this PR do?
It fixes [issue 483](https://github.com/SerenityOS/jakt/issues/483)

# Description of the bug it fixes
On the main function, given no explicit return value it does not check the inferred type against `is_integer`, it is related to another bug I found where if the expression is a `NamespacedVar` it does not check for its type either.

### DIFF
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -3457,6 +3457,20 @@ fn typecheck_function(
         resolve_type_var(function_return_type_id, parent_scope_id, project)
     };
 
+    // After inferring the type we need to check again for main since it can only return c_int
+    if function.name == "main" {
+        let return_type_error = Some(JaktError::TypecheckError(
+            "Main function must return c_int".to_string(),
+            function.name_span,
+        ));
+        if !is_integer(return_type_id) {
+            match return_type_id {
+                VOID_TYPE_ID | UNKNOWN_TYPE_ID => (),
+                _ => error = error.or(return_type_error),
+            }
+        }
+    }
+
     if function_linkage != FunctionLinkage::External
         && return_type_id != VOID_TYPE_ID
         && !block.definitely_returns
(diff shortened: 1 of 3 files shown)


src/typechecker.rs
```diff
@@ -4674,10 +4674,16 @@ pub fn typecheck_expression(
...
-                            check_accessibility(scope_id, variable_scope_id, var, span, project),
-                        )
+                        let checked_namespaced_var = CheckedExpression::NamespacedVar(checked_namespace, var.clone(), *span);
+                        let (_, unify_error) = unify_with_type(var.type_id, type_hint, *span, project);
+                        let access_error = check_accessibility(scope_id, variable_scope_id, var, span, project);
+
+                        let namespaced_var_error: Option<JaktError> = match unify_error {
+                            Some(_) => unify_error,
+                            None => access_error,
+                        };
```
awesomekling, line 4698: This seems like it could be simplified. Something like...
```rust
let error = unify_error.or(access_error);
```

awesomekling (changes requested):
This fails the `rustfmt` check on CI.

src/typechecker.rs
Pablo96: Indeed it was a better solution. Rust newbie mistake :/

awesomekling:
This is still failing `rustfmt` on CI. Also, please squash the fixup into the original commit instead of adding a separate commit :)

src/typechecker.rs
```diff
@@ -3449,6 +3449,20 @@ fn typecheck_function(
         resolve_type_var(function_return_type_id, parent_scope_id, project)
     };
 
+    // After infering the type we need to check again for main since it can only return c_int
```
Jonta, line 3452: Spelling error:

inferring

:^)

Pablo96: Thank you for pointing that out! I fixed it now.

AtkinsSJ:

PR #749, opened 2022-06-25
selfhost/lexer: Return from next() in case of Eof after whitespace

mnlrsn:
Accept files where the last line ends in a space (' ') character.
Accomplished by returning a Token::Eof constructed in the same way
as used for normal Eof cases.

### DIFF
--- a/selfhost/lexer.jakt
+++ b/selfhost/lexer.jakt
@@ -765,14 +765,20 @@ struct Lexer {
     }
 
     function next(mut this) throws -> Token? {
-        if .index == .input.size() {
-            ++.index
-            return Token::Eof(Span(start: .index - 1, end: .index - 1))
-        }
-        if .eof() {
-            return None
-        }
+        // Consume whitespace until a character is encountered or Eof is
+        // reached. For Eof return a token.
         loop {
+            if .index == .input.size() {
+                ++.index
+                return Token::Eof(Span(start: .index - 1, end: .index - 1))
+            }
+            // FIXME: Once the handling of Token::Eof is fully implmented,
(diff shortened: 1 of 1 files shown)


selfhost/lexer.jakt
```diff
@@ -774,7 +774,14 @@ struct Lexer {
         }
         loop {
             let ch = .peek()
-            if ch == b' ' or ch == b'\t' or ch == b'\r' {
+            if ch == 0 {
+                // ch == 0 means that .peek() has detected the end of the input
+                // stream and reported with a Token:Eof using the same construct
+                // for the token as above. This includes incrementing the index, 
+                // even if it is already past the end of the input array.
```
awesomekling, line 781: Instead of relying on `.peek()` to return `0`, let's explicitly check for `.eof()` here instead.

mnlrsn: > Instead of relying on `.peek()` to return `0`, let's explicitly check for `.eof()` here instead.

Sure - also makes it express the intention more clearly. However, that would effectively put four checks for EOF in a row:
- One at the beginning of the function by checking `.index == .input.size()`
- One right after to check if the index has been incremented further
- One at the top of the loop
- One inside eof() (which then returns 0 if at EOF)

To avoid adding more tests for specific cases, I suggest to put it all in the loop, which has the purpose of consuming white space and bail out if EOF is encountered. Left a FIXME for the `if` block that returns `None`. Once that FIXME is handled, this will leave only a single test for EOF in the function.

I believe that this makes for cleaner code, but most happy to hear if that is not the case.

PR #751, opened 2022-06-25
lexer: Better string/char error handling

mattco98:
Previously when parsing a string, the lexer would just continue to eat characters until a closing quote or EOF was found. It did produce a correct error for such a case, however this meant it could possible consume a _lot_ of input if there were no more strings for a while. This commit makes the lexer stop parsing if a newline is encountered. 

This was causing issues for my IntelliJ plugin, here are the results of this change: 

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

Also,  better invalid char escape errors:

<details>
<summary>Before</summary>

<br />
Idk why there's no pretty error message for this one

</details>

<details>
<summary>After</summary>

<br />

</details>

### DIFF
--- a/selfhost/lexer.jakt
+++ b/selfhost/lexer.jakt
@@ -388,6 +388,12 @@ struct Lexer {
         mut escaped = false;
 
         while not .eof() and (escaped or .peek() != b'\'') {
+            if escaped and (.index - start > 3) {
+                break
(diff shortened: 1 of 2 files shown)


mattco98:
Draft while I also make the changes to the selfhost compiler, since I might as well.

PR #754, opened 2022-06-26
neovim: Add LSP support for neovim

cybergsus:
When a jakt file is opened, the LSP server in
`editors/vscode/out/server/src/server.js` is spawned for the current
instance, and the buffer is attached to the client. Requires Nvim 0.7.2.

Before merging, it must be checked that:
  - [x] The neovim version where it works is notified in the documentation, 
      since Neovim's Lua API might change in the future.
  - [x] Instructions to run `npm run compile` so that the `out` path exists.
  - [x] Make `editors/vim/lua/lsp.lua` check for the existing of the script and 
      notifying if it's not there, before trying to start the client.

More yaks will be added to the stack if necessary.

### DIFF
--- a/editors/vim/lua/lsp.lua
+++ b/editors/vim/lua/lsp.lua
@@ -1,23 +1,37 @@
+-- Set per Neovim *instance*, so that opening another Jakt file (buffer)
+-- in the same instance reuses the already spawned client.
 local client = nil
 
 local function start()
   if client == nil then
+    -- Resolve paths so we don't depend on neovim's CWD:
+    -- `script_dir` will always point to
+    --  <symlink by plugin manager>/lua
     local script_dir = debug.getinfo(1, "S").source:match('@?(.*/)')
+    -- We use `resolve` so that `plugin_dir` points to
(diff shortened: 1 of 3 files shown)


editors/vim/lua/lsp.lua
```diff
@@ -0,0 +1,21 @@
+local client = nil
+
+local function start()
+  if client == nil then
+    local script_dir = debug.getinfo(1, "S").source:match('@?(.*/)')
+    local plugin_dir = vim.fn.resolve(script_dir .. '../')
+    local ls_path = vim.fn.resolve(plugin_dir .. '/../vscode/out/server/src/server.js')
```
jntrnr, line 7: does this mean you can only start nvim from a particular directory?

cybergsus: nope. `script_dir` is set so it always points to `<jakt symlink by package manager>/editors/vim/lua/`.
I use `resolve` to resolve the symlink and make the plugin dir point to `<path/to/jakt>/editors/vim`, so I can access `editors/vscode/out/...` correctly. I'l add comments to avoid confusion.

PR #755, opened 2022-06-26
selfhost/lexer: Define whitespace in a single function

mnlrsn:
A single location to define whitespace, and a name to make intent clear when scanning for the next token in the lexer. If useful, it could be made static to allow use from other parts of the compiler.

### DIFF
--- a/selfhost/lexer.jakt
+++ b/selfhost/lexer.jakt
@@ -847,4 +847,8 @@ struct Lexer {
             else => .lex_number_or_name()
         }
     }
+
+    function is_whitespace(this, anon ch: u8) -> bool {
+        return ch == b' ' or ch == b'\t' or ch == b'\r' or ch == b'\f' or ch == b'\v'
+    }    
 }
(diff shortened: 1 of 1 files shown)


awesomekling:
Doesn't this mean that we'll ignore newlines instead of creating the expected `Token::Eol`?

mnlrsn:
> Doesn't this mean that we'll ignore newlines instead of creating the expected `Token::Eol`?

No, it does (hopefully) not ignore newlines. The definition of whitespace came out of a discussion on Discord in the #jakt-design channel (2022-06-26) where it was suggested to just skip `\r` (carriage return) since the Eol is identified by `\n` (newline). This is why `\n` is *not* in the list of characters being identified as whitespace in the function, but `\r` is. 

So if calling this function, all whitespace in the traditional Unix sense ("\f\n\r\t\v ") will be skipped *except* newline, which will produce an Eol token.

A condensed version of this comment could be included in the code, and maybe the name of the function could be changed to avoid confusion with the traditional definition.

PR #757, opened 2022-06-27
selfhost/typechecker: Add basic typechecking of enums

greytdepression:
This adds most of the necessary components to typecheck enums, it still needs support for generic parameters and struct-like enums however.
The test results with these commits are
```
107 passed
225 failed
7 skipped
```
as opposed to the current main results
```
104 passed
228 failed
7 skipped
```
with the following additional tests passing:
* `samples/enums/empty_enum.jakt`
* `tests/parser/unclosed_brace_in_enum.jakt`
* `tests/parser/unclosed_brace_in_ref_enum.jakt`


greytdepression:
Apparently @lanmonster is already working on this. Oops :^)

PR #758, opened 2022-06-27
selfhost/codegen: Add comments to the C++ code that link back to the original Jakt statement

greytdepression:
This PR adds a flag `-d` to the selfhosted compiler that will put comments into the C++ code that link to the span of the corresponding statement in the Jakt source file.
This will be helpful for debugging compile errors of the generated C++ code as well as debugging runtime errors :^)
Example output:
```C++
#include <lib.h>
namespace Jakt {
ErrorOr<int> main();
ErrorOr<int> main(Array<String>) {
{
const i64 x = static_cast<i64>(80LL); /* samples/control_flow/if4.jakt:5:5 */
if ((x == static_cast<i64>(100LL))) /* samples/control_flow/if4.jakt:7:5 */ {
outln(String("true")); /* samples/control_flow/if4.jakt:8:9 */
}
else if ((x == static_cast<i64>(80LL))) /* samples/control_flow/if4.jakt:9:12 */ {
outln(String("it is 80")); /* samples/control_flow/if4.jakt:10:9 */
}
}
return 0;
}
}
```
The jakttest output is the same as the one of the current main branch independently of whether or not we pass the debug flag to the compiler
```
115 passed
217 failed
7 skipped
```

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -1206,7 +1206,7 @@ struct CodeGenerator {
     }
 
     function codegen_statement(mut this, statement: CheckedStatement) throws -> String => match statement {
-        Throw(expression) => "return " + .codegen_expression(expression) + ";"
+        Throw(expr) => "return " + .codegen_expression(expr) + ";"
         Continue => match .control_flow_state.passes_through_match {
             true => "return JaktInternal::LoopContinue{};"
             else => "continue;"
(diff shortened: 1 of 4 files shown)


greytdepression:
@AtkinsSJ Oh so I did mess up the rebase :| I'll fix it real quick

PR #764, opened 2022-06-29
selfhost/typechecker: Fix error about types not matching

lanmonster:


lanmonster:
> 

CheckedExpression.span is reporting an error that it expects a `Span` but is getting a `Span?` from the `Defer` branch. The expected and actual types are flipped, but this change makes the function return `Span?` in every case.

awesomekling:
Where is this error reported?

lanmonster:
in VSCode
image

lanmonster:
Oops! I just needed to `meta/update-vscode-extension.sh`.

PR #767, opened 2022-06-30
jakttest: Check for selfhost & binary modified times before recompiling

cybergsus:


cybergsus:
Something looks wrong revising this, please don't merge.

cybergsus:
OK, it was that `sort` gave them in ascending order, so I used `-r` to get the latest one

cybergsus:
Hmm, we have to check *all* the files under `selfhost/`. Reason behind this is that the directories' mtime is not updated when you update one of their inner files.

cybergsus:
I guess an `any file under selfhost/*.jakt` is newer than `build/main` could work better, perhaps.

cybergsus:
PD I also forgot to remove the commented `cargo run`

jntrnr:
Latest main should have the CI errors fixed

PR #769, opened 2022-06-30
Make all of `samples/variables/` pass

lanmonster:


jntrnr:
Those clippy warnings are coming from the latest Rust release. We'll probably need to fix them here or have a separate PR that fixes them up so the CI can go green again.

jntrnr:
Latest main should have the CI errors fixed

PR #771, opened 2022-07-01
selfhost/main: Add "build and run" flag `-r`

TophEvich:
Add `-r` "build and run" flag to selfhost compiler.

_(sjakt is my local alias for selfhost build/main)_
```
$ sjakt -h
usage: jakt [-h] [OPTIONS] <path>

Flags:
  -h            Print this help and exit.
  -l            Print debug info for the lexer.
  -p            Print debug info for the parser.
  -b            Build an executable file.
  -r            Build and run an executable file.
  -d            Insert debug statement spans in generated C++ code.
```

Example:
```
$ sjakt -r samples/basics/hello.jakt 
Well, hello friends.
```

Also split up the output filename and the cpp filename for a future `-o` option.

PR #772, opened 2022-07-01
jaktest: Simplify, speed up and enable imports

TophEvich:
**_This depends on #771 to be merged first._**

Combine the compile and run steps into one by using the new `-r` "build and run" compiler flag introduced with #771.
Instead of writing the given jakt source file to output.jakt use it directly, which also enables the testing of applications that use imports.

Don't try to remove files that are no longer generated in run-all.

### DIFF
--- a/jakttest/jakttest.jakt
+++ b/jakttest/jakttest.jakt
@@ -77,38 +77,22 @@ function main(args: [String]) {
     let parsed_test = Parser::parse(tokens, errors);
 
     // Windows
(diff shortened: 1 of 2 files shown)


greytdepression:
I think you mean #771. #711 won’t help you much here ;)

TophEvich:
> I think you mean #771. #711 won’t help you much here ;)

Thanks, I had #771 just one line above, and typoed it on the second :D

PR #778, opened 2022-07-01
editors/vscode: Implement goto-ing to `import`ed files

sppmacd:
This PR add support for "go to file" when declaration is in another (imported) source file. Note that this doesn't work for prelude types (e.g String, Error, File etc.)


awesomekling:
Awesome!

PR #783, opened 2022-07-02
Everywhere: Added `build.sh` for ease-of-use installation

kernelk14:
I've added a `build.sh` script so that the Jakt installation will be so
mush easier.


awesomekling:
I don't think we need a script that copies Jakt headers into `/usr/include`.

PR #786, opened 2022-07-02
jakttest: Recompile jakttest if it is not up-to-date

TophEvich:
Same as was done in  #767 for jakt, but this time for jakttest.
Minor adjustment to only check for `.jakt` files and fix of the info message.

### DIFF
--- a/jakttest/run-all.sh
+++ b/jakttest/run-all.sh
@@ -5,13 +5,23 @@ if [ ! -x ./build/jakttest ]; then
     cargo run jakttest/jakttest.jakt
 fi
 
+# check for jakttest/ mtime and build/jakttest mtime
+{
+    jakttest_binary_mtime=$(stat -c %Y build/jakttest)
+    latest_jakttest_source_mtime=$(stat -c %Y jakttest/*.jakt | sort | tail -n1)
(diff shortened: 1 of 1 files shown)


jakttest/run-all.sh
```diff
@@ -5,13 +5,23 @@ if [ ! -x ./build/jakttest ]; then
     cargo run jakttest/jakttest.jakt
 fi
 
+# check for jakttest/ mtime and build/jakttest mtime
+{
+    jakttest_binary_mtime=$(stat -c %Y build/jakttest)
+    latest_jakttest_source_mtime=$(stat -c %Y jakttest/*.jakt | sort | tail -n1)
```
cybergsus, line 11: Maybe it would be better to have a separate function since we're reusing the same cryptic `stat -c | sort | ...` pipeline

TophEvich: I don't know if that abstraction would be worth it here, if it is done a 3rd time and the script grows a lot more (it should even shrink with #772 by quite a bit, at least in %), maybe ...

cybergsus:
Maybe you want to wait for #787 where I move the logic to a function that does all the checking that is needed.
Maybe that function can be split in the future when we *know* that the binary is there

PR #787, opened 2022-07-02
jakttest: Move checking whether binary is updated to a separate function

cybergsus:
Since there might be more checking work to do between source directories
and their binary targets, I moved the logic into a function that
pre-checks if the binary exists before checking its mtime.
It also can't do proper module dependency analysis yet but it's a starting point.

PD: This function *should* only be used when we don't know the state of the source or binary, and in that case it *should* be split up to avoid unnecessary work; e.g when we know that the binary exists or
we know the exact list of files we want to check.

### DIFF
--- a/jakttest/run-all.sh
+++ b/jakttest/run-all.sh
@@ -1,30 +1,33 @@
 #!/bin/bash
 
-if [ ! -x ./build/jakttest ]; then
-    echo "jakttest binary does not exist. Building it now."
+
+is_binary_behind_src() {
+    local binary_path=$1
+    local src_dir=$2
+    # if the target file does not exist, then
+    # we'll say yes
+    if [ ! -x "$binary_path" ]; then
+        return 0
+    else
+        local binary_mtime=$(stat -c %Y "$binary_path")
+        local latest_source_mtime=$(\
+            stat -c %Y $src_dir/*.jakt |\
+            sort |\
+            tail -n1)
+        [ "$binary_mtime" -lt "$latest_source_mtime" ]
+    fi
+}
+
+if is_binary_behind_src build/jakttest jakttest; then
+    echo "(re)compiling jakttest to match source"
     cargo run jakttest/jakttest.jakt
 fi
 
-# check for jakttest/ mtime and build/jakttest mtime
-{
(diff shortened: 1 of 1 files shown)


TophEvich:
Yeah I might wanna rebase #786 on this one.
Would it be OK if I integrate your commit in my PR?

Also I don't know if we want to go recursive just yet, as the `import` is a bit immature at the moment, I think it was only thought of to be used in the same directory for now. But I guess it doesn't hurt.

Also the way you implemented it here you'd also find "hidden" files in "hidden" dirs, I think that is not intended.
```bash
$ find "selfhost/" -name '*.jakt'
selfhost/main_with_flags_dict.jakt
selfhost/error.jakt
selfhost/typechecker.jakt
selfhost/.hidden/.foobar.jakt <---
selfhost/parser.jakt
selfhost/utility.jakt
selfhost/main.jakt
selfhost/prelude.jakt
selfhost/compiler.jakt
selfhost/lexer.jakt
selfhost/codegen.jakt
```
Maybe a 
```diff
- find "$src_dir" -name '*.jakt' -exec stat -c %Y \{\} + |\
+ find "$src_dir" -not -path '*/.*' -type f -name '*.jakt' -exec stat -c %Y \{\} + |\
```

cybergsus:
Don't integrate my commit, just wait for this PR to be merged and then rebase your changes with mine.
Also trying to hide a source file from rebuilding selfhost/jakttest doesn't really make sense. Stay practical.

TophEvich:
> Also trying to hide a source file from rebuilding selfhost/jakttest doesn't really make sense, and adds work that isn't needed to run, in a function that is used specifically to avoid work.
I don't quite understand. When excluding the hidden directories and files you are minimizing the work to be done, it doesn't add work. Hidden folders and files will not be build, so why would we take them into account?

cybergsus:
While merging I saw that we don't need the recursive aspect of `find` yet and can rely on the (faster) one-depth globbing. Not much but it shaves unneeded work.

TophEvich:
Ok, you changed it back from find to globbing, I'm fine with that.

cybergsus:
Superseeded by #818.

PR #788, opened 2022-07-02
typechecker: Error when prefilling array in functions that don't throw

devBoi76:
Prefilling an array is wrapped inside a TRY() in generated cpp code so its scope should throw

This also should fix issue #137

### DIFF
--- a/src/typechecker.rs
+++ b/src/typechecker.rs
@@ -4772,6 +4772,14 @@ pub fn typecheck_expression(
                     typecheck_expression(fill_size_expr, scope_id, project, safety_mode, None);
                 checked_fill_size_expr = Some(Box::new(checked_expr));
                 error = error.or(err);
+
+                let parent_throws = project.get_scope(scope_id).throws;
+                if !parent_throws {
+                    error = error.or(Some(JaktError::TypecheckError(
+                        "Prefilling an array may throw and needs to be in a try statement or a function marked as throws".to_string(),
+                        fill_size_expr.span(),
+                    )))
+                }
             }
 
             for v in vec {


src/typechecker.rs
```diff
@@ -4766,6 +4766,14 @@ pub fn typecheck_expression(
                 }
             }
 
+            let parent_throws = project.get_scope(scope_id).throws;
+            if fill_size_expr.is_some() && !parent_throws {
+                error = error.or(Some(JaktError::TypecheckError(
+                    "Prefilling an array may throw and needs to be in a try statement or a function marked as throws".to_string(),
+                    *span,
```
jntrnr, line 4773: can we use the span of the `fill_size_expr` as the location of the error?

devBoi76: Sure, I changed it :)

jntrnr:
@devBoi76 - Great, now can you squash the two commits into one? With that, I think we'll be okay

devBoi76:
> @devBoi76 - Great, now can you squash the two commits into one? With that, I think we'll be okay

Done
Thank's for pointing out these issues btw :)

src/typechecker.rs
```diff
@@ -4766,6 +4766,14 @@ pub fn typecheck_expression(
                 }
             }
 
+            let parent_throws = project.get_scope(scope_id).throws;
+            if fill_size_expr.is_some() && !parent_throws {
+                error = error.or(Some(JaktError::TypecheckError(
+                    "Prefilling an array may throw and needs to be in a try statement or a function marked as throws".to_string(),
+                    fill_size_expr.as_ref().unwrap().span(),
```
jntrnr, line 4773: if you move this lower, inside of the `if let Some(fill_size_expr) ...` you can simplify and not have to do the `is_some()` or the `.as_ref().unwrap()` step. Should make it a little easier.

jntrnr:
Looks like you'll need to run `cargo fmt` on your code to make sure the Rust code is in the standard code formatting.

devBoi76:
> Looks like you'll need to run `cargo fmt` on your code to make sure the Rust code is in the standard code formatting.

Hi, I won't have access to my PC until Sunday because of IRL stuff so that's when I'll be able to fix it. Sorry for taking up so much time for a simple change, I'll make sure to remember all of these steps next time.

devBoi76:
Ok, I ran `rustfmt` on the file so it should work now!

jntrnr:
@devBoi76 - looks like you need to run `cargo fmt` on it

PR #790, opened 2022-07-02
selfhost/typechecker: Make better use of hashmaps

greytdepression:
Use hashmap.get(...) instead of iterating over keys to find an item.
Potentially brings the runtime down from Omega(n) to O(1) :^)


greytdepression:
The number of test that pass is not affected (as it should be)

PR #792, opened 2022-07-02
selfhost/lexer+typechecker: Add support for floats

mnlrsn:
Add support for simple floats that consist of just the integer part,
e.g., 1f64 but not 1.5f64. Defines two functions to create the token,
one for integers and one for floats in place of make_number_token().


jntrnr:
@mnlrsn - can you add the full support for floats? We could land this, but it feels a bit incomplete if it doesn't handle floats yet

mnlrsn:
Will give it a go.

mnlrsn:
Added a floats to the lexer so a `let a = 1.1` works. This change makes `samples/basics/float_literals.jakt` pass.

mnlrsn:
> @mnlrsn - can you add the full support for floats? We could land this, but it feels a bit incomplete if it doesn't handle floats yet

Done. Partially inspired by the parser in samples/app/json.jakt

mnlrsn:
At the current state it does not work with ranges (`0..1`) which are lexed as a float (`0.`) with one dot too many. Converted to draft until it can be fixed.

mnlrsn:
Now it will not try to lex (`0..3`) as a malformed float and also checks that there is only a single '_' between two digits, and that a number does not end with a dangling underscore.

awesomekling:
This has a conflict

mnlrsn:
The conflicts were from the change from `Span()` to `.span()`. Fixed.

mnlrsn:
Before:
270 passed
63 failed
7 skipped
After:
273 passed
60 failed
7 skipped

mnlrsn:
Previously added tests have been removed to keep the total number of tests constant. `samples/basics/float_literals.jakt` updated to include floats with underscore and suffix (plus one with a negative sign).

PR #796, opened 2022-07-03
selfhost/typechecker+codegen: Pass on match result type and implement yield 

greytdepression:
@jntrnr I decided to work on the matches some more, but before I could push it you already merged the PR :D

This allows an additional test `match_block_yield.jakt` to pass, which means we now have 166 passing and 166 failing o.o


jntrnr:
nice!

PR #802, opened 2022-07-03
typechecker+tests+selfhost/parser+typechecker: Parse and typecheck sets

greytdepression:
As the name suggests we now parse and typecheck sets. With this there are three additional tests passing:
```
samples/sets/empty_literal_without_hint.jakt
tests/typechecker/set_mismatching_value_types.jakt
tests/typechecker/set_with_void_values.jakt
```
Generating code for sets does not make much sense right now, since all other tests rely on method calls which don't work yet. So the compiler never gets past the typechecking phase.


jntrnr:
that's fine... I was thinking the method calls should be close to working (I fixed some issues earlier today) but maybe it still needs a bit more work.

PR #809, opened 2022-07-03
selfhost: Work on generic functions

AtkinsSJ:
I haven't met my original goal of making `samples/generics/generic_fun.jakt` succeed, but that turned out to be a much deeper yak-hole than I anticipated. :sweat_smile: 

The last commit causes an infinite loop in `substitute_typevars_in_type()`, so leaving this as a draft until that's figured out.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -334,6 +334,8 @@ class Scope {
     public parent: ScopeId?
     public children: [ScopeId]
     public can_throw: bool
+
+    public debug_name: String
 }
 
 class Module {
@@ -1010,7 +1012,7 @@ struct Typechecker {
         )
 
         let PRELUDE_SCOPE_ID: ScopeId = typechecker.prelude_scope_id()
-        let root_scope_id = typechecker.create_scope(parent_scope_id: PRELUDE_SCOPE_ID, can_throw: false)
+        let root_scope_id = typechecker.create_scope(parent_scope_id: PRELUDE_SCOPE_ID, can_throw: false, debug_name: "root")
         typechecker.typecheck_module(parsed_namespace, scope_id: root_scope_id)
 
         return typechecker.program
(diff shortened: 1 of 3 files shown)


selfhost/typechecker.jakt
```diff
@@ -5176,17 +5240,54 @@ struct Typechecker {
...
 
                     args.push((call.name, checked_arg))
                 }
+
+                // We've now seen all the arguments and should be able to substitute the return type, if it's contains a
+                // type variable. For the moment, we'll just checked to see if it's a type variable.
+                if type_hint.has_value() {
+                    .check_types_for_compat(lhs_type_id: return_type, rhs_type_id: type_hint!, generic_inferences: generic_substitutions, span)
+                }
+                return_type = .substitute_typevars_in_type(type_id: return_type, generic_inferences: generic_substitutions)
```
AtkinsSJ, line 5253: This line causes an infinite loop inside `substitute_typevars_in_type()`. I'm not sure why, as that seems to be a faithful recreation of the `substitute_typevars_in_type()` in the rust-based compiler, which works fine.

selfhost/typechecker.jakt
```diff
@@ -78,13 +78,13 @@ struct TypeId {
     function from_string(anon type_id_string: String) throws -> TypeId {
         let parts = type_id_string.split('_')
         if not (parts.size() == 2) {
-            throw Error::from_errno(999) // FIXME: good error message
+            panic(format("Failed to convert string `{}` to a TypeId: Wrong number of parts. (Wanted 2, got {})", type_id_string, parts.size()))
```
TophEvich, line 81: I can't comment on the commit message directly. The header says `Fix inverted condition`, but you only change from error to panic with a message, I see no change in conditions.

TophEvich (commented):
I only had a quick glance, will continue time permitting.

selfhost/typechecker.jakt
AtkinsSJ: Good catch! It *did* fix the condition, but then someone else did and got their fix merged first. I am being too slow. :sweat_smile:

AtkinsSJ:
Fixed a merge conflict, and changed a commit message that no longer made sense. This now doesn't infinite loop on `samples/enums/simple_match.jakt` (someone else must have fixed that) so this should be ready to review now. :^)

AtkinsSJ:
Fixed some things!

main:
```
261 passed
72 failed
7 skipped
```

This PR:
```
262 passed
71 failed
7 skipped
```

(All that for 1 test? :sob:) Like I said though, this isn't a full generic-functions thing, just some work towards it.

jntrnr:
Just gave it a quick spin. I'm seeing some different numbers for some reason.

Main:
```
267 passed
66 failed
7 skipped
```

This PR:
```
262 passed
71 failed
7 skipped
```

Are you testing against the latest main, by chance?

jntrnr:
Oh okay, it's just the main wasn't quite updated. After merging main:
```
268 passed
65 failed
7 skipped
```

AtkinsSJ:
PHEW! You scared me there, I thought I had to hunt the bugs again

PR #810, opened 2022-07-04
selfhost/parser: Error when if and else are identical

adleroliveira:
This adds a pass for the test at samples/control_flow/if_else_identical_error.jakt
In order to compare the ParsedBlocks, I had to add the equals function to the following Enums:
- ParsedBlock
- ParsedStatement
- BinaryOperator
- UnaryOperator
- ParsedExpression
- ParsedCall

Turns out is very ackward at the moment to compare Enum variants, so the code got a bit verbose.

### DIFF
--- a/selfhost/parser.jakt
+++ b/selfhost/parser.jakt
@@ -2826,14 +2826,8 @@ function parsed_statement_equals(anon lhs_statement: ParsedStatement, anon rhs_s
         }
         else => false
     }
-    Break => match rhs_statement {
-        Break => true
-        else => false
-    }
-    Continue => match rhs_statement {
-        Continue => true
-        else => false
-    }
+    Break => rhs_statement is Break
+    Continue => rhs_statement is Continue
     Return(expr: lhs_expr) => match rhs_statement {
         Return(expr: rhs_expr) => {
             if not lhs_expr.has_value() {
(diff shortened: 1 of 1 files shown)


selfhost/parser.jakt
```diff
@@ -155,6 +170,164 @@ boxed enum ParsedStatement {
     Try(stmt: ParsedStatement, error_name: String, error_span: Span, catch_block: ParsedBlock, span: Span)
     Garbage(Span)
 
+    function equals(this, anon parsed_statement: ParsedStatement) -> bool => match this {
```
jntrnr, line 173: Can we put the equality logic in a separate file? Probably better separate so that the parser file is focused on the parser.

jntrnr: This may mean moving the definitions to a different file, or using a function instead of a method.

adleroliveira: That's probably for the best. Will change it to a different file.

selfhost/parser.jakt
```diff
@@ -155,6 +170,164 @@ boxed enum ParsedStatement {
...
+            }
+        }
+        While(condition, block) => {
+            let lhs_condition = condition
+            let lhs_block = block
+            yield match parsed_statement {
+                While(condition, block) => lhs_condition.equals(condition) and lhs_block.equals(block)
+                else => false
+            }
+        }
```
alimpfard, line 248: You can actually write these a bit more compactly:
```suggestion
        While(condition: lhs_condition, block: lhs_block) => {
            yield match parsed_statement {
                While(condition, block) => lhs_condition.equals(condition) and lhs_block.equals(block)
                else => false
            }
        }
```

selfhost/parser.jakt
adleroliveira: I remove all the .equals methods into separate functions. For now, this functions live on the parser file since it's not possible to put them in a separate file due to imports not handling circular references.

selfhost/parser.jakt
adleroliveira: Changed into the compact form as suggested

PR #814, opened 2022-07-04
Ide: Add fancier variable signature

greytdepression:
As seen on discord :^)


greytdepression:
whoops, forgot to rebase

greytdepression:
i'll quickly add a `public` keyword in front of all public fields. i think it's better to be explicit about that (especially since classes are private by default)

greytdepression:
I'm working on some more features. I'll open a new PR when I implemented them.

PR #815, opened 2022-07-04
main: Fix enum codegen for empty struct-like case

maddanio:
This fixes codegen for struct-like enums with an empty case.
Previously this tried to emit inheritance and template declarations
which were broken since there was nothing to inherit from.

### DIFF
--- a/tests/codegen/enum.jakt
+++ b/tests/codegen/enum.jakt
@@ -1,0 +1,9 @@
+/// Expect:
+/// - output: ""
+
+enum Test {
+    Some(s: String)
+    None()
+}
+
+function main() {}
--- a/src/codegen.rs
+++ b/src/codegen.rs
@@ -614,7 +614,7 @@ fn codegen_nonrecursive_enum(
     for variant in &enum_.variants {
         match variant {
             CheckedEnumVariant::StructLike(name, members, _) => {
-                if is_generic {
+                if is_generic && !template_args.is_empty() {
                     output.push_str("    template<");
                     output.push_str(template_args.as_str());
                     output.push_str(">\n");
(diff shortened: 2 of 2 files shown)


tests/codegen/enum.jakt
```diff
@@ -0,0 +1,9 @@
+/// Expect:
+/// - output: ""
+
+enum Test {
+    Some(s: String)
+    None()
```
alimpfard, line 6: Can we just disallow this instead? or transparently convert this typed variant into an untyped one?

greytdepression: Yeah, imo we should emit a warning and convert it to untyped.

maddanio: I don't understand, why do you want to disallow this? This is extremely useful. And its not untyped at all. This will be extremely useful as a return type for example returning an event, some events having payload, others not.

maddanio: That and it allows for a very simple and beautiful in-language optional, or variants which have a none state, i.e. shorthanding Optional<Varian>, which makes these much nicer to use imho. Please don't disallow this, I really want it :).

greytdepression: @maddanio 
What we mean is that
```
enum Test {
    Some(s: String)
    None
}
```
is already supported and `None` (without the parentheses) should probably be either the only way to write it or highly encouraged to be how you write it.
Also there already is an in-language optional:
```
enum Optional<T> {
    Some(T)
    None
}
```
Every `type?` gets converted to `Optional<type>` behind the scenes :)
I'm not entirely sure what you mean with how to handle events, but maybe I already addressed that. If not I'm happy to discuss this more :D

maddanio: aaaaah, now I get it. you lost me with untyped. None is a type in my head. so yes, basically the parentheses should be forbidden, ok, will try that :-D

PR #816, opened 2022-07-04
selfhost+runtime: Add support for inline cpp codegen

jntrnr:
This adds the `replace` method to the String runtime and then uses it to add support for inline cpp codegen.

Before:
```
239 passed
93 failed
7 skipped
```

After:
```
241 passed
91 failed
7 skipped
```


jntrnr:
checked by @awesomekling

PR #818, opened 2022-07-05
jakttest: Move selfhost & jakttest to custom build dirs

cybergsus:
If any other file named `main.jakt` is compiled with no extra arguments, 
its resulting binary will be `build/main`. This will make `run-all.sh` 
think that selfhost is already built.

By moving both selfhost and jakttest to a separate build directory we 
reduce the chance of this situation happening.

I also made it a variable in case anyone wants a different output 
directory, with the corresponding changes in Jakttest to accept the 
custom executable path.

Superseeds #787, since it resolves both the merge conflict and the 
extrapolation of checking mtimes to a separate function.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@ target
 a.out
 output.cpp
 build
+selfhost/build
+jakttest/build
 
 .vscode/
 .idea/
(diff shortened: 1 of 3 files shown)


cybergsus:
Closing since the run-all script was made redundant.

PR #820, opened 2022-07-05
jakttest: Parallelize Jakttest

cybergsus:
Rewrote Jakttest so it accepts all the test files and runs them through a process scheduler to be able to run the tests in parallel. The binary gets a clean temporary directory where it places one temporary directory per job. It uses that to store both C++ temporaries from Jakt compiler's output and stderr/stdout results, which it will read once it knows the job has finished with the directory. Doesn't use any locks since it makes sure to only run one job per directory.

Most of the "changes" here are just removing unneeded files (mainly the 900loc from lexer/parser) since I wrote a custom `Expect` section parser that runs through each file to know what to expect from its output. This new parser does not care about the Jakt language; instead it only looks for the expect section, reducing the possibility of having a Jakt lexer bug propagate into Jakttest.

### DIFF
--- a/jakttest/jakttest.jakt
+++ b/jakttest/jakttest.jakt
@@ -1,7 +1,24 @@
 import error { JaktError, print_error }
 import parser { Parser }
+import lexer { Lexer }
 
-function usage() => "usage: jakttest [-h] [OPTIONS] <path>"
+namespace process {
+    /// Starts a background job by means of fork() and exec(). Returns
+    /// the PID of the launched process.
+    extern function start_background_process(args: [String]) throws -> i32
+    /// Checks whether a process has finished executing. Returns its
+    /// exit code if it has.
+    extern function poll_process_exit(pid: i32) throws -> i32?
+    /// Kills a process by sending SIGKILL to it
+    extern function forcefully_kill_process(pid: i32) throws
+}
+
+function print_usage()  {
+    eprintln("usage: jakttest [OPTIONS...] <temp-dir> <path> [... <path>]")
+    eprintln("OPTIONS:")
+    eprintln("\t-h\t\tShow this message and exit.")
+    eprintln("\t-j,--jobs <JOBS>\t\tUse JOBS processes. Defaults to 8.")
+}
 
 function compare_test(bytes: [u8], expected: String) throws -> bool {
     // first, build up the content to compare to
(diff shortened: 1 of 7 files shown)


jakttest/jakttest.jakt
```diff
@@ -59,13 +63,96 @@ function strip_line_breaks(anon input: String) throws -> String {
...
-    if args.size() <= 1 {
-        eprintln("{}", usage())
-        return 1
+    let parsed_options = Options::from_args(args)
+
+    if not parsed_options.errors.is_empty() {
+        for error in parsed_options.errors.iterator() {
+            eprintln("Error: {}.", error)
+        }
+        print_usage()
```
greytdepression, line 151: `return` otherwise it will try to run despite invalid arguments

jakttest/jakttest.jakt
```diff
@@ -59,13 +63,96 @@ function strip_line_breaks(anon input: String) throws -> String {
...
+                let temp_dir = strip_prefix(args[index], prefix: "--temp-dir")
+                if temp_dir.has_value() {
+                    if temp_dir.value().byte_at(0) != b'=' {
+                        options.errors.push("--temp-dir given without a value")
+                        // skip it so the rest of arguments can be validated
+                        index++
+                        continue
+                    }
+
+                    options.temp_dir = temp_dir!.substring(start: 1, length: temp_dir!.length() - 1)
```
greytdepression, line 117: maybe we could check if the dir name starts with `"` or `'` and remove those too?

cybergsus: Getting the quote to be inside the dirname from the shell arguments is actually difficult to pull off. When you use `'` or `"` in a shell argument what you mean is that the inside of the quotes should be a single argument: `git commit -m 'hello world'` makes `hello world` the third argument

cybergsus: So when you do `jakttest --temp-dir="hello world" sample.jakt`, `args` is: `["jakttest", "--temp-dir=hello world", "sample.jakt"]`.

greytdepression: Oh yeah, true. :^)

cybergsus:
It currently has some issues at the moment, namely that the lockfiles seem to not be enough to ensure jobs don't step over each other, which is currently making the result counts variable.

AtkinsSJ:
It might be easier to coordinate threads if jakttest itself gathered up and ran all the tests instead of a shell script. Not sure if Jakt has good enough multithreading support for that though.

cybergsus:
I might be able to hack in some C++ code using `std::thread` or `fork()`&`exec()`s, but Jakt doesn't have support for anything besides prelude stuff (File, Array, Optional...)

cybergsus:
The reason behind using "locks" is to correctly map jobs that are freed to their directory. I'm thinking of having a sort of associative array with processID+directory name. Although I'm not comfortable with Bash arrays, so this will take me a while to think through.

cybergsus:
I'm going to do this as a single commit; trying to do multiple incremental commits breaks the tests since I'm modifying the test runner.

cybergsus:
image
It seems like it's leaving everything empty... I'm pretty curious on why though

cybergsus:
Hmm, looks like selfhost is trying to use `build/output` as its temporary build object instead of using/reusing its own temporary directory. I'll add a flag so build output is customizable through an argument.

cybergsus:
Welp, I managed to fit in some multiprocessing support for Jakt :)

cybergsus:
Wait, this seems to be producing false negatives:
main
main+new jakttest

jntrnr:
I tried running the tests on the mac, and seeing:

```
     Running `/Users/jt/Source/jakt/target/debug/jakt -o build --emit-cpp-source-only jakttest.jakt`
[2/3] sed -e '1a#include "../process.h...ttest.cpp > build/jakttest-patched.cpp
FAILED: build/jakttest-patched.cpp
sed -e '1a#include "../process.h"' build/jakttest.cpp > build/jakttest-patched.cpp
sed: 1: "1a#include "../process.h"
": command a expects \ followed by text
ninja: build stopped: subcommand failed.
```

cybergsus:
Hmmm, different `sed` versions I see... Welp, I'll just make a quick python patch script then :)

PR #823, opened 2022-07-05
selfhost/typechecker: Make `samples/ranges/range.jakt` pass

ntg28:
Make the sample generate code by setting up all the structure generics into
`generic_inferences` and call `substitute_typevars_in_type` for the function
return type


awesomekling:
Had to revert this in bdd49bf747c1080bcff9384e91451799236bccf9 as it made `run-all.sh` hang.

PR #824, opened 2022-07-06
jakttest: Build selfhost if binary does not exist

0verse:
run-all.sh script doesn't check for if selfhost binary is
already built when comparing if it is updated or not. This
results with tests failing without ever using the selfhosted
binary. Building selfhost before running the tests fixes
the issue.

### DIFF
--- a/jakttest/run-all.sh
+++ b/jakttest/run-all.sh
@@ -5,6 +5,11 @@ if [ ! -x ./build/jakttest ]; then
     cargo run jakttest/jakttest.jakt
 fi
 
+if [ ! -x ./build/main ]; then
+    echo "selfhost binary does not exist. Building it now."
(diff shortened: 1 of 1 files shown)


cybergsus:
LGTM, good catch ;). Maybe a better name for the commit would be `Build selfhost if binary does not exist`.

AtkinsSJ:
> Yeah, yours is much clear to understand. I would change the title but I'm new to git to do complicated changes.

To change the previous commit's title and message, you use `git commit --amend`. Then when you're done, `git push --force` to replace the existing commit here. :^)

0verse:
Thanks @AtkinsSJ :)

AtkinsSJ:
You're welcome!

PR #833, opened 2022-07-08
selfhost/typechecker+codegen: Make `optional/none_coalescing.jakt` pass

ntg28:


jntrnr:
Since you both preferred #834 I'll land that and close this. Thanks all

PR #839, opened 2022-07-09
selfhost/typechecker: Allow `T` assignment to `T?`

ntg28:
Allow assignment of a value to a `Optional` of the same type


jntrnr:
Does this allow any additional tests to pass?

If not, can you add a test to show this is working?

awesomekling:
This makes samples/optional/assignment.jakt pass :)

PR #842, opened 2022-07-09
selfhost/lexer: Re-add error messages in number lexing lost in rebase

mnlrsn:
Commit #138186a erroneously erased two error messages in the lexing of
numbers. Re-introduced with this fix. Test for this can be added later so as 
not to change the number of tests used to track the progress for the
selfhosted compiler.


awesomekling:
FWIW it's fine to add tests. Don't worry about the tracking number :)

PR #843, opened 2022-07-09
selfhost/typechecker: Fix function return typecheck

adleroliveira:
This fix some bugs on function return type typeckecking and adds a PASS for the following tests:

**tests/typechecker/return_value_type_mismatch.jakt
tests/typechecker/function_return_type_with_fat_arrow.jakt**

Previously, when going through the typechecking namespaces declaration, the current_function_id was not being updated so the typechecker didn't got the return type hint correctly.

Also, in order for this to work correctly I created the function unify_with_type based on the same function on the Rust compiler. This function properly check for type compatibility.

There is a unify function that just compares type_id's and is being used on several places as a placeholder. We should probably start replacing it for the new one.

Before:
273 passed
60 failed
7 skipped

After
275 passed
58 failed
7 skipped

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -1431,6 +1431,22 @@ struct Typechecker {
         }
     }
 
+    function unify_with_type(mut this, found_type: TypeId, expected_type: TypeId?, span: Span) throws -> TypeId {
+        if expected_type.has_value() {
+            if expected_type!.equals(unknown_type_id()) {
+                return found_type
+            }
+            let generic_inferences: [String:String] = [:]
+            if .check_types_for_compat(lhs_type_id: expected_type!, rhs_type_id: found_type, generic_inferences, span) {
+                return found_type
+            }
+
+            return .substitute_typevars_in_type(type_id: found_type, generic_inferences)
+        } else {
+            return found_type
+        }
+    }
+
     function find_or_add_type_id(mut this, anon type: Type) throws -> TypeId {
         let module = .program.modules[.current_module_id.id]
         let module_id = module.id
@@ -2235,7 +2251,9 @@ struct Typechecker {
         }
 
         for fun in parsed_namespace.functions.iterator() {
+            .current_function_id = .find_function_in_scope(parent_scope_id: scope_id, function_name: fun.name)
             .typecheck_function(parsed_function: fun, parent_scope_id: scope_id)
+            .current_function_id = None
         }
     }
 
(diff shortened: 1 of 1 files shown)


selfhost/typechecker.jakt
```diff
@@ -1431,6 +1431,22 @@ struct Typechecker {
...
+                let generic_inferences: [String:String] = [:]
+                if .check_types_for_compat(lhs_type_id: expected_type!, rhs_type_id: found_type, generic_inferences, span) {
+                    return found_type
+                }
+
+                return .substitute_typevars_in_type(type_id: found_type, generic_inferences)
+        } else {
+            return found_type
+        }
+    }
```
awesomekling, line 1448: Please use 4 spaces for indentation.

selfhost/typechecker.jakt
```diff
@@ -2697,7 +2715,7 @@ struct Typechecker {
         }
 
         parsed_function.must_instantiate = true
-
+
```
awesomekling, line 2718: Unrelated whitespace change.

selfhost/typechecker.jakt
```diff
@@ -2896,7 +2914,7 @@ struct Typechecker {
                                 "Type mismatch: expected ‘{}’, but got ‘{}’"
                                 .type_name(TypeId::from_string(seen_type_id_string.value()))
                                 .type_name(rhs_type_id)
-                            )
+                            ),
```
awesomekling, line 2917: Unrelated change. Also comma is not required here.
There are a whole bunch of unrelated changes like this in the commit, please remove them :)

selfhost/typechecker.jakt
```diff
@@ -3247,7 +3265,7 @@ struct Typechecker {
                     .check_types_for_compat(
                         lhs_type_id: checked_block.yielded_type.value()
                         rhs_type_id: type_
-                        generic_enferences: generic_inferences
+                        generic_inferences: generic_inferences
```
awesomekling, line 3268: Unrelated change. (Also, the fact that this worked before is probably a bug on its own!)

awesomekling (changes requested):

selfhost/typechecker.jakt
```diff
@@ -4175,7 +4193,10 @@ struct Typechecker {
         }
         SingleQuotedString(val, span) => CheckedExpression::CharacterConstant(val, span)
         SingleQuotedByteString(val, span) => CheckedExpression::ByteConstant(val, span)
-        QuotedString(val, span) => CheckedExpression::QuotedString(val, span)
+        QuotedString(val, span) => {
+            .unify_with_type(found_type: TypeId::from_string("0_13"), expected_type: type_hint, rhs_span: span)
```
awesomekling, line 4197: What is the significance of `"0_13"` here? There is no comment, or explanation in the commit message.

adleroliveira: replaced it for builtin(BuiltinType::String) which represents the same thing. Much more readable now.

awesomekling: Great!

PR #849, opened 2022-07-10
selfhost/codegen: Codegen WeakPtr type

onsah:
Adds codegen for `WeakPtr` built in type.

Fixes `assign_to_weak_on_declaration` test

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -911,6 +911,8 @@ class CheckedProgram {
         return .modules[id.module_id.id].scopes[id.id]
     } 
 
+    public function prelude_scope_id(this) -> ScopeId => ScopeId(module_id: ModuleId(id: 0), id: 0)
+
     public function set_loaded_module(mut this, module_name: String, loaded_module: LoadedModule) throws {
         .loaded_modules.set(
             key: module_name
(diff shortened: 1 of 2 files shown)


tests/typechecker/field_access_on_enum.jakt
```diff
@@ -1,5 +1,5 @@
 /// Expect:
-/// - error: "Member field access on value of non-struct type ‘E’"
+/// - Error: "Member field access on value of non-struct type ‘E’"
```
jntrnr, line 2: you'll want to leave this as `error` as it's part of the unit testing framework

onsah: Oh I committed this change by mistake. Fixing it.

selfhost/typechecker.jakt
```diff
@@ -964,6 +966,69 @@ class CheckedProgram {
     public function get_bits(this, anon type_id: TypeId) => .get_type(type_id).get_bits()
 
     public function is_signed(this, anon type_id: TypeId) => .get_type(type_id).is_signed()
+
+    public function find_struct_in_scope(this, scope_id: ScopeId, name: String) throws -> StructId? {
```
onsah, line 970: I am open to suggestions for naming. But having this function separately is nice because it handles all the necessary verifications if the type is weak ptr.

PR #859, opened 2022-07-11
jakttest: Generate diffs from different runs

cybergsus:
This allows us to know if any regressions have ocurred or if any extra tests
have been fixed by new changes.


cybergsus:
Remove the `sh` wraps since this problem also happens on the main branch; therefore it wasn't introduced by this.

cybergsus:
Closing since the cleaner ninja-like output makes it much easier to visually parse what has changed

PR #860, opened 2022-07-11
selfhost/typechecker: Output type mismatch error after generic type p…

onsah:
…arameter number mismatch error

* Output `Type mismatch` after `mismatched number of generic parameters for T` error
* Extract `Type mismatch` error to a helper function

This gets us closer to pass `constructor_with_different_generic_instance.jakt`. But because currently `unknown` and `void` points to the same type the error message is not exactly as wanted. This change now additionally outputs
```
Error: Type mismatch: expected ‘B’, but got ‘[void]’
----- tests/typechecker/constructor_with_different_generic_instance.jakt:12:13
 11 | function main() { 
 12 |     mut a = A(b: []) 
                  ^- Type mismatch: expected ‘B’, but got ‘[void]’
 13 | } 
-----
```
The expected output is `Type mismatch: expected ‘B’, but got ‘[unknown]’`

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -3015,6 +3015,21 @@ struct Typechecker {
         else => false
     }
 
+    function error_type_mismatch(mut this, lhs_type_id: TypeId, rhs_type_id: TypeId, span: Span) throws {
+        .error(
+            format(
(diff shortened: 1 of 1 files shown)


awesomekling:
This has conflicts.

AtkinsSJ:
Closing this since it's had conflicts for 2 months with no further work.

PR #863, opened 2022-07-11
jakttest: Fix bad use of waitpid()

cybergsus:
waitpid() doesn't return the exit code on *wstatus*.
Exit code is returned by the macro WEXITSTATUS.

### DIFF
--- a/jakttest/process.cpp
+++ b/jakttest/process.cpp
@@ -81,8 +81,8 @@ ErrorOr<i32> start_background_process(Array<String> args)
 
 ErrorOr<Optional<i32>> poll_process_exit(i32 pid)
 {
-    i32 exit_code;
-    auto const result = waitpid(pid, &exit_code, WNOHANG);
+    i32 wstatus;
+    auto const result = waitpid(pid, &wstatus, WNOHANG);
     if (result == -1) {
         return Error::from_errno(errno);
     }
@@ -92,7 +92,7 @@ ErrorOr<Optional<i32>> poll_process_exit(i32 pid)
     }
     // NOTE: compiler seems to struggle to find a constructor when
     // nesting error & optional
-    return Optional<i32>(exit_code);
+    return Optional<i32>(WEXITSTATUS(wstatus));
 }
 
 ErrorOr<void> forcefully_kill_process(i32 pid)


jakttest/process.cpp
```diff
@@ -92,7 +92,7 @@ ErrorOr<Optional<i32>> poll_process_exit(i32 pid)
     }
     // NOTE: compiler seems to struggle to find a constructor when
     // nesting error & optional
-    return Optional<i32>(exit_code);
+    return Optional<i32>(WEXITSTATUS(wstatus));
```
ADKaster, line 95: This isn't sufficient. You need to check WIFEXITED before calling WEXITSTATUS. it's possible the process was signaled (abort, panic, ...) Instead of exiting normally.

cybergsus: Hmm, I didn't think we'd care if the test has crashed. Maybe I should make `poll_process_exit` return a `bool` telling whether it was signaled?

ADKaster: That is, if you get stopped/continued, you need to keep waiting for that pid until it's *really* dead

jakttest/process.cpp
```diff
@@ -92,7 +92,7 @@ ErrorOr<Optional<i32>> poll_process_exit(i32 pid)
     }
     // NOTE: compiler seems to struggle to find a constructor when
     // nesting error & optional
-    return Optional<i32>(exit_code);
+    return Optional<Tuple<bool, i32>>(Tuple { WIFEXITED(wstatus) != 0, WEXITSTATUS(wstatus) });
```
bgianfo, line 95: Return a real struct with named members instead of (ab)using Tupe?

cybergsus: I can't think of a name for such a structure. Can you suggest one?

jakttest/process.cpp
cybergsus: Enums are hard to get from C++ into Jakt, since Jakt enums work with variants. It's not easy to have transparent types between C++ and Jakt and have that amount of information given. Best I can think of is making the returned code be the signal if `WIFSIGNALED` is set.

jakttest/process.cpp
cybergsus: Nevermind, I could use a `RawExitStatus` struct that contains this info and can be used by Jakt code.

jakttest/process.cpp
cybergsus: I'm going to revert all changes on this PR beyond just returning the exit code. The scheduler doesn't care if the process exited normally or abruptly right now, it just wants to know if it's in the exited state so it can fit a new task. If signal handling is deemed necessary later we can just land a new PR.

cybergsus:
I'm going to revert all changes besides *just returning the exit code*. The scheduler only cares about whether the process exited. The only reason the exit code is to have an easier to use API, making space for features that were already planned (like #864).

PR #864, opened 2022-07-11
jakttest: Add --show-reasons option

cybergsus:
This options shows the reasons why a test or multiple tests failed. To
do this it makes run-one.sh return different codes depending on the
stage that failed, and direct stderr of each stage to different
files. This needs #863 merged to run correctly, since it uses the
exit code that was badly returned.


awesomekling:
This has a conflict in `jakttest/run-one.sh`

PR #872, opened 2022-07-12
selfhost: Separate unknown from void

onsah:
Currently both `unknown` and `void` types have the same id. This causes ambiguity because one can't know what was the intended meaning of the value.

This PR does following:
* Introduce new `Type` variant `Unknown`
* Fix the tests broken by this change.
  * Fix checking for `unknown` in enum declaration
  * Fix not generating code for enum match when enum has type `void`

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -816,7 +816,7 @@ struct CodeGenerator {
     function codegen_enum_predecl(mut this, enum_: CheckedEnum) throws -> String {
         mut output = ""
 
-        if not enum_.underlying_type_id.equals(unknown_type_id()) {
+        if not enum_.underlying_type_id.equals(void_type_id()) {
             if .program.is_integer(enum_.underlying_type_id) {
                 return format("enum class {}: {};", enum_.name, .codegen_type(enum_.underlying_type_id))
             } else {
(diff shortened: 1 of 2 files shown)


selfhost/codegen.jakt
```diff
@@ -1148,7 +1148,7 @@ struct CodeGenerator {
                 else => {
                     panic("Internal error: range expression doesn't have Range type")
                     // FIXME: panic() should be "noreturn" and this can be removed
-                    yield TypeId(module: ModuleId(id: 0), id: 0)
+                    yield TypeId(module: ModuleId(id: 0), id: 16)
```
onsah, line 1151: This should point to the `Unknown` builtin type if possible.

jntrnr:
Can you add `Unknown` to `BuiltinType` and use that instead of using the numbers?

But yeah, good catch. unknown should not be pointing at void's type id.

selfhost/codegen.jakt
onsah: nvm: Realized we can just use static functions from typechecker.

selfhost/typechecker.jakt
```diff
@@ -889,7 +890,7 @@ struct CheckedCall {
     callee_throws: bool
 }
 
-function unknown_type_id() -> TypeId => TypeId(module: ModuleId(id: 0), id: 0)
+function unknown_type_id() -> TypeId => TypeId(module: ModuleId(id: 0), id: 16)
```
jntrnr, line 893: is it possible to say `TypeId(module: ModuleId(id: 0), id: BuiltinType::Unknown)` here so we don't have to remember that it's `16`?

onsah: Yeah I think it's possible. Let me try.

PR #883, opened 2022-07-14
jakttest: Remove run-all.sh from Jakttest

cybergsus:
This PR implements a series of steps to remove the `jakttest/run-all.sh` script
and make `jakttest` the entrypoint. These are the things that `jakttest` should
do by itself:

- [x] Discover tests by itself if not given explicitly
- [x] Discover the number of jobs if not given explicitly [^1]
- [x] Make a temporary directory and clean it up unless given explicitly
- [x] Re-compile selfhost if it's out of date
- [x] Re-compile jakttest & re-exec automatically if updated

[^1]: These are achieved through `sysconf`, which is specified by POSIX.1-2008.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
           toolchain: ${{ matrix.rust }}
           override: true
       - name: Test Jakt Selfhost
-        run: ./jakttest/run-all.sh
+        run: ninja -C jakttest && ./jakttest/build/jakttest
   clippy:
     strategy:
       fail-fast: true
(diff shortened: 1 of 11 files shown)


cybergsus:
After looking at the docs, seems like `sched_getaffinity` (what I've used so that jakttest knows the number of CPUs) is only available in Linux.

ADKaster:
how about `sysconf(_SC_NPROCESSORS_CONF)` ?

cybergsus:
oh, that's a good one. Thanks!

jntrnr:
@cybergsus - trying this out locally and I see this:

```
     Running `/home/jt/Source/jakt/target/debug/jakt -o build --emit-cpp-source-only jakttest.jakt`
fs
fs
fs
fs
fs
fs
fs
fs
fs
fs
fs
```

Any idea what that's coming from?

cybergsus:
I traced it back a while ago: https://github.com/SerenityOS/jakt/blob/3c81d4018b75ab52731286e2c9b58650f877e359/src/codegen.rs#L2001

cybergsus:
You might have not stumbled upon it while developing selfhost because selfhost doesn't use namespaces to declare external C++ stuff. It just uses modules to organize the Jakt code.

PR #889, opened 2022-07-15
misc: Add copyright & license notices to all files under jakttest/

cybergsus:
I used git blame to get other contributors' names; but there might be
some cases I have missed.


awesomekling:
Please don't add anyone but yourself.

Jonta:
A bunch of the changes look unrelated to the PR-title

cybergsus:
OK, I updated everywhere that I added the license notice to only include myself. I don't really like it since I feel at least Andreas and JT should be in there, but if you say so, I'll just add myself and let you guys decide whether you want to add yourselves to the list :^)

PR #897, opened 2022-07-17
jakttest: Make Jakttest not poll for processes

cybergsus:
Instead of polling for every running test and spinning, we call
sigwait() when we don't have anything to run, so that Jakttest's process
is suspended until any child exits (SIGCHLD). When it wakes up it does
generic polls with waitpid(-1, ...) to process all the tests that
finished executing and mark their directories as freed, launch more test
processes and go back to sleep.

This way we avoid spinning which gets in the way of the N jobs running
clang++ that are compiling the generated C++ from running the tests.

On my system (i7-8550U, 4 physical cores, 8 processing units) this shaved around
10 seconds, which means Jakttest was previously getting in the way a LOT.

cc @alimpfard @ADKaster

### DIFF
--- a/jakttest/jakttest.jakt
+++ b/jakttest/jakttest.jakt
@@ -394,6 +394,12 @@ enum TestFailedReason {
     AbruptExit(i32)
 }
 
+enum TestExitedResult {
+    Passed
+    Failed(file: String)
+}
+
+
 // A test scheduler that has its process rate limited by
 // the number of directories it has available.
 struct TestScheduler {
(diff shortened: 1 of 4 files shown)


cybergsus:
Apart from resolving the merge conflict, the scheduler should temporarily `sigignore(SIGCHLD)` while it's processing the finished tests' results, because that checking method is **not** signal safe.

cybergsus:
thinking about the above, it doesn't matter if it ignores SIGCHLD or not when processing the finished tests' results; since the signal handler that we set (`[](int){}`) will exit and return back to what was being done before the signal was received.

cybergsus:
While merging I've caught a bug in codegen: Turns out that it doesn't get that `continue` is not allowed inside a function,  for some reason.

PR #909, opened 2022-07-18
selfhost/typechecker: Make `samples/enums/simple_match.jakt` pass

ntg28:


jntrnr:
@ntg28 - oops, I landed this because you said that the test passes. I'm trying it now but don't see the test passing

ntg28:
I mean to say that the test passes the type checking but still does not compile

PR #910, opened 2022-07-18
selfhost: fix number_types sample

miller-time:
I wrote this small c++ program because I was curious what was happening:

```cpp
#include <cstdint>
#include <stdint.h>
#include <stdio.h>

int main(int argc, char **argv) {
  printf("INT64_MIN: %ld\n", INT64_MIN);
  int64_t small_number = -9223372036854775808;
  printf("small number: %ld\n", small_number);

  printf("INT64_MAX: %ld\n", INT64_MAX);
  int64_t large_number = 9223372036854775808;
  printf("large number: %ld\n", large_number);
}
```

```
➜ clang++ main.cpp 
main.cpp:7:27: warning: integer literal is too large to be represented in a signed integer type, interpreting as unsigned [-Wimplicitly-unsigned-literal]
  int64_t small_number = -9223372036854775808;
                          ^
main.cpp:11:26: warning: integer literal is too large to be represented in a signed integer type, interpreting as unsigned [-Wimplicitly-unsigned-literal]
  int64_t large_number = 9223372036854775808;
```

```
➜ ./a.out 
INT64_MIN: -9223372036854775808
small number: -9223372036854775808
INT64_MAX: 9223372036854775807
large number: -9223372036854775808
```

so creating the constant `9223372036854775808` and assigning it to `int64_t` is "wrong", but it works if you negate!

I came up with the idea for this patch from my VSCode tooltip for `INT64_MIN`:

```cpp
#define INT64_MIN (-__INT64_C(9223372036854775807) - 1)
```


miller-time:
note that the Rust compiler was generating c++ that produced a warning:

```
➜ cargo run samples/math/number_types.jakt
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
     Running `target/debug/jakt samples/math/number_types.jakt`
projects/jakt/build/number_types.cpp:19:49: warning: integer literal is too large to be represented in a signed integer type, interpreting as unsigned [-Wimplicitly-unsigned-literal]
        const i64 small_i64 = static_cast<i64>(-9223372036854775808LL);
```

so there is still the question:

what should the Rust compiler do when it encounters `let small_i64 = -9_223_372_036_854_775_808`?
    * should it continue to generate c++ with a warning?
    * would it be better to panic?
    * what about generating c++ code that does the equivalent of `let small_i64 = -9_223_372_036_854_775_807 - 1`?

the selfhost compiler should probably behave the same as the Rust compiler, so we'll want to apply the same fix there too.

PR #925, opened 2022-07-19
Proposal: Add support for destructuring assigment

adleroliveira:
This is a proposal for implementing basic support for destructuring assignment of tuples. It passes the test I created.

For now, I don't know if this breaks any other test since jakttest is not working on Mac. I will leave this PR as a draft until jakttest if fixed.


jntrnr:
How is the rust based testing skipping this one?

I just landed the PR that let's you mark tests as selfhost-only

adleroliveira:
> How is the rust based testing skipping this one?
> 
> I just landed the PR that let's you mark tests as selfhost-only

It is in a new directory, which I understand that the Rust compiler doesn't know about. I can use the new method if you prefer.

adleroliveira:
I moved the test to the original directory and added the "selfhost-only" to the test. I think we are ready to land this.

PR #928, opened 2022-07-19
jakttest: Fix struct stat member naming on MacOS

adleroliveira:
Although it is supposed to follow POSIX.1-2008, struct stat in MacOS
headers doesn't have conforming names for struct stat timing members
(st_atime, st_mtime, st_ctime); leaving with us with no choice but to
patch our own headers to implement a fix that should've happened 16
years ago.

### DIFF
--- a/jakttest/fs.cpp
+++ b/jakttest/fs.cpp
@@ -7,6 +7,11 @@
 #include <Jakt/StringBuilder.h>
 #include <fcntl.h>
 #include <sys/stat.h>
+#if defined(__APPLE__)
+#define st_atim st_atimespec
+#define st_ctim st_ctimespec
+#define st_mtim st_mtimespec
+#endif
 
 namespace Jakt::fs {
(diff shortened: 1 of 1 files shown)


drunderscore:
Instead of the commit title simply saying "Fix x on y" -- describe the change made in the title. The rationale can go in the message :^)

cybergsus:
jakttest: Fix `struct stat` member declaration renaming on MacOS

Although it is supposed to follow POSIX.1-2008, `struct stat` in MacOS headers doesn't have a consistent naming for `struct stat` timing members (`st_atime`, `st_mtime`,...); leaving with us with no choice but to patch our own headers to implement a fix that should've been there 16 years ago.

adleroliveira:
Thanks @cybergsus 
I updated the title and description with your suggestion to make it more understandable

adleroliveira:
My bad. I amended the commit to reflect the same message.
Thank you again

PR #930, opened 2022-07-20
selfhost/codegen: Update seen_types to store TypeId

robryanx:
There were collisions in type ids due to seen_types not also storing the
ModuleId, resulting in no codegen for some enums/structs


robryanx:
Nevermind with this one, needs more work.

PR #933, opened 2022-07-20
selfhost/main: Create build dir before writing into it

miller-time:
If you do not have a build directory, the selfhost compiler gets an error if it tries to write into that directory.

```
➜ jakt_selfhost -b -R ~/projects/jakt/runtime hello_world.jakt
Runtime error: Error(code=2)
```


selfhost/main.jakt
```diff
@@ -154,6 +154,7 @@ function main(args: [String]) {
         let basename_without_extension = file_path.basename().substring(start: 0, length: file_path.basename().length() - 5)
         let cpp_filename = binary_dir + "/" + basename_without_extension + ".cpp"
         let output_filename = binary_dir + "/" + basename_without_extension
+        mkdir(binary_dir)
```
miller-time, line 157: This is an unknown function, even though I attempted to add `mkdir` to the prelude. I need help :sweat_smile:

ADKaster: You didn't add  `#include <IO/fs.cpp>` to `lib.h`. The runtime is all compiled as one massive translation unit still.

miller-time: Thanks! After adding that, there's a different error. CI shows the same thing as I see when I run it locally:

```
/usr/include/x86_64-linux-gnu/sys/stat.h:317:12: note: candidate function not viable: requires 2 arguments, but 1 was provided
extern int mkdir (const char *__path, __mode_t __mode)
           ^
```

miller-time: yeah I mean, that's what I figured. I copied this from jakttest and was confused when I saw that because it seems to work in jakttest...?

miller-time: Alright, I think the fs.h/fs.cpp thing was incorrect. All I needed to do was have the extern functions listed in prelude.jakt. 

I was getting: `projects/jakt/build/main.cpp:27673:17: error: use of undeclared identifier 'mkdir'`

I fixed that by adding `#include <sys/stat.h>` (though maybe not in the right place)

The only part left is figuring out how to represent the `mode` parameter in jakt

selfhost/main.jakt
```diff
@@ -154,6 +154,7 @@ function main(args: [String]) {
         let basename_without_extension = file_path.basename().substring(start: 0, length: file_path.basename().length() - 5)
         let cpp_filename = binary_dir + "/" + basename_without_extension + ".cpp"
         let output_filename = binary_dir + "/" + basename_without_extension
+        mkdir(binary_dir.c_string(), 0)
```
miller-time, line 157: this 0 allows it to compile, but I need to figure out how to represent something like `0777`. also the extern function in the prelude calls the parameter a `c_int` which might be wrong.

miller-time:
the signature of mkdir on Windows is different :sweat:

jntrnr:
@miller-time - still seems it's not quite finding the right mkdir

runtime/IO/File.h
```diff
@@ -10,7 +10,9 @@
 #include <Jakt/Error.h>
 #include <Jakt/RefCounted.h>
 #include <Jakt/String.h>
+#include <dirent.h>
 #include <stdio.h>
+#include <sys/stat.h>
```
ADKaster, line 15: you need to guard these additions with `#ifndef _WIN32`, because on Windows, mkdir comes from `<direct.h>`

But more than that, the includes should go into the .cpp file because they are only used  in the implementation. Nothing in the public interface needs either of these. 

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/mkdir?view=msvc-170>
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/mkdir-wmkdir?view=msvc-170

```c++
#ifdef _WIN32
#    include <direct.h>
#else
#    include <dirent.h>
#    include <sys/stat.h>
#endif
```

runtime/prelude.jakt
```diff
@@ -131,6 +131,10 @@ extern class File {
     public function exists(anon path: String) -> bool
 }
 
+import extern "IO/File.cpp" {
+    extern function mkdir(anon path: String) throws
+}
```
ADKaster, line 136: This should import from the header file, not the cpp.

However, since this is the prelude, and the prelude already includes the IO/File.h header it should instead just be a free function

```suggestion
extern function mkdir(anon path: String) throws
```

runtime/IO/File.cpp
```diff
@@ -92,4 +92,11 @@ bool File::exists(String path)
     return true;
 }
 
+ErrorOr<void> mkdir(String path)
+{
+    if (::mkdir(path.c_string(), S_IRWXU) == -1)
```
ADKaster, line 97: the 2 argument function does not exist on Windows. Therefore, you can either

```c++
#ifdef _WIN32
   if (::mkdir(path.c_string()) == -1)
#else
  if (::mkdir(path.c_string(), S_IRWXU) == -1)
#endif
```

or reproduce the entire function for each platform, like so:

```
#ifdef _WIN32
ErrorOr<void> mkdir(String path)
{
    if (::mkdir(path.c_string()) == -1) {
        return Error::from_errno(errno);
    }
    return ErrorOr<void> {};
}
#else
ErrorOr<void> mkdir(String path)
{
    if (::mkdir(path.c_string(), S_IRWXU) == -1)
        return Error::from_errno(errno);
    return ErrorOr<void> {};
}
#endif
```

ADKaster (changes requested):
This is close, but needs the suggested changes to work on windows.

miller-time:
@ADKaster thanks so much for all the corrections! 

rebased the branch, squashed all the messy commits... 

It still says `Call to unknown function: ‘mkdir’` but that's happening during this step:

```
[3/5] Building jakt file selfhost/main.jakt with jakt_stage0
```

Does that mean jakt_stage0 compiled with my changes, but for some reason jakt_stage1 doesn't work?

ADKaster:
Actually it means that you need to copy your runtime changes over onto bootstrap/stage0/runtime, including the prelude changes. 

The stage0 compiler uses its own copy of the prelude to compile things when using the CMake build, which is distinct from the top level runtime directory used by the stage 1/2 compiler. Since we are using stage0 to compile stage1, the updated prelude functionality needs duplicated.

miller-time:
@ADKaster yeah I just noticed that there's another runtime there (right after posting that message).

I tried copying in my runtime changes, but the error persists. Did I miss something?

ADKaster:
You need to do this in stages, apparently. 

1) Add new methods to prelude
2) build compiler
3) copy generated stage1.cpp over stage0/jakt.cpp
4) *now* you can use the new methods in the compiler itself.

I don't think this is how this was intended to work, but that's the only way I could add a method to the prelude locally.

alimpfard:
We could read it at runtime I guess, if we assume its location relative to the runtime dir.

miller-time:
@ADKaster thanks, I think that got me a little closer...

```
jakt/build/jakt_stage1_main.cpp:55053:6: error: no matching function for call to 'mkdir'

/usr/include/x86_64-linux-gnu/sys/stat.h:389:12: note: candidate function not viable: requires 2 arguments, but 1 was provided
extern int mkdir (const char *__path, __mode_t __mode)
```

PR #935, opened 2022-07-20
jakt: Implement `import extern`

cybergsus:
Ports  the `import extern` feature to selfhost, which was in the process of
 being added before feature freeze. Thought out by @alimpfard, ported &
 refactored by me. Original semantics haven't changed.


selfhost/main.jakt
```diff
@@ -33,6 +33,9 @@ function help() -> String{
     output += "  -R,--runtime-path PATH\t\tPath of the Jakt runtime headers.\n\t\tDefaults to $PWD/runtime.\n"
     output += "  -o,--binary-dir PATH\t\tOutput directory for compiled files.\n\t\tDefaults to $PWD/build.\n"
     output += "  -C,--cxx-compiler-path PATH\t\tPath of the C++ compiler to use when compiling the generated sources.\n\t\tDefaults to clang++.\n"
+    output += "  -I PATH\t\tAdd PATH to compiler's include list. Can be specified multiple times.\n"
+    output += "  -L PATH\t\tAdd PATH to linker's search list. Can be specified multiple times.\n"
+    output += "  -l,--link-with LIB\t\tLink executable with LIB. Can be specified multiple times.\n"
```
alimpfard, line 38: `-l` is used for the lexer's token debug stream (I'd suggest changing that one instead).

samples/imports/import_extern.jakt
```diff
@@ -0,0 +1,51 @@
+/// Expect: selfhost-only
+/// - output: "PASS\n"
+
+import extern c "stdlib.h"  {
```
alimpfard, line 4: A C++ test would also be nice, though it doesn't actually work as it is - here's the test I had with my branch:
```
import extern "LibJS/Interpreter.h" {
    namespace AK {
        extern struct NonnullRefPtr<T> {
            function ptr(this) -> raw T
        }
        extern struct NonnullOwnPtr<T> {
            function ptr(this) -> raw T
        }
        extern struct StringView {
            function StringView(anon s: raw c_char) -> StringView
        }
        extern struct Vector<T> {
            function size(this) -> usize
            function at(this, anon index: usize) -> T
        }
        extern struct Result<T, E> {
            function is_error(this) -> bool
            function release_value(mut this) -> T
        }
    }
    namespace JS {
        namespace Parser {
            extern struct Error {
            }
        }
        extern struct GlobalObject {
        }
        extern struct Value {
            function as_i32(this) -> i32
        }
        extern struct VM {
            // This is a lie, this takes default parameters.
            function create() -> AK::NonnullRefPtr<JS::VM>
        }
        extern struct ThrowCompletionOr<T> {
            function is_error(this) -> bool
            function release_value(mut this) -> T
        }
        extern struct Realm {
            function create(anon vm: JS::VM) -> raw JS::Realm
        }
        extern struct Script {
            function parse(source: AK::StringView, anon realm: JS::Realm) -> AK::Result<AK::NonnullRefPtr<JS::Script>, AK::Vector<JS::Parser::Error> >
        }
        extern struct Interpreter {
            function create<GlobalObjectType>(anon vm: JS::VM) -> AK::NonnullOwnPtr<JS::Interpreter>
            // This is a lie, this is not const.
            function run(this, anon script: JS::Script) -> JS::ThrowCompletionOr<JS::Value>
        }
    }
}

function main() {
    unsafe {
        mut vm = JS::VM::create();
        mut realm = JS::Realm::create(*vm.ptr());
        let src = "42 + 27;"
        mut result = JS::Script::parse(source: AK::StringView(src.c_string()), *realm);
        if result.is_error() {
            return 1
        }

        mut script = result.release_value();
        mut interpreter = JS::Interpreter::create<JS::GlobalObject>(*vm.ptr());
        mut exec_result = (*interpreter.ptr()).run(*script.ptr());
        if exec_result.is_error() {
            return 2
        }

        let value = exec_result.release_value();
        println("{}", value.as_i32())
    }
}
```

selfhost/typechecker.jakt
```diff
@@ -1945,11 +1939,60 @@ struct Typechecker {
...
+                        record.name)
+                    record.name_span, 
+                    format("this {} may not be generic", record.record_type.record_type_name())
+                    record.name_span)
+            }
+        }
+
+
+
+        parent_namespace.namespaces.push(import_.assigned_namespace)
```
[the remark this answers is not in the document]
cybergsus: `parent_namespace` is a copy, but it copies the array reference for namespaces. I'll change the parameter since previously I was trying to  do more printing with `parent_namespace`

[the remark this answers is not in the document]
cybergsus: Changed it to passing the reference to the parent's children array. Tried running `typecheck_predecl` there but does not seem to be recognized by the rest of the typechecker.

cybergsus:
We'll have to wait for #940 so that the CI test passes, since this discovered a bug in the typechecker.

alimpfard:
See https://github.com/cybergsus/jakt/pull/1 for the fixes to make it work with the test I pasted :P

samples/imports/import_extern.jakt
cybergsus: That will need extending Jakttest to accept extra flags, and make sure we have serenity/lagom & js setup in CI to add that as a test. Otherwise it will fail with an include/link error from clang++

alimpfard: Right, that one will also need a serenity build, perhaps we could just use a regular STL thing to test this?
e.g. 
```
import extern "vector" {
    namespace std {
        extern struct vector<T> {
             extern function push_back(mut this, anon value: T)
             extern function back(this) -> T
        }
    }
}

function main() {
     let v = std::vector<i32>()
     v.push_back(69i32)
     if v.back() == 69i32 {
        println("PASS")
     }
}
```

cybergsus: nice, this triggered a couple of errors. Fixing those

cybergsus: Seems like using STL confuses clang++ about forward().

cybergsus: Got it to work by just initializing an empty vec and checking it's empty.

PR #941, opened 2022-07-21
prelude+typechecker: Add never type

lanmonster:
Functions that don't return can be marked as returning `never` so the
typechecker won't complain about not returning a value after calling
a never-returning function.

### DIFF
--- a/samples/control_flow/never_return.jakt
+++ b/samples/control_flow/never_return.jakt
@@ -1,0 +1,13 @@
+/// Expect: selfhost-only
+/// - output: ""
+
+function test() -> String {
+    if false {
+        return "hi"
+    }
+    abort()
+}
+
+function main() -> c_int {
+    abort()
+}
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -123,6 +123,7 @@ enum BuiltinType: usize {
     CChar = 14
     CInt = 15
     Unknown = 16
+    Never = 17
 }
 
 boxed enum Type {
@@ -143,6 +144,7 @@ boxed enum Type {
     CChar
     CInt
     Unknown
+    Never
     TypeVariable(String)
     GenericInstance(id: StructId, args: [TypeId])
     GenericEnumInstance(id: EnumId, args: [TypeId])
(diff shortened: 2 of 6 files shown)


samples/control_flow/never_return.jakt
```diff
@@ -0,0 +1,13 @@
...
+
+function test() -> String {
+    if false {
+        return "hi"
+    }
+    abort()
+}
+
+function main() -> c_int {
+    abort()
```
cybergsus, line 12: Wait. The fact that this test passes means Jakttest has a bug. It should error since the test aborts. Although it also suggests that we may need a `- no-run` flag so it only tries compiling it.

cybergsus: Or maybe Jakttest is OK as it is, and we shouldn't care about the exit code of the test.

selfhost/typechecker.jakt
```diff
@@ -3161,6 +3194,13 @@ struct Typechecker {
         While(condition, block) => block.definitely_returns
         Expression(expr) => match expr {
             Match() => checked_expression_definitely_returns(expr)
+            // Call(call) => {
+            //     mut definitely_returns = false
+            //     if call.function_id.has_value() {
```
cybergsus, line 3199: Please don't submit commented out code.

selfhost/typechecker.jakt
```diff
@@ -2811,9 +2820,14 @@ struct Typechecker {
 
         if not structure_linkage is External and not return_type_id.equals(VOID_TYPE_ID) and not block.definitely_returns {
             // FIXME: Use better span
-            .error("Control reaches end of non-void function", func.name_span)
+            if return_type_id.equals(never_type_id()) and not block.never_returns {
+                .error("Control reaches end of never-returning function", func.name_span)
```
cybergsus, line 2824: I suggest a different message, like "never-returning function should not return"

cybergsus (commented):
LGTM overall, just a couple nitpicks. If you can add "never return" detection from locked loops (`loop` with no `break`/`return`) and error if a function's main block that *should* return never returns, that'll be great, although you can do this in a separate if you wish.

selfhost/typechecker.jakt
lanmonster: whoops :^)

selfhost/typechecker.jakt
lanmonster: that message would not make sense for a void returning function

cybergsus: I'm talking about the message for a function that never returns, I thought it was clear by the context I'm giving.

PR #942, opened 2022-07-22
selfhost: Avoid underflowing when filename doesn't contain extension

sin-ack:
Because i was a usize (through String::length()) the condition i >= 0
would always be true and we'd hit an out-of-bounds error.

### DIFF
--- a/runtime/lib.h
+++ b/runtime/lib.h
@@ -82,24 +82,45 @@ struct Range {
     T end {};
     T current {};
     bool forwards { true };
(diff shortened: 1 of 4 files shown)


lanmonster (commented):
Now that we can iterate ranges backwards, can we keep i as usize and use a for loop instead of while?

sin-ack:
Had to implement end-inclusive `Range`s in the process, but should be good-to-go now.

PR #948, opened 2022-07-23
IDE: Porting the Rust-based IDE support to selfhost

jntrnr:
A port of the IDE support from the Rust-based compiler to the selfhost compiler.

Functionality is ported and connected to the commandline params. You can use it on simple tests, but more complex applications (namely selfhost itself) do not currently seem to yield info for hovers in all places it should.

Probably worth landing even though we're not at feature parity because it's big enough it'll bitrot quickly. We can continue to evolve it. With luck, we're only a few logic fixes away from being back to parity with the Rust-based compiler.

co-authored with: @cybergsus

### DIFF
--- a/selfhost/compiler.jakt
+++ b/selfhost/compiler.jakt
@@ -1,4 +1,4 @@
-import error { JaktError, print_error }
+import error { JaktError, print_error, print_error_json }
 import utility
 import utility { FilePath, FileId }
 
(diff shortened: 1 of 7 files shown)


cybergsus:
I wired some flags so I can get some feedback in my editor. (PR https://github.com/jntrnr/jakt/pull/1). I suggest that anyone that is also working on this changes their LSP's client `settings` to set `settings.compilerPath` to `<path/to/jakt>/build/main`, so that you can start seeing your own progress :)

PR #950, opened 2022-07-23
samples: Verify that arrays are initialisable with size from a function

mnlrsn:
A sample to show that a fat-arrow member function can be used to 
provide the size as an integer without indicating the type (e.g. via a
suffix), and hence must be inferred to be an integer. The initialiser of
the arrays must accept this integer. 

At the time of writing this works with the Rust-based compiler but 
not the self-hosted.

### DIFF
--- a/samples/arrays/array_size_from_function.jakt
+++ b/samples/arrays/array_size_from_function.jakt
@@ -1,0 +1,18 @@
+/// Expect:
+/// - output: "[0]\n"
+
+class C {
+    public e: [f64]
+
+    public function f(mut this) throws {
+        .e = [0f64; .n()]
+    }
+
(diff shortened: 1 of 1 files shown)


mnlrsn:
Ah, it won't accept failing tests. 

Issue created at #951 .

robryanx:
> Ah, it won't accept failing tests.
> 
> Issue created at #951 .

Yeah I think it would be useful to allow them. Probably need a flag to indicate that they are currently expected to fail.

mnlrsn:
Closing this pull request as it was intended to provide a test even if failing at the current state of development. Since there is no collection of tests for features that still need to be developed, it might as well be closed.

PR #953, opened 2022-07-23
jakttest: Fail if no explicit skip tag is set

cybergsus:
Now tests that have no skip tag and that have an incorrect
format (i.e Jakttest can't figure out what the test is expecting) are
not skipped but reported as failures.

### DIFF
--- a/jakttest/jakttest.jakt
+++ b/jakttest/jakttest.jakt
@@ -314,10 +314,11 @@ struct Options {
         }
 
         if options.files.is_empty() {
-            // find files under samples, tests, and selfhost
+            // find files under samples and tests
             get_jakt_files_under(directory: "samples",  results: &mut options.files)
             get_jakt_files_under(directory: "tests",    results: &mut options.files)
-            get_jakt_files_under(directory: "selfhost", results: &mut options.files)
+            // compile selfhost to check it can compile itself
+            options.files.push("selfhost/main.jakt")
         }
 
         if options.job_count == 0 {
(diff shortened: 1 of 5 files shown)


selfhost/parser.jakt
```diff
@@ -6,6 +6,7 @@
 //
 // SPDX-License-Identifier: BSD-2-Clause
 
+/// Expect: Skip
```
ADKaster, line 9: 🤔 this is uh. Kinda sus. We are putting expectations in the compiler source code? The tests and the compiler should not live in the same directory imo.

ADKaster (commented):
Seems okay, I'm not sure why we need expectation comments on the compiler source code though.

I also don't know what "dirty" means in this context. "Did not find Expectation or Skip directive in file {}", file sounds like an error the user could act on pretty quickly.

selfhost/parser.jakt
cybergsus: Hmm, maybe instead of adding everything under selfhost we just add selfhost/main.jakt

cybergsus: Adding selfhost dir to the tests is done so that Jakttest checks if selfhost can compile itself.

cybergsus:
This will need a revisit to remove the (explicit) check for `selfhost-only` that isn't needed anymore since we removed the Rust based compiler.

AtkinsSJ:
Just trying to tidy up some old PRs, and saw this has conflicts. Also I feel like this feature might have been implemented already now, but I've not been keeping track of what's going on with jakttest.

cybergsus:
Thanks for noticing :). Merge conflict resolved.

robryanx:
Guess it would be worth fixing the failing examples as a commit as part of this so the CI passes for it.

cybergsus:
Those changes were part of the original commit but I unknowingly rejected them for being outdated.

PR #954, opened 2022-07-23
debug: Add a newline before the closing bracket in debug prints

0x6D70:
Add a newline and proper indentation in front of every closing bracket
in debug prints.

For example till now debug prints looked like this:
```
Point2(
  x: Point(
    x: 10,
    y: 5),
  y: Point(
    x: 10,
    y: 5))
```
This PR fixes the issue by adding a new line and indentations in front of closing brackets. The result looks like this:
```
Point2(
  x: Point(
    x: 10,
    y: 5
  ),
  y: Point(
    x: 10,
    y: 5
  )
)
```

### DIFF
--- a/runtime/Jakt/StringBuilder.h
+++ b/runtime/Jakt/StringBuilder.h
@@ -106,6 +106,7 @@ struct Formatter<JaktInternal::Array<T>> : Formatter<StringView> {
                 TRY(string_builder.append(", "));
         }
         JaktInternal::_pretty_print_level--;
+        TRY(JaktInternal::_output_pretty_indent(string_builder));
         TRY(string_builder.append("]"));
         return Formatter<StringView>::format(builder, TRY(string_builder.to_string()));
     }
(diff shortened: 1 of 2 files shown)


cybergsus:
It may just be me, but putting the closing pairs in separate lines makes for worse debug output. In my compiler implementations I've gone through the hassle of writing a custom format function because half of the scrollback was just closing braces, it gets really frustrating. Here you can try telliing Jakt to dump the AST or checked AST, you'll see that the former approach makes for a smoother experience when trying to find the stuff you wanted in the representation.

0x6D70:
Yeah, at the end it is a personal preference. Personally, I prefer it when the output looks properly formatted.
But I also understand your point of view.

MacDue:
You can't do this nicely without also handling the empty braces cases, otherwise you get a lot of clutter (and more newlines) for each empty list or whatever. But I also like the current output (but I may be biased since I introduced it).

0x6D70:
Ok, maybe it should like it was for now. Thanks for the discussion!

PR #955, opened 2022-07-23
Runtime: Fix size check which caused an assertion

RealBlueta:
Fixes an assertion that kept happening

### DIFF
--- a/runtime/Jakt/String.h
+++ b/runtime/Jakt/String.h
@@ -39,7 +39,7 @@ public:
 
     char const& operator[](size_t i) const
     {
-        VERIFY(i < m_length);
+        VERIFY(i <= m_length);
         return c_string()[i];
     }
 


MacDue:
Where was this assertion happening? 
`VERIFY(i < m_length);` is the correct check here, `i <= m_length` would allow you index one passed the string (which would be the null-terminator here, so ok, but not really desired behaviour).

runtime/Jakt/String.h
```diff
@@ -39,7 +39,7 @@ class StringStorage : public RefCounted<StringStorage> {
 
     char const& operator[](size_t i) const
     {
-        VERIFY(i < m_length);
+        VERIFY(i <= m_length);
```
AtkinsSJ, line 42: This change isn't correct, it was right before. The character at `m_length` is the null byte, you should never be accessing that. If you are hitting an assertion, the problem is in your code or in the compiler, not here.

PR #957, opened 2022-07-23
selfhost: Add support for bindings with is operator

adleroliveira:
This implements variable bindings for enum variants for both
typed and struct-like variants. It de-sugars variable bindings
into typed optionals + pattern mathing with rebinding of inner
values followed by a forced unwrap.

Example:

```
enum Foo {
    Bar(i64)
}

let foo = Foo::Bar(5)

// This:
if foo is Foo::Bar(n) {
    println("{}", n)
}

// De-sugars to this:
mut tmp_n: i64? = None
if foo is Foo::Bar {
    match foo {
        Bar(n) => {
            tmp_n = n
        }
    }
    let n = tmp_n!
    println("{}", n)
}
```

### DIFF
--- a/selfhost/parser.jakt
+++ b/selfhost/parser.jakt
@@ -217,6 +217,7 @@ struct ParsedGenericParameter {
 
 struct ParsedBlock {
     stmts: [ParsedStatement]
+    bindings: [ParsedStatement]
 
     function find_yield_span(this) -> Span? {
         for stmt in .stmts.iterator() {
@@ -544,7 +545,13 @@ struct ParsedCall {
 
 boxed enum ParsedType {
     Name(name: String, span: Span)
-    NamespacedName(name: String, namespaces: [String], params: [ParsedType], span: Span)
+    NamespacedName(
+        name: String
+        namespaces: [String]
+        params: [ParsedType]
+        bindings: [EnumVariantPatternArgument]
+        span: Span
+    )
     GenericType(name: String, generic_parameters: [ParsedType], span: Span), // FIXME: tuple should be dictionary
     JaktArray(inner: ParsedType, span: Span)
     Dictionary(key: ParsedType, value: ParsedType, span: Span)
@@ -559,7 +566,7 @@ boxed enum ParsedType {
 
     function span(this) => match this {
         Name(name, span) => span
-        NamespacedName(name, namespaces, params, span) => span
+        NamespacedName(name, namespaces, params, bindings, span) => span
         GenericType(name, generic_parameters, span) => span
         JaktArray(inner, span) => span
         Dictionary(key, value, span) => span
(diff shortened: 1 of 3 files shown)


awesomekling (changes requested):
Typo: selhost -> selfhost
Typo: IS -> is
Typo: de-sugar -> de-sugars
Typo: bu -> by

The `tmp_n` in the desugared code should be scoped inside the `if` block, not outside it.

jntrnr:
Oh sweet!

Does `if foo is Foo::Bar(n)` also let you do things like: `if foo is Foo::Bar(n) and n > 100` ?

selfhost/typechecker.jakt
```diff
@@ -4435,6 +4438,155 @@ struct Typechecker {
         return CheckedStatement::If(condition: checked_condition, then_block: checked_block, else_statement: checked_else, span)
     }
 
+    // FIXME: Desugar of variable binding when using IS operator should probably be done in codegen stage
+    function expand_block_bindings(this, anon block: ParsedBlock, checked_condition: CheckedExpression, span: Span) throws -> ParsedBlock {
```
alimpfard, line 4442: I'm not quite sure what _precisely_ this desugars to, but you can directly codegen this, no need to involve a match
```
if foo is Bar(x: y) and expr {
    true_branch
}
```
(provided that `foo` is of an enum type)
```c++
if (auto _value = foo.get_pointer<FooDetails::Bar>()) {
    auto& y = _value.x;
    if (codegen.expr) {
        codegen.true_branch;
    }
}
```

for simplicity, `or` could simply be split up into multi `if` statements before entering the codegen stage.

selfhost/typechecker.jakt
```diff
@@ -4435,6 +4438,155 @@ struct Typechecker {
...
+        mut then_block = block
+        mut typed_bindings: [ParsedStatement] = []
+        mut match_expr: ParsedExpression? = None
+        mut unwrapping_declarations: [ParsedStatement] = []
+        mut var_name = ""
+        mut is_enum_exaustive = false
```
alimpfard, line 4448: typo: `is_enum_exaustive` -> `is_enum_exhaustive`

adleroliveira:
Converting it into a Draft while I work no the various comments.

selfhost/parser.jakt
```diff
@@ -544,7 +545,13 @@ struct ParsedCall {
...
 boxed enum ParsedType {
     Name(name: String, span: Span)
-    NamespacedName(name: String, namespaces: [String], params: [ParsedType], span: Span)
+    NamespacedName(
+        name: String
+        namespaces: [String]
+        params: [ParsedType]
+        bindings: [EnumVariantPatternArgument]
+        span: Span
+    )
```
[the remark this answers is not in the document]
adleroliveira: Where would be the right place then?

When parsing something like Foo::Bar(a) what is generated is a ParsedType::NamespacedName
This specific enum variant have name, namespaces, and params (for generics) where then would
you store the arguments?

adleroliveira:
Did a complete refactor of this and will open a new PR. Closing this one.

PR #964, opened 2022-07-24
RustFmt & Clippy: Improve readability, Update pico-args

LeoMeinel:
- Reformat rust code with RustFmt
- Fix some clippy warnings
- Update `pico-args` to `0.5.0`


awesomekling:
We are removing the rust-based compiler shortly, so let's not make feng shui changes to it. :)

PR #968, opened 2022-07-25
selfhost/main: Fix behaviour after unsuccessful builds

mattisboeckle:


robryanx:
Just to note for cleanup that is can be closed as it was implemented here: https://github.com/SerenityOS/jakt/commit/67bc8b31a517cf4dd99407b8568c1f3dd73bee0c

PR #969, opened 2022-07-25
runtime: Avoid some gcc specific errors

robryanx:
For better support with building with gcc as part of a serenity build

### DIFF
--- a/runtime/Jakt/Platform.h
+++ b/runtime/Jakt/Platform.h
@@ -42,7 +42,7 @@
 #ifdef ALWAYS_INLINE
 #    undef ALWAYS_INLINE
 #endif
-#define ALWAYS_INLINE
+#define ALWAYS_INLINE __attribute__((always_inline)) inline
 
 #ifdef NEVER_INLINE
 #    undef NEVER_INLINE
--- a/runtime/Jakt/String.h
+++ b/runtime/Jakt/String.h
@@ -19,6 +19,11 @@
 
 namespace Jakt {
 
+bool operator<(char const* characters, String const& string);
+bool operator>=(char const* characters, String const& string);
+bool operator>(char const* characters, String const& string);
+bool operator<=(char const* characters, String const& string);
+
 class StringStorage : public RefCounted<StringStorage> {
 public:
     static ErrorOr<NonnullRefPtr<StringStorage>> create_uninitialized(size_t length, char*& buffer);
--- a/runtime/Jakt/StringBuilder.h
+++ b/runtime/Jakt/StringBuilder.h
@@ -59,6 +59,8 @@ public:
                 append(separator);
             appendff(fmtstr, item);
         }
+
+        return {};
     }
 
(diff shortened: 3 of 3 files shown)


runtime/Jakt/Platform.h
```diff
@@ -42,7 +42,7 @@
 #ifdef ALWAYS_INLINE
 #    undef ALWAYS_INLINE
 #endif
-#define ALWAYS_INLINE
+#define ALWAYS_INLINE __attribute__((always_inline)) inline
```
ADKaster, line 45: This was intentionally removed I think. Why is it needed?

runtime/Jakt/String.h
```diff
@@ -19,6 +19,11 @@
 
 namespace Jakt {
 
+bool operator<(char const* characters, String const& string);
```
ADKaster, line 22: What code complains when these aren't here? Where are they implemented?

runtime/Jakt/StringBuilder.h
```diff
@@ -49,7 +49,7 @@ class StringBuilder {
     [[nodiscard]] bool is_empty() const { return m_buffer.is_empty(); }
 
     template<class SeparatorType, class CollectionType>
-    ErrorOr<void> join(SeparatorType const& separator, CollectionType const& collection, StringView fmtstr = "{}"sv)
+    void join(SeparatorType const& separator, CollectionType const& collection, StringView fmtstr = "{}"sv)
```
ADKaster, line 52: This looks wrong. Now ErrorOrs from append and appendff will be dropped on the floor -- which shouldn't be possible, ErrorOr is [[nodiscard]]

runtime/Jakt/Platform.h
robryanx: This error is from:

runtime/Jakt/Memory.h:19:20: error: no previous declaration for 'void fast_u32_copy(u32*, const u32*, size_t)' [-Werror=missing-declarations]
ALWAYS_INLINE void fast_u32_copy(u32* dest, u32 const* src, size_t count)

runtime/Jakt/Memory.h:30:20: error: no previous declaration for 'void fast_u32_fill(u32*, u32, size_t)' [-Werror=missing-declarations]
ALWAYS_INLINE void fast_u32_fill(u32* dest, u32 value, size_t count)

runtime/Jakt/String.h
robryanx: This error for each of them:

runtime/Jakt/String.cpp:174:6: error: no previous declaration for 'bool Jakt::operator<(const char*, const String&)' [-Werror=missing-declarations]

runtime/Jakt/StringBuilder.h
robryanx: Yeah I think this is the wrong fix looking now, it probably just needs a return at the end, I think `return {};`

robryanx:
Closing this, as of today seems to compile fine with gcc and the existing options.

PR #970, opened 2022-07-25
utility: Looping issue as length is usize

robryanx:
As i is unsigned i>=0 would always be true


robryanx:
Closing as I see that it has been solved here: https://github.com/SerenityOS/jakt/pull/942

PR #972, opened 2022-07-25
DRAFT: Traits MVP

cybergsus:
This is a draft PR to port over and extend the trait implementation. Used #378
with @kleinesfilmroellchen's work as a base, and I'd want to merge their commits
regarding documentation since there's no extra work for me to do there rather
than writing extra stuff or extra tests.

How it works (probably document this somewhere else?):
- Traits are registered almost at the top of the predecl chain, to ensure later passes know what traits exist.
- When the user tells that a type implements X trait, first assume it is correct, i.e register that trait as implemented, and let the rest of the typechecking run with that assumption.
- If an implementation for a given trait and a given type is incomplete/methods implementations aren't correct, errors will be reported and the code won't compile.

This way of deferring the typechecking makes sure we don't output more errors than strictly necessary:

When the user states that type `Foo` implements trait `Bar`, this means they want to use it in a generic function/struct that requires trait `Bar`.

If we prevented a trait implementation from being registered when such implementation didn't typecheck, we'll just confuse the user by telling them that `'Bar' trait is not implemented` (they'll think "But I *just* told you that it's implemented!").

By registering the implementation and later checking it without removing it from the assumptions our error report is much more clear: `Hey, I know you wanted to implement 'Bar' trait, but I can't compile because your implementation is wrong`.

---

Tasks:
- [x] Parser:
    - [x] Parse trait definitions
    - [x] Parse some way to know what traits are implemented for a record type (no duck typing!)
    - [x] Parse trait requirements
- [ ] Typechecker:
    - [x] Instantiate traits as types
    - [x] Register trait implementations for deferred checking
        - [x] Records
    - [x] Check that all trait methods are implemented for a struct that mentions that implements that trait
        - [x] Check that those methods' signatures correspond with the traits that each method implements
        - [x] Check that there are no missing methods to be implemented
        - [ ] Check that there are no extra methods than what the trait requires?
    - [ ] Check generic types' trait requirements upon instantiation
        - [x] Function
        - [ ] Struct/Class
        - [ ] Enum 
    - [x] Create tests for each of the errors possible from traits:
        - [x] Method not found
        - [x] Invalid signature
    - [ ] Codegen:
        - [ ] Make classes work (`Jakt::NonNullRefPtr`)
- [ ] Meta:
    - [ ] Merge documentation from original PR
    - [ ] Make it work with the REPL (?)

### DIFF
--- a/samples/traits/basic.jakt
+++ b/samples/traits/basic.jakt
@@ -1,5 +1,5 @@
 /// Expect:
-/// - output: "100\n42\n69\n144\n"
+/// - output: "100\n69\n"
 
 // trait definition
 trait Hashable {
@@ -10,11 +10,6 @@ trait Bar {
     function bar(mut this)
 }
 
-
-// implementation on a class
-class Foo implements(Hashable) {
-    public function hash(this) => 42u64
-}
 
 // implementation on a struct
 struct S implements(Hashable, Bar) {
@@ -49,6 +44,5 @@ function print_hash<T requires(Hashable)>(anon value: T) {
 
 function main() {
     print_hash(S())
-    print_hash(Foo())
     print_hash(U64Hash(value: 69u64))
 }
--- a/samples/traits/basic.jakt
+++ b/samples/traits/basic.jakt
@@ -35,20 +35,7 @@ enum Baz implements(Hashable) {
 }
 
 // implementing traits for builtin types can be done
-// through a newtype pattern. Also, requirements also apply for
-// generic classes, structs and enums!
-struct SumHash<T requires(Hashable)> implements(Hashable) {
-    values: [T]
-
-    function hash(this) -> u64 {
-        mut value: u64 = 0
-        for item in .values.iterator() {
-            value += item.hash()
-        }
-        return value
-    }
-}
-
+// through a newtype pattern.
 struct U64Hash implements(Hashable) {
     value: u64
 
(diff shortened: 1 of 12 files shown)


cybergsus:
[6088957](https://github.com/SerenityOS/jakt/pull/972/commits/6088957865d0a8d963312c6171f2cdf5014f9bb7) and other commits related to tests will be reordered when submitting the PR so all commits pass CI.

cybergsus:
I've seen a couple of commits mix stuff that should have been added later, I'll have to review the patches on all commits and make sure every part of the feature implementation is added incrementally, with all test passing.

cybergsus:
Note: squashed everything into a single commit so that merging isn't painful. I'll distribute the patches once the work is finished.

maddanio:
just curious: would it be thinkable not to have to state `implements` in the implementing classes? Then a trait requirement would always be fulfilled if the given method(field) exists, thus allowing structural matching.

cybergsus:
Leaving trait implementation to be implicitly accepted by just having a method backfires in go, where you implement traits without no one knowing what it's implementing. And doing it specifically makes it easier for the readers know what functionality you're implementing, besides being able to have two traits with different meanings and have methods whose signatures match.

cybergsus:
A method check as a requirement might be doable in the future, but it will be explicit in the form of a comptime function (where you can go read its definition), or more explictly, directly inside the requires clause.

maddanio:
There are/will be requires clauses? Ok, i always thought traits are an alternative

alimpfard:
Tracking continuation in #1242.

PR #975, opened 2022-07-26
selfhost: Implement match on Optional Enums

mattisboeckle:
Unwraps the Optional into an enum and handles the no value case with a default- or None-case.
Enables things like this:
```
enum A {
    B
    C
}

function main() {
    let a: A? = None
    let b: A? = A::B

    let res_a: String = match a {
        B => "FAIL"
        else => "PASS"
    }

    let res_b: String = match b {
        None => "NONE"
        B => "PASS"
        else => "FAIL"
    }

    print("{}, {}", res_a, res_b)
}
```

### DIFF
--- a/samples/match/match_optional_enum.jakt
+++ b/samples/match/match_optional_enum.jakt
@@ -1,0 +1,30 @@
+/// Expect: selfhost-only
+/// - output: "PASS, PASS, PASS"
+
+enum A {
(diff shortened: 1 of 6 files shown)


maddanio:
hmm, using else for the None check removes the check for match completeness even if it is desired

PR #976, opened 2022-07-26
Closures but for real this time

lanmonster:
Andreas's video about references inspired me to start this again and take a more systematic approach. The result is a much better product in my opinion. Returning a function is explicitly not allowed right now as there are some things that need to be worked out.

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -1507,12 +1507,16 @@ struct CodeGenerator {
 
             yield output
         }
-        Function(params, block) => {
+        Function(params, can_throw, block, return_type_id) => {
             mut generated_params: [String] = []
             for param in params.iterator() {
                 generated_params.push(format("{} {}", .codegen_type(param.variable.type_id), param.variable.name))
             }
-            yield format("[&]({}){}", join(generated_params, separator: ", "), .codegen_block(block))
+            let return_type = match can_throw {
+                true => format("ErrorOr<{}>", .codegen_type(return_type_id))
+                else => .codegen_type(return_type_id)
+            }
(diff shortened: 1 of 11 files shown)


selfhost/codegen.jakt
```diff
@@ -1507,6 +1507,13 @@ struct CodeGenerator {
 
             yield output
         }
+        Function(params, block) => {
+            mut generated_params: [String] = []
+            for param in params.iterator() {
+                generated_params.push(format("{} {}", .codegen_type(param.variable.type_id), param.variable.name))
+            }
+            yield format("[&]({}){}", join(generated_params, separator: ", "), .codegen_block(block))
```
awesomekling: So right now we're just spitting out `[&]` in C++ codegen to capture *everything* by reference.

Let's force the user to be explicit about what's being captured, instead of doing it implicitly. Also, Jakt needs to be aware of captures, so that they can be used inside the lambda. Otherwise there's no point to capturing in the first place :)

lanmonster: How should the syntax look? rip off cpp's syntax and pass it through to codegen or something else?

maddanio:
so the "function" word will still be required? Both js and c++ let you drop it, which makes things much cleaner sometimes imho

lanmonster:
> so the "function" word will still be required? Both js and c++ let you drop it, which makes things much cleaner sometimes imho

For now, yes. It makes parsing much easier.

maddanio:
> > so the "function" word will still be required? Both js and c++ let you drop it, which makes things much cleaner sometimes imho
> 
> 
> 
> For now, yes. It makes parsing much easier.

Ah, yes, that would indeed be quite complex and require lots of backtracking

PR #979, opened 2022-07-27
selfhost: Bunch of fixes to the IDE integration

awesomekling:
With these fixes, we can now show diagnostics, type hints, and try hints when using selfhost as the backend for the IDE extension.

There are still some little issues to figure out, but this is a huge step towards usability.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -1322,12 +1322,12 @@ struct Typechecker {
 
     function dump_type_hint(this, type_id: TypeId, span: Span) throws {
         println("{{\"type\":\"hint\",\"file_id\":{},\"position\":{},\"typename\":\"{}\"}}"
-                span.file_id, span.end, .type_name(type_id))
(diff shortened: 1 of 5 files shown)


selfhost/typechecker.jakt
```diff
@@ -1322,12 +1322,12 @@ struct Typechecker {
 
     function dump_type_hint(this, type_id: TypeId, span: Span) throws {
         println("{{\"type\":\"hint\",\"file_id\":{},\"position\":{},\"typename\":\"{}\"}}"
-                span.file_id, span.end, .type_name(type_id))
+                span.file_id.id, span.end, .type_name(type_id))
     }
 
-    function dump_try_hint(this, span: Span) throws {
+    function dump_try_hint( this, span: Span) throws {
```
Dexesttp, line 1328: Is that extra space intentional?

awesomekling: :catdogalert: No

PR #987, opened 2022-07-28
meta: Build selfhost compiler

lanmonster:
When updating the vscode extension, build the selfhost and remind to set the executable path in the extension's settings

### DIFF
--- a/meta/update-vscode-extension.sh
+++ b/meta/update-vscode-extension.sh
@@ -31,9 +31,9 @@ then
     code --uninstall-extension JT.jakt
 fi
 
-# build and install the new version of the compiler
-echo "Building and installing most recent version of jakt compiler"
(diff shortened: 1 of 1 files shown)


awesomekling:
We should make sure to build with optimizations here, since using the IDE extension with an unoptimized build of selfhost is obnoxious.

meta/update-vscode-extension.sh
```diff
@@ -46,7 +46,9 @@ code --install-extension "$NEW_VSIX_PATH"
 
 if code --list-extensions | grep -q jakt
 then
-    echo "Successfully updated"
+    echo "Successfully updated."
+    echo -e "\033[33m" # Yellow text color
```
alimpfard, line 50: please clear that afterwards

PR #991, opened 2022-07-28
jakttest: Remove dirty patch-includes script

cybergsus:
Now that we've moved to selfhost, we can use the import extern feature
that comes with it to make codegen generate the headers we need! No more
hacky python script!


jakttest/jakttest.jakt
```diff
@@ -10,48 +10,57 @@ import parser { Parser }
...
-    // Calls `stat` on a file. Silences `ENOENT` error by setting
-    // StatResults.exists to false.
-    extern function stat_silencing_enoent(path: String) throws -> StatResults?
 
+    }
+}
 
+// FIXME: introducing this inside a namespace seems to override the namespace
+// from import extern
+//namespace fs {
```
lanmonster, line 56: remove

jakttest/jakttest.jakt
```diff
@@ -90,33 +99,37 @@ namespace fs {
 
         return results
     }
-}
+//}
```
lanmonster, line 102: remove

jakttest/jakttest.jakt
cybergsus: why? It has a reason to be commented. But maybe just putting in the fixme that it should be inside a namespace `fs` is enough. Thanks

jakttest/jakttest.jakt
```diff
@@ -10,113 +10,125 @@ import parser { Parser }
...
+}
 
-    function mkdir_if_not_present(path: String) throws {
-        let stat = stat_silencing_enoent(path)
-        if not stat.has_value() or not stat!.is_directory() {
-            fs::mkdir(path)
-        }
+// FIXME: These three functions should be under a `namespace fs`, but
+// introducing this inside a namespace seems to override the namespace from
+// import extern (which shouldn't happen, it should rather add these)
```
ADKaster, line 56: do we have an issue for this? be good to reference it (either `#1234` or with a link to the issue)

cybergsus: no, but I can raise it.

ADKaster (approved):

PR #992, opened 2022-07-29
Meta: Build new bootstrap compiler with CMakeLists build flow

ADKaster:
- Remove corrosion as a dependency, and build all 3 stages of the compiler using cmake.
- Merge the two stages of selfhost CI into one with the power of `if:`
- Use the cmake build in CI

Still needs addressing: the cmake build and the top level ninja build can't co-exist in the same `./build` directory


documentation/cmake-bootstrap.md
```diff
@@ -1,11 +1,11 @@
 # Bootstapping with CMake
 
 The included CMakeLists.txt at the top level of the project provides cross-platform build rules for
-auto-bootstrapping the compiler all the way to stage 3. The build takes advantage of the [Corrosion](https://github.com/corrosion-rs/corrosion) library for building the Rust-based compiler and exposing it as an imported target to CMake.
+auto-bootstrapping the compiler all the way to stage 3.
 
 For reference, the stages:
 
-Stage 1: Rust-based compiler. This compiler, written in Rust, is able to parse Jakt source code and output C++ code.
+Stage 1: The bootstrap compiler. This compiler, written in autogenerated C++ is able to parse Jakt source code and output C++ code.
```
awesomekling, line 8: This is actually `bootstrap/stage0` so maybe it should be called "Stage 0" here as well :)

ADKaster: Addressed in #1000

PR #994, opened 2022-07-29
Everywhere: Implement explicit struct/class constructor

maddanio:
This makes a user-defined static create function act as an
explicit constructor that is called instead of the implicit one.
The explicit contructor is then the only function which can see
the implicit constructor and uses it to actually create the instance.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -203,7 +203,7 @@ boxed enum Type {
                     }
                 }
                 GenericInstance(id: lhs_id, args: lhs_args) => {
-                    guard rhs is GenericInstance(id: rhs_id, args: rhs_args) 
+                    guard rhs is GenericInstance(id: rhs_id, args: rhs_args)
                         and lhs_id.equals(rhs_id)
                         and lhs_args.size() == rhs_args.size() else {
                         return false
@@ -436,6 +436,10 @@ class CheckedFunction {
     public parsed_function: ParsedFunction?
 
     public function is_static(this) -> bool {
+        if .type is ExplicitConstructor {
+            return true
+        }
+
         if .params.size() < 1 {
             return true
         }
@@ -2956,65 +2960,67 @@ struct Typechecker {
         let module_id = .current_module_id
         .current_struct_type_id = struct_type_id
 
-        let constructor_id = .find_function_in_scope(parent_scope_id: checked_struct_scope_id, function_name: record.name)
-        if constructor_id.has_value() {
-            if record.record_type is Class and record.definition_linkage is External {
-                // XXX: The parser always sets the linkage type of an extern class'
-                //      constructor to External, but we actually want to call the
-                //      class' ::create function, just like we do with a
-                //      ImplicitConstructor class.
-                mut func = .get_function(constructor_id!)
-                func.linkage = FunctionLinkage::External
+        match record.definition_linkage {
+            External => {
+                let constructor_id = .find_function_in_scope(parent_scope_id: checked_struct_scope_id, function_name: record.name)
+                if constructor_id.has_value() {
+                    mut func = .get_function(constructor_id!)
+                    // XXX: The parser always sets the linkage type of an extern class'
+                    //      constructor to External, but we actually want to call the
+                    //      class' ::create function, just like we do with a
+                    //      ImplicitConstructor class.
+                    func.linkage = FunctionLinkage::External
+                }
             }
-        } else if not record.definition_linkage is External {
-            // No constructor found, so let's make one
+            Internal => {
 
-            let constructor_can_throw = record.record_type is Class
-            let function_scope_id = .create_scope(parent_scope_id, can_throw: constructor_can_throw, debug_name: format("generated-constructor({})", record.name))
-            let block_scope_id = .create_scope(parent_scope_id: function_scope_id, can_throw: constructor_can_throw, debug_name: format("generated-constructor-block({})", record.name))
+                let constructor_can_throw = record.record_type is Class
+                let function_scope_id = .create_scope(parent_scope_id, can_throw: constructor_can_throw, debug_name: format("generated-constructor({})", record.name))
+                let block_scope_id = .create_scope(parent_scope_id: function_scope_id, can_throw: constructor_can_throw, debug_name: format("generated-constructor-block({})", record.name))
 
-            let checked_constructor = CheckedFunction(
-                name: record.name
-                name_span: record.name_span
-                visibility: Visibility::Public
-                return_type_id: struct_type_id
-                return_type_span: None
-                params: []
-                generic_params: []
-                block: CheckedBlock(
-                    statements: []
-                    scope_id: block_scope_id
-                    control_flow: BlockControlFlow::MayReturn
-                    yielded_type: TypeId::none()
+                let checked_constructor = CheckedFunction(
+                    name: record.name
+                    name_span: record.name_span
+                    visibility: Visibility::Public
+                    return_type_id: struct_type_id
+                    return_type_span: None
+                    params: []
+                    generic_params: []
+                    block: CheckedBlock(
+                        statements: []
+                        scope_id: block_scope_id
+                        control_flow: BlockControlFlow::MayReturn
+                        yielded_type: TypeId::none()
+                    )
+                    can_throw: constructor_can_throw
+                    type: FunctionType::ImplicitConstructor
+                    linkage: FunctionLinkage::Internal
+                    function_scope_id
+                    is_instantiated: true
+                    parsed_function: None
                 )
-                can_throw: constructor_can_throw
-                type: FunctionType::ImplicitConstructor
-                linkage: FunctionLinkage::Internal
-                function_scope_id
-                is_instantiated: true
-                parsed_function: None
-            )
 
-            // Internal constructor
-            mut module = .current_module()
-            module.functions.push(checked_constructor)
+                // Internal constructor
+                mut module = .current_module()
+                module.functions.push(checked_constructor)
 
-            mut func = module.functions.last()!
-            for field_id in .get_struct(struct_id).fields.iterator() {
-                let field = .get_variable(field_id)
-                func.params.push(CheckedParameter(
-                    requires_label: true
-                    variable: field
-                ))
+                mut func = module.functions.last()!
+                for field_id in .get_struct(struct_id).fields.iterator() {
+                    let field = .get_variable(field_id)
+                    func.params.push(CheckedParameter(
+                        requires_label: true
+                        variable: field
+                    ))
+                }
+
+                // Add constructor to the struct's scope
+                .add_function_to_scope(
+                    parent_scope_id: checked_struct_scope_id
+                    name: record.name
+                    function_id: FunctionId(module: module_id, id: .current_module().functions.size() - 1)
+                    span: record.name_span
+                )
             }
-
-            // Add constructor to the struct's scope
-            .add_function_to_scope(
-                parent_scope_id: checked_struct_scope_id
-                name: record.name
-                function_id: FunctionId(module: module_id, id: .current_module().functions.size() - 1)
-                span: record.name_span
-            )
         }
 
         for method in record.methods.iterator() {
@@ -3028,13 +3034,14 @@ struct Typechecker {
         mut parent_generic_parameters: [TypeId] = []
         mut scope_id = .prelude_scope_id()
         mut definition_linkage = DefinitionLinkage::Internal
-
+        mut function_return_type_id = unknown_type_id()
         match parent_id {
             Struct(struct_id) => {
                 mut structure = .get_struct(struct_id)
                 parent_generic_parameters = structure.generic_parameters
                 scope_id = structure.scope_id
                 definition_linkage = structure.definition_linkage
+                function_return_type_id = structure.type_id
             }
             Enum(enum_id) => {
                 let enum_ = .get_enum(enum_id)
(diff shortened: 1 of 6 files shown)


selfhost/typechecker.jakt
```diff
@@ -6194,26 +6230,41 @@ struct Typechecker {
...
-            let maybe_function_id = .find_function_in_scope(parent_scope_id: structure.scope_id, function_name: call.name)
-            if maybe_function_id.has_value() {
-                return maybe_function_id!
+            let maybe_explicit_constructor_id = .find_function_in_scope(parent_scope_id: structure.scope_id, function_name: "create")
+            let maybe_implicit_constructor_id = .find_function_in_scope(parent_scope_id: structure.scope_id, function_name: call.name)
+            if maybe_explicit_constructor_id.has_value() {
+                let explicit_constructor = .get_function(maybe_explicit_constructor_id!)
+                if explicit_constructor.type is ExplicitConstructor {
+                    if .scope_can_access(accessor: scope_id, accessee: explicit_constructor.block.scope_id) {
+                        // fixme: can't we assert this?
```
awesomekling, line 6255: `maybe_implicit_constructor_id!` has a built-in `VERIFY()`, so can we just drop the `has_value()` check to begin with?

maddanio: I am not sure I understand. The `!` always come after a has value check afaict. there is the question wether we can assert the existence of an explicit constructor, but I am not 100% sure.

maddanio: I.e. in theory an implicit constructor should always exist, but I don't quite understand how external constructors work, so I didnt dare yet :).

AtkinsSJ:
This has conflicts. Please fix them. :^)

maddanio:
> This has conflicts. Please fix them. :^)

ok, they must be from today, will have a look

maddanio:
> This has conflicts. Please fix them. :^)

merged and tests passed

awesomekling:
First off, please add tests when introducing a new feature like this. :^)

So, I like the general idea of custom constructors, but the specifics here feel a bit clunky and "magical" with the separation of explicit and implicit constructors.

Here's an alternative proposal:

```
struct Foo {
    field: i64

    ctor {
        // This is the default constructor -> Foo::Foo() in C++
        // It is invoked by doing:
        //     let foo = Foo()
        // You cannot throw in this context.
        field = 0
    }

    ctor(a: i64) {
        // This is an overload constructor -> Foo::Foo(i64 a) in C++
        // It is invoked by doing:
        //     let foo = Foo(a: 123)
        // You cannot throw in this context.
        field = a
    }

    dtor {
        // This is a destructor -> Foo::~Foo() in C++
        // It is invoked when a Foo goes out of scope.
        // You cannot throw in this context.
    }

    function create() -> Foo {
        // This is a factory function that returns a new Foo
        // There's nothing special about it.
        // You can throw in this context.
        return Foo()
    }

    function create(a: i64) -> Foo {
        // This is an overloaded factory function that returns a new Foo
        // There's nothing special about it.
        // You can throw in this context.
        return Foo(a)
    }
}
```


And of course, `struct` constructors are `public` by default, and `class` ones are always `private`.

Basically, modeling constructors and destructors much closer to how they end up in C++. This will make it easier to interface with our existing C++ libraries as well.

maddanio:
> First off, please add tests when introducing a new feature like this. :^)

Ah crap, somehow the three test files I had added got lost. Just added them back and pushed. Sorry about that.

> So, I like the general idea of custom constructors, but the specifics here feel a bit clunky and "magical" with the separation of explicit and implicit constructors.
> 
> Here's an alternative proposal:
> 
> ```
> struct Foo {
>     field: i64
> 
>     ctor {
>         // This is the default constructor -> Foo::Foo() in C++
>         // It is invoked by doing:
>         //     let foo = Foo()
>         // You cannot throw in this context.
>         field = 0
>     }
> 
>     ctor(a: i64) {
>         // This is an overload constructor -> Foo::Foo(i64 a) in C++
>         // It is invoked by doing:
>         //     let foo = Foo(a: 123)
>         // You cannot throw in this context.
>         field = a
>     }
> 
>     dtor {
>         // This is a destructor -> Foo::~Foo() in C++
>         // It is invoked when a Foo goes out of scope.
>         // You cannot throw in this context.
>     }
> 
>     function create() -> Foo {
>         // This is a factory function that returns a new Foo
>         // There's nothing special about it.
>         // You can throw in this context.
>         return Foo()
>     }
> 
>     function create(a: i64) -> Foo {
>         // This is an overloaded factory function that returns a new Foo
>         // There's nothing special about it.
>         // You can throw in this context.
>         return Foo(a)
>     }
> }
> ```
> 
> And of course, `struct` constructors are `public` by default, and `class` ones are always `private`.

Ok sure. Then there was a misunderstanding, I had heard from someone else that using `create` and `destruct` had been agreed. So basically you want `ctor` instead and also allow dropping the function keyword and return type and even there brackets (are you sure about the latter one? looks weird to me...). that will require a little parser magic, but not too much.

awesomekling:
> > First off, please add tests when introducing a new feature like this. :^)
> 
> Ah crap, somehow the three test files I had added got lost. Just added them back and pushed. Sorry about that.
> 
> > So, I like the general idea of custom constructors, but the specifics here feel a bit clunky and "magical" with the separation of explicit and implicit constructors.
> > Here's an alternative proposal:
> > ```
> > struct Foo {
> >     field: i64
> > 
> >     ctor {
> >         // This is the default constructor -> Foo::Foo() in C++
> >         // It is invoked by doing:
> >         //     let foo = Foo()
> >         // You cannot throw in this context.
> >         field = 0
> >     }
> > 
> >     ctor(a: i64) {
> >         // This is an overload constructor -> Foo::Foo(i64 a) in C++
> >         // It is invoked by doing:
> >         //     let foo = Foo(a: 123)
> >         // You cannot throw in this context.
> >         field = a
> >     }
> > 
> >     dtor {
> >         // This is a destructor -> Foo::~Foo() in C++
> >         // It is invoked when a Foo goes out of scope.
> >         // You cannot throw in this context.
> >     }
> > 
> >     function create() -> Foo {
> >         // This is a factory function that returns a new Foo
> >         // There's nothing special about it.
> >         // You can throw in this context.
> >         return Foo()
> >     }
> > 
> >     function create(a: i64) -> Foo {
> >         // This is an overloaded factory function that returns a new Foo
> >         // There's nothing special about it.
> >         // You can throw in this context.
> >         return Foo(a)
> >     }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And of course, `struct` constructors are `public` by default, and `class` ones are always `private`.
> 
> Ok sure. Then there was a misunderstanding, I had heard from someone else that using `create` and `destruct` had been agreed. So basically you want `ctor` instead and also allow dropping the function keyword and return type and even there brackets (are you sure about the latter one? looks weird to me...). that will require a little parser magic, but not too much.

I'm not married to the specific words `ctor` and `dtor`, but we need to call them *something*. We can always change them later. :)

But yes, dropping the `function` keyword makes it clear that these are different from ordinary member functions.

maddanio:
> > > First off, please add tests when introducing a new feature like this. :^)
> > 
> > 
> > Ah crap, somehow the three test files I had added got lost. Just added them back and pushed. Sorry about that.
> > > So, I like the general idea of custom constructors, but the specifics here feel a bit clunky and "magical" with the separation of explicit and implicit constructors.
> > > Here's an alternative proposal:
> > > ```
> > > struct Foo {
> > >     field: i64
> > > 
> > >     ctor {
> > >         // This is the default constructor -> Foo::Foo() in C++
> > >         // It is invoked by doing:
> > >         //     let foo = Foo()
> > >         // You cannot throw in this context.
> > >         field = 0
> > >     }
> > > 
> > >     ctor(a: i64) {
> > >         // This is an overload constructor -> Foo::Foo(i64 a) in C++
> > >         // It is invoked by doing:
> > >         //     let foo = Foo(a: 123)
> > >         // You cannot throw in this context.
> > >         field = a
> > >     }
> > > 
> > >     dtor {
> > >         // This is a destructor -> Foo::~Foo() in C++
> > >         // It is invoked when a Foo goes out of scope.
> > >         // You cannot throw in this context.
> > >     }
> > > 
> > >     function create() -> Foo {
> > >         // This is a factory function that returns a new Foo
> > >         // There's nothing special about it.
> > >         // You can throw in this context.
> > >         return Foo()
> > >     }
> > > 
> > >     function create(a: i64) -> Foo {
> > >         // This is an overloaded factory function that returns a new Foo
> > >         // There's nothing special about it.
> > >         // You can throw in this context.
> > >         return Foo(a)
> > >     }
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > And of course, `struct` constructors are `public` by default, and `class` ones are always `private`.
> > 
> > 
> > Ok sure. Then there was a misunderstanding, I had heard from someone else that using `create` and `destruct` had been agreed. So basically you want `ctor` instead and also allow dropping the function keyword and return type and even there brackets (are you sure about the latter one? looks weird to me...). that will require a little parser magic, but not too much.
> 
> I'm not married to the specific words `ctor` and `dtor`, but we need to call them _something_. We can always change them later. :)
> 
> But yes, dropping the `function` keyword makes it clear that these are different from ordinary member functions.

ok, implemented now. basically now it is named `ctor` and you have to omit the `function` keyword before it and the return type and are allowed to omit the parentheses entirely if you don't have parameters. last thing I am wondering is should I make the class ctor `throws` implicitly and disallow specifying it and keep it optional for `struct`?

maddanio:
forgot to test constructor args and of course failed. fixing

maddanio:
> forgot to test constructor args and of course failed. fixing

actually works. I returned the wrong type and return type checking of functions seems fragile, but not going to fix this here

awesomekling:
> last thing I am wondering is should I make the class ctor throws implicitly and disallow specifying it and keep it optional for struct?

We cannot throw in our C++ constructors, so let's not support it in Jakt constructors either. At least not initially. We typically use a factory function pattern where something like a `ErrorOr<T> Foo::create()` does all the fallible things, and if successful, returns a newly-constructed `Foo`.

maddanio:
Hmm, but if you really want raii then you need to handle the case where the resource acquisition fails. Or would you the always require a factory function for resource like classes like connections or files or the like?

maddanio:
So instead of having Socket(adress) you would have Socket::create(address) and male the ctor private (after all what good is an unopened socket)? Seems unnecessarily verbose to me. What i will do is "hide" the implicit throws from the pointer allocation, but i still think constructors should be able to throw.

maddanio:
Maybe restrict throw to classes? Feels to me like the right balance?

maddanio:
Also disallowing throws on class ctors will make the implementation quite a bit more tricky because in reality it will still throw because in reality it will have to use new. Sure doable just saying :)

maddanio:
Unless of course class allocation failure were a panic onstead of a throws :). Anywho, after all I will accept the opinination either way, just trying to make my case :)

maddanio:
Ok, after pondering a minute i think i can coerce typechecker and codegen into doing the right thing without too mich complications. Might take a while though because "real life"

maddanio:
@awesomekling I worked in your suggestions: no `throws` allowed for `ctor`. the last issue of this PR may be that basically out of implementation lazyness explicit struct ctors will introduce a factory function and also not make the old "implicit" ctor private in the generated c++ code. Calling the factory function and implicit ctor is prevented in jakt via typechecker, but wouldnt be in c++. Would it be important to codegen the explicit struct ctor as a plain c++ constructor to make calling it from c++ easier? I couldn't find any docs regarding that, but it seemed on discord there was some reservations about "deterioating" that interface.

ADKaster:
Re: factory functions for structs

My question is, Hlhow important is it to keep the use case of "C++ application or library calling into Jakt library" looking comfortable from the C++ side? 

We'll have to maintain both side of that interface, C++ calling Jakt and Jakt calling C++ (and maybe multiple hops across the boundary in one call stack) for quite some time during any experimentation or transition period for the Serenity codebase. It would be unfortunate if it was difficult or extra verbose to create jakt types in C++ code, but perhaps that's thinking too far ahead?

maddanio:
Also for explicit class ctors the factory function will be codegened as `ctor` and the generated implicit constructor still be visible as `create`. Is it a necessity to codegen the explicit ctor as `create` abd make the implicit ctor private in codegen?

maddanio:
Well, if there is desire to keep the c++ appearance of jakt types `clean` I can go that extra mile. Might be a while though until I get around to it.

maddanio:
Ah, I just figured it will be quite hard to not codegen the explicit struct constructor as a factory function since c++ does not actually allow code in the constructor before the field initializers which the ctors from this pr do (and which i find very convenient). Likely to codegen struct ctors as c++ constructors will require ctors to not have a block but only contain the implicit ctor call

maddanio:
But i could still make sure the factory functions are called create and the implicit constructors private in the generated c++ if desired

jntrnr:
Thinking through this a bit - there are a few ways jakt can go with constructors.

There's a kind of more traditional C++ approach: you get a default constructor, and you can replace it with a custom one. There is some special treatment for constructions and constructors, for example getting a new keyword just to construct and whatnot.

There's also the Rust approach, which is to not treat constructors and construction any different from any other method. You just make a static method that does the type of construction you want and call it. For example `Foo::new(...)`

I've tried both, and there are advantages/disadvantages to each. Having said that, I think the Rust approach is the simpler of the two. You just make a new static method to build the value how you want. Now that we have default arg values, this can get even cleaner.

As we already support the Rust style today, we could probably stick with that, unless you see advantages it doesn't have.

maddanio:
The one thing is you also want to disallow the default constructor in many cases, as it gives you access to private members, which you want to avoud. This is exactly what the ctor does: once it exists you no longer get the default constructor

maddanio:
The alternative is to disallow the default constructor whenever there are private fields, but that would break lots of existing code

maddanio:
Hmm, so that's it? No explicit ctor and just keave the default ctor with all the private fields spilling out?

maddanio:
closing this myself now, as there is no concensus on this, instead we will likely simply hide the implicit ctor if private fields are present

PR #1000, opened 2022-07-30
meta+ci: Renumber compiler stages to {0..2} instead of {1..3}

ADKaster:
Add a CMake cache variable FINAL_STAGE to control whether we're stopping
at stage 1, or continuing to stage 2 for that extra validation step.

Also add CI stages to independently test stage1 and stage2.

Add Windows CI while we're at it. Only build selfhost stage 1/2 with clang-cl.exe, but don't run any tests.

Two extra commits for some cleanup.


.github/workflows/ci.yml
```diff
@@ -49,12 +46,23 @@ jobs:
...
+
+      - name: Build Jakt Stage 2 Selfhost
         run: |
-          cmake -GNinja -B build -DCMAKE_CXX_COMPILER=${{ matrix.cxx_compiler }} -DCMAKE_BUILD_TYPE=Release
+          # Modify CMake cache to set final stage to stage 2
+          cmake -B build -DFINAL_STAGE=2
           cmake --build build
 
-      - name: Test Jakt selfhost
+      - name: Test Jakt Stage 2 selfhost
```
ADKaster, line 65: Is it worthwhile to test both the stage1 and stage2 compilers? or should be pick one to run the tests, and just make sure to build the stage2 compiler to make sure nothing breaks/if a new bootstrap is needed?

PR #1004, opened 2022-07-30
parser: Allow newlines in certain situations

0GreenClover0:
Make something like:

```
class Test
{
}
--
enum Test
{
}
--
struct Test
{
}
```

as well as:

```
if something == something_else
{
}
else
{
}
```

valid. Not sure if it is a design decision to not allow that 🤔


awesomekling:
Sure, let's be lenient.

PR #988, opened 2022-07-28
fix: function parameter and argument separator parsing

feniljain:
- Fix parsing of parameters and argument separator parsing by restricting usage to comma and new line
- Should close https://github.com/SerenityOS/jakt/issues/966
- Would like to hear points on:
https://github.com/feniljain/jakt/blob/f3bc872739fb3a165bd5c30ae0c8d062e6e9af1b/selfhost/parser.jakt#L1465-L1472

I have tried to add appropriate testcases, if you feel something's wrong, do point out!

PR #1006, opened 2022-07-30
parser: Allow function parameters/arguments to be separated with spaces

mattco98:
See https://github.com/SerenityOS/jakt/pull/988#issuecomment-1199936147.

### DIFF
--- a/selfhost/parser.jakt
+++ b/selfhost/parser.jakt
@@ -1450,26 +1450,27 @@ struct Parser {
 
         // Have we already found an error?
         mut error = false
-        mut parameter_complete = false
+        mut parsing_parameter = false
 
         while not .eof() {
             match .current() {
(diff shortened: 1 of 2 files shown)


AtkinsSJ:
I don't know about anyone else, but this makes me really uncomfortable. 😬

jntrnr:
@AtkinsSJ - the intent here is that you will run Jakt with a formatter that automatically formats your code. Being a bit loose with syntax like allowing you to use a space, allows the formatter to just put it in for you without you having to do it.

AtkinsSJ:
@jntrnr I do get your point. Personally though, I prefer the clear separation of a comma/newline over the convenience of not having to type one character. (I'd also have semicolons in the language if it was up to me though. 😅)

PR #1011, opened 2022-07-31
jakttest: Add Windows OSAL pieces for filesystem and process management

ADKaster:
Add a CMake build rule for jakttest as well.

The file traversal and discovery code works, but the process spawning
code will remain "looks correct according to the docs" until there's
an alternative to run-one.sh. We can't run a bash script in powershell
with ./ the way Unix OS's do it.

Behold: it doesn't work

```
PS C:\Users\Andrew\Source\Repos\jakt> ./out/build/x64-Debug/jakttest.exe --assume-updated-selfhost --assume-updated
[ SKIP ] samples/modules/a.jakt
[ SKIP ] samples/modules/b.jakt
[ SKIP ] samples/imports/foo.jakt
[ SKIP ] samples/imports/assign_across_import/helper.jakt
[ SKIP ] selfhost/codegen.jakt
[ SKIP ] selfhost/compiler.jakt
[ SKIP ] selfhost/error.jakt
[ SKIP ] selfhost/ide.jakt
[ SKIP ] selfhost/lexer.jakt
[ SKIP ] selfhost/parser.jakt
[ SKIP ] selfhost/prelude.jakt
[ SKIP ] selfhost/typechecker.jakt
[ SKIP ] selfhost/utility.jakt
Runtime error: Error(code=193)
```

But! It does parse all the files under selfhost, samples, and tests, and it does create a temp directory under 
` C:\Users\Andrew\AppData\Local\Temp/jakttest/jakttest-tmp-11` on my machine.

It's likely that a suitable clang command line and powershell script to keep parity with run-one.sh would be tricky to develop.
If we could defer those steps to some C++ or build system code (CMake? CTest?) instead, that would be easier to keep cross-platform.


jntrnr:
@ADKaster - for this PR, are you aiming to have jakttest running and passing on Windows, or are you thinking of doing it in multiple steps?

ADKaster:
> are you aiming to have jakttest running and passing on Windows, or are you thinking of doing it in multiple steps?

Multiple steps.

 The current blocker on executing to completion is the `run-one.sh` shell script. I believe gsus is working on refactoring the "test driver" part of the application, which may or may not include that script directly.

jakttest/os.cpp
```diff
@@ -22,12 +31,39 @@ ErrorOr<size_t> get_num_cpus()
...
+    if (sigaction(SIGCHLD, &action, NULL) == -1)
+        return Error::from_errno(errno);
+    return {};
+}
+#else
+ErrorOr<size_t> get_num_cpus()
+{
+    SYSTEM_INFO sys_info = {};
+    GetSystemInfo(&sys_info);
+    return __builtin_popcountl(sys_info.dwActiveProcessorMask);
```
MacDue, line 54: Can this just be:
```suggestion
    SYSTEM_INFO sys_info = {};
    GetSystemInfo(&sys_info);
    return sys_info.dwNumberOfProcessors;
```
(This seems to be the implementation used in [Boost.Thread](https://github.com/boostorg/thread/blob/6fca28bfed46ee89822d4fffc0f6e7386e98c7b1/src/win32/thread.cpp#L508))

MacDue: The wording of the description on the Microsoft docs is a little confusing, but think the "To retrieve this value, use the [GetLogicalProcessorInformation](https://docs.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getlogicalprocessorinformation) function." bit is referring to the processor group.

ADKaster: ah, looks like I definitely misread the docs here then, I trust boost on this one.

jntrnr:
Giving this a try - I got this error. Is it expected?

```
[5/11] Linking CXX executable jakt_stage1.exe
FAILED: jakt_stage1.exe
cmd.exe /C "cd . && C:\PROGRA~2\MICROS~2\2022\BUILDT~1\VC\Tools\Llvm\x64\bin\CLANG_~1.EXE -fuse-ld=lld-link -nostartfiles -nostdlib -O3 -DNDEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrt -Xlinker /subsystem:console CMakeFiles/jakt_stage1.dir/jakt_stage1_main.cpp.obj -o jakt_stage1.exe -Xlinker /MANIFEST:EMBED -Xlinker /implib:jakt_stage1.lib -Xlinker /pdb:jakt_stage1.pdb -Xlinker /version:0.0   -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 -loldnames  && cmd.exe /C "cd /D C:\Source\jakt\build && "C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" -E create_symlink jakt_stage1.exe C:/Source/jakt/build/jakt""
CMake Error: failed to create symbolic link 'C:/Source/jakt/build/jakt': A required privilege is not held by the client.
```

ADKaster:
> CMake Error: failed to create symbolic link 'C:/Source/jakt/build/jakt': A required privilege is not held by the client.

Hmm. is your user bychance not a member of the `Administrator` Group?

https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/create-symbolic-links

> By default, members of the Administrators group have this right.

I can probably add a CMake flag to disable the handy dandy symlink, or convert it into a copy instead

jntrnr:
@ADKaster - possibly, though why would we require that? Seems like we'd want people to be able to contribute in normal user mode.

ADKaster:
>  possibly, though why would we require that? Seems like we'd want people to be able to contribute in normal user mode.

@jntrnr For sure, i've made hardlinks the default for windows, but folks who do commit code on a user with admin privs can set the NO_SYMLINKS flag to OFF in their CMakeCache to get the same behavior as unix-y envs.

jntrnr:
I'm still seeing a failure. Am I doing something wrong?

```
CMake Error: failed to create symbolic link 'C:/Source/jakt/build/jakt': A required privilege is not held by the client.
```

ADKaster:
@jntrnr I created a new local account and cloned my branch, and it set the settings per the screenshot from VS 2022 Preview below. I think what happened is you either didn't clear the cmake cache/build directory and build fresh, or you didn't manually set the  NO_SYMLINKS cmake cache variable (which is now the default, on Windows).

Jakt-settings

jntrnr:
@ADKaster can you try it from cmd instead of powershell?

jntrnr:
Bummer, I just tried to delete the contents of build and re-run cmake and still get the same error.

PR #1012, opened 2022-07-31
meta: Use CMake to build Jakttest

cybergsus:
Removes the "self-updating" feature from Jakttest since it isn't really needed
and updating the binary is as simple as running ninja, and moves the rules for
compiling it to the CMake build.


cybergsus:
CI will fail due to not being updated to building it correctly. Though now I see that CI wants to test every build stage of the compiler, so I might have to add more commands to CMake.

cybergsus:
#1011 will override it, since its solution is much simpler and readable. I'm keeping it here because it has a couple commits regarding Jakttest that might go into #1011 as well, so that it doesn't use the ninja file anymore.

ADKaster:
The relevant changes have been pulled into #1011, namely the removal of the "rebuild self" functionality and the build.ninja file for jakttest

cybergsus:
Great! Closing this then.

PR #1013, opened 2022-07-31
jakttest+ci: Turn off progress status line in CI

cybergsus:
CI output gets pretty messy since it doesn't interpret carriage
returns.


ADKaster:
I don't like removing the progress bar from CI entirely

It's invaluable when debugging CI flakes to know exactly which test was currently executing when the test process or some test hangs in the future.

cybergsus:
Huh, since the progress indicator tells the last test that was launched, maybe having all instances of the progress line in CI isn't so bad. Thanks for explaining.

PR #1014, opened 2022-07-31
Selfhost: Add shorthand '-p' for prettify-cpp-source

SmellyJay:
Adds back the '-p' argument the rust compiler had, as a shorter version of '--prettify-cpp-source'.


lanmonster:
Don't forget to update the help message

PR #1017, opened 2022-08-01
typechecker: argument offset fixes for methods

robryanx:


awesomekling:
I'm not sure we want to support calling static member functions on instances. In fact, some C++ static analyzers will warn about this (`clang-tidy` and some other one IIRC).

robryanx:
> I'm not sure we want to support calling static member functions on instances. In fact, some C++ static analyzers will warn about this (`clang-tidy` and some other one IIRC).

Yes, the purpose of this is just to fix the error. Previously when attempting to call a static method on an instance there would be an array out of bounds issues because the code was assuming that the `this` parameter would be present.

awesomekling:
Hm, all right. So can we expand this to also disallow calling static member functions on an instance? :)

robryanx:
> Hm, all right. So can we expand this to also disallow calling static member functions on an instance? :)

Not sure if I am missing something but I think that is already implemented.
The problem I am solving here is that after hitting the existing error `Cannot call static method on an instance of an object` the typechecker would continue on and parse the arguments list, where it would end up with a C++ error `Assertion `index < m_size' failed.` for out of bounds because `arg_offset` was set assuming that `this` would be a param because it was being called on an instance.

If I am missing a case can you give me an example call that doesn't error currently and I can add it.

awesomekling:
Oh duh, I'm silly. I didn't realize the tests were expected to fail 😅

Issue #1025, opened 2022-08-02
Code generation does not add template parameters for struct name when calling static methods


cybergsus:
Let's say I have the following Jakt code:
```jakt
import extern "my_generic.h" {
   extern struct Generic<T> {
      function create(secret: T) throws -> Generic<T>
      function get_secret(this) -> T
   }
}

function main() {
    let generic = Generic::create(secret: 42i32)
    println("The answer to everything is: {}", generic.get_secret())
}
```

It will typecheck and generate: 
```cpp
#include <lib.h>
#include <my_generic.h>
namespace Jakt {
ErrorOr<int> main(Array<String>) {
{
// Generic::create doesn't have a template arguments
const Generic<i32> generic = TRY((Generic::create(static_cast<i32>(42))));
outln(String("The answer to everything is: {}"),((generic).get_secret()));
}
return 0;
}

} // namespace Jakt
```
Here, the typechecking is correct. It inferred correctly that I have created a `Generic<i32>`, and that `Generic<T>::create` throws. The problem is that codegen did not add the template arguments in `Generic::create`, and this confuses clang++:
```
❯ clang++ -w -I. -I<path/to/runtime> test.cpp -std=c++20
test.cpp:6:35: error: 'Generic' is not a class, namespace, or enumeration
const Generic<i32> generic = TRY((Generic::create(static_cast<i32>(42))));
                                  ^
./my_generic.h:5:8: note: 'Generic' declared here
struct Generic {
       ^
test.cpp:6:20: error: no viable conversion from 'void' to 'const Generic<i32>' (aka 'const Generic<int>')
const Generic<i32> generic = TRY((Generic::create(static_cast<i32>(42))));
                   ^         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./my_generic.h:5:8: note: candidate constructor (the implicit copy constructor) not viable: cannot convert argument of incomplete type 'void' to 'const Jakt::Generic<int> &' for 1st argument
struct Generic {
       ^
./my_generic.h:5:8: note: candidate constructor (the implicit move constructor) not viable: cannot convert argument of incomplete type 'void' to 'Jakt::Generic<int> &&' for 1st argument
2 errors generated.
```
Clang fails at recognizing `Generic` as a struct, since it's really a template. Here are the contents of `./my_generic.h`:
```cpp
#include <Jakt/Error.h>

namespace Jakt {
template<typename T>
struct Generic {
[...]

PR #1026, opened 2022-08-02
DRAFT: Staged scheduler

cybergsus:
This is an attempt to remove the `run-one.sh` script by integrating the
different stages of compiling a test binary into the Jakttest scheduler.

EDIT: Managed to work around #1025. Now scheduler doesn't work correctly, but at least I can compile it.

### DIFF
--- a/jakttest/jakttest.jakt
+++ b/jakttest/jakttest.jakt
@@ -47,11 +47,12 @@ import extern "fs.h" {
         // StatResults.exists to false.
         extern function stat_silencing_enoent(path: String) throws -> StatResults?
 
-
+        // Changes the current working directory of the running process
+        extern function chdir(path: String) throws
     }
 }
 
-// FIXME: These three functions should be under a `namespace fs`, but due to
+// FIXME: These four functions should be under a `namespace fs`, but due to
 // #1008 it will ignore the stuff from the import extern if I do.
 
 function is_directory(path: String) throws -> bool {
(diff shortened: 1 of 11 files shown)


jakttest/jakttest.jakt
```diff
@@ -334,31 +405,124 @@ struct Options {
...
+
+class CollectedReader {
+    collected_buffer: [u8]
+    raw_buffer: raw u8
+    raw_buffer_read: u16
+    fd: i32
+
+    public function init(fd: i32) throws -> CollectedReader {
+        // NOTE: mut so codegen does not declare it as const u8 *
+        mut raw_buffer = libc::malloc(size: 4096)
```
ADKaster, line 461: this feels very awkward. does Jakt runtime not have AK::ByteBuffer?

cybergsus: Probably it's there but I just didn't know of it. Anyway, this is nowhere near finished. The scheduler gets stuck after the first chunk of jobs finishes, panics sometimes (maybe it's the queue implementation, I don't know), fails to properly collect stdout/stderr over the pipes (thus marking tests as failed because it couldn't find the error). Plus, the fork()/dup2()/exec() has to be replaced with an API that implements the needed `posix_spawn`* functions while allowing a similar approach for Windows, since the Windows approach goes through a 'builder pattern' in a similar way that `posix_spawn` functions are meant to be used.

cybergsus: I'm going to throw out the pipes and just create the `.err`/`.out`/`.cpp` files directly and redirect the output to them, saving just their handle. This will just need a little add-on to `File` to get `fileno` functionality, and will remove the polling of buffers that is not even working properly.

PR #1027, opened 2022-08-02
parser+typechecker+tests: Add support for 'is' in 'guard'

jntrnr:
This adds support for using the `is` clause in `guard` statements. The result is a way to match and bind without rightward drift:

```
> cat ./samples/guard/is_guard_compound_matched.jakt 
/// Expect: selfhost-only
/// - output: "PASS\n"

enum Foo {
    Bar(i64)
}

function main() {
    let foo = Foo::Bar(5)
    let bar = Foo::Bar(7)

    guard foo is Bar(x) and x < 10 else {
        println("not matched")
        return -1
    }

    println("hello there: {}", x)
}
> ./build/is_guard_compound_matched 
hello there: 5
```

### DIFF
--- a/samples/enums/is_variant_binding_compound_unmatched.jakt
+++ b/samples/enums/is_variant_binding_compound_unmatched.jakt
@@ -1,0 +1,16 @@
+/// Expect: selfhost-only
+/// - output: "did not match\n"
+
+enum Foo {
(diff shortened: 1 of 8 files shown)


jntrnr:
Found another case I want to add. Turning to draft for now.

PR #1030, opened 2022-08-03
all: Implement array slices

robryanx:
Create an ArraySlice from an Array using `arr[start_index..end_index]`
where start_index is the first position in the slice and end_index is
one more than the last position in the slice.

### DIFF
--- a/samples/arrays/slice.jakt
+++ b/samples/arrays/slice.jakt
@@ -1,0 +1,25 @@
+/// Expect:
+/// - output: "slice size: 2\n1\n2\n[1, 2]\nslice first: 1\nslice last: 2\nslice contains 2: true\nslice contains 3: false\narray index 0: 0\nslice index 0: 0\n"
+
+function main() {
+    let x = [1, 2, 3, 4, 5]
+    mut slice = x[0..2]
+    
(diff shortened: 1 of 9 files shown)


samples/arrays/slice.jakt
```diff
@@ -0,0 +1,25 @@
+/// Expect:
+/// - output: "slice size: 2\n1\n2\n[1, 2]\nslice first: 1\nslice last: 2\nslice contains 2: true\nslice contains 3: false\narray index 0: 0\nslice index 0: 0\n"
+
+function main() {
+    let x = [1, 2, 3, 4, 5]
+    mut slice = x[0..2]
```
jntrnr, line 6: I like this - out of curiosity what is the type in jakt for these? And can a slice be converted into an array?

I could imagine wanting to do `x[0..2].into_array()` or something if you know where you wanted to slice.

robryanx: This is using `ArraySlice` in the runtime.
Yes, it would be useful to be able to convert it, I can look into that.
Also going to look into extending the range expression to support for example `x[..2]`, `x[2..]`, `x[..]`.
Probably best to get this one finished do those as a new PR.

PR #1034, opened 2022-08-03
runtime:  Do not error when writing zero bytes to a File

cybergsus:
File::write might get an empty array, so fwrite() returning zero does not
always mean that an error occurred. Better check that an error actually
occurred before returning an Error(errno=0).


awesomekling:
How about we check for 0 before even calling `fwrite`, and early return if so?

cybergsus:
Seems like a better option, since that's the only other case where fwrite() can return zero without having an error.

Issue #677, opened 2022-06-17
Typechecker: Enum member functions cannot see other enums' members that are declared afterwards


greytdepression:
The following code
```jakt
enum FooBar {
    A

    function make_foo() -> Foo {
        return Foo::Bar
    }
}

enum Foo {
    Bar
}

function main() {
    let b = FooBar::make_foo()
}
```
throws compile error
```
Error: variable 'Bar' not found
-----
 6  |     function make_foo() -> Foo { 
 7  |         return Foo::Bar 
                     ^- variable 'Bar' not found
 8  |     } 
-----
```
On the other hand, if we replace `enum FooBar` with `struct FooBar` the code does compile.
To fix this we should adapt the code for parsing enum member functions to work similarly to how struct member functions are parsed :^)

greytdepression:
I played around with it some more and this problem is not unique to enums. One can provoke the same problem with structs:
```jakt
struct FooBar {
    a: i64

    function make_foo() -> Foo {
        return Foo(bar: 0)
    }
}

struct Foo {
    b: i64
}
```
The reason it works when we mix structs and enums seems to be that enum function declarations get resolved before struct ones do. So the bug seems to be in the logic of how predeclaration is handled.

lanmonster:
fixed as of 1ea81c6

PR #1040, opened 2022-08-06
typechecker: typechecking ordering fixes

### LINKED ISSUE #527: Typechecker: enum methods can't see structs
Example:
```
struct Foo {
    id: usize
}

enum What {
    Bar
    Baz

    function count(this) {
        Foo(id: 100)
    }
}

function main() {
    
}
```

Errors with: "Could not resolve call to Foo"

robryanx:
Fixes #527 #677
This also allows a number of refactors such as moving `expression_type` to a method `type` in the `CheckedExpression` enum.
I will submit the refactoring as a separate PR.

### DIFF
--- a/samples/enums/enum_recursive_method.jakt
+++ b/samples/enums/enum_recursive_method.jakt
@@ -1,0 +1,22 @@
+/// Expect:
+/// - output: "true\n"
+
+struct CheckedVariable {
+    is_mutable: bool
+}
+
+boxed enum Foo {
+    Var(var: CheckedVariable)
+    IndexedStruct(expr: Foo)
+
+    function is_mutable(this) -> bool => match this {
+        Var(var) => var.is_mutable
+        IndexedStruct(expr) => expr.is_mutable()
+    }
+}
+
+function main() {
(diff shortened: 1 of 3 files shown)


selfhost/parser.jakt
```diff
@@ -291,6 +259,38 @@ boxed enum ParsedStatement {
     }
 }
 
+struct ParsedBlock {
```
jntrnr, line 262: does this need to move because of what you're doing in the PR?

robryanx: Yes, previously `typecheck_enum` on enums was called before `typecheck_struct` on structs. As they are in the same loop now the ordering for type checking methods in enums/structs becomes important.

awesomekling: Hmm, the language is supposed to allow arbitrary declaration order. If the ordering is significant, then I think something has gone wrong :)

robryanx: Ah, yes that does make sense :)
The second commit here for enum function return type actually fixed the ordering issue, I have removed the reorder.

PR #1043, opened 2022-08-06
build: Use -fno-exceptions when compiling jakt-generated cpp code

ilyapopov:
This cuts down build time spent in Clang by 2.7s, or 23%.

The open question is if jakt will ever support C++ exceptions (generated, for example, int the inline c++ code blocks, or third party c++ libraries called from jakt code)

```
hyperfine -r 5 "cd jakt/build/ && clang++ -std=c++20 -I ../runtime/ -Wno-parentheses-equality -Wno-trigraphs -Wno-user-defined-literalsi jakt.cpp" "cd jakt/build/ && clang++ -std=c++20 -I ../runtime/ -Wno-parentheses-equality -Wno-trigraphs -Wno-user-defined-literals -fno-exceptions jakt.cpp"
Benchmark 1: cd jakt/build/ && clang++ -std=c++20 -I ../runtime/ -Wno-parentheses-equality -Wno-trigraphs -Wno-user-defined-literalsi jakt.cpp
  Time (mean ± σ):     14.772 s ±  0.092 s    [User: 14.208 s, System: 0.552 s]
  Range (min … max):   14.666 s … 14.881 s    5 runs
 
Benchmark 2: cd jakt/build/ && clang++ -std=c++20 -I ../runtime/ -Wno-parentheses-equality -Wno-trigraphs -Wno-user-defined-literals -fno-exceptions jakt.cpp
  Time (mean ± σ):     11.989 s ±  0.069 s    [User: 11.431 s, System: 0.541 s]
  Range (min … max):   11.874 s … 12.059 s    5 runs
 
Summary
  'cd jakt/build/ && clang++ -std=c++20 -I ../runtime/ -Wno-parentheses-equality -Wno-trigraphs -Wno-user-defined-literals -fno-exceptions jakt.cpp' ran
    1.23 ± 0.01 times faster than 'cd jakt/build/ && clang++ -std=c++20 -I ../runtime/ -Wno-parentheses-equality -Wno-trigraphs -Wno-user-defined-literalsi jakt.cpp'
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ set(FINAL_STAGE "1" CACHE STRING "Compiler stage to stop at, either 1 (default) 
 
 macro(add_jakt_compiler_flags target)
   target_compile_options("${target}" PRIVATE
+    -fno-exceptions
     -Wno-unused-local-typedefs
     -Wno-unused-function
     -Wno-unknown-warning-option
--- a/build.ninja
+++ b/build.ninja
@@ -3,7 +3,7 @@
 # Copyright (c) 2022, JT <jt@serenityos.org>
 #
 # SPDX-License-Identifier: BSD-2-Clause
-cxxflags = -fcolor-diagnostics -std=c++20 -Wno-user-defined-literals -Wno-deprecated-declarations -Wno-parentheses-equality -Wno-unqualified-std-cast-call -Wno-unknown-warning-option -Wno-trigraphs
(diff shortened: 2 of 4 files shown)


awesomekling:
Super awesome! We don't care about C++ exceptions, so it's not a problem :)

maddanio:
> Super awesome! We don't care about C++ exceptions, so it's not a problem :)

Hmm, but with extern c you might want to call throwing apis and guard them with catch.

awesomekling:
SerenityOS does not use C++ exceptions, and this is a language specifically for SerenityOS.

PR #1044, opened 2022-08-06
parser+typechecker: Implement default args for functions/methods

adleroliveira:
This PR adds the possibility to use default arguments for functions and method calls.

```
function say(text: String = "WHF", number: i64 = 5, boolean: bool = false, optional: String? = None) {
    println("{} {} {} {}", text, number, boolean, optional)
}

function main() {
    say()
}

// output: "WHF 5 false None\n"
```

### DIFF
--- a/samples/functions/default_arguments.jakt
+++ b/samples/functions/default_arguments.jakt
@@ -1,0 +1,10 @@
+/// Expect:
+/// - output: "WHF\n"
+
+function say(text: String = "WHF") {
+    println("{}", text)
+}
+
+function main() {
+    say()
+}
--- a/samples/functions/default_arguments_bad_type.jakt
+++ b/samples/functions/default_arguments_bad_type.jakt
@@ -1,0 +1,9 @@
+/// Expect:
+/// - output: error: "Type mismatch: expected ‘String’, but got ‘i64’\n"
+function say(text: String = 3) {
+    println("{}", text)
+}
+
+function main() {
+    say()
+}
--- a/samples/functions/default_arguments_bad_type.jakt
+++ b/samples/functions/default_arguments_bad_type.jakt
@@ -1,5 +1,6 @@
 /// Expect:
-/// - output: error: "Type mismatch: expected ‘String’, but got ‘i64’\n"
+/// - error: "Type mismatch: expected ‘String’, but got ‘i64’\n"
+
 function say(text: String = 3) {
     println("{}", text)
 }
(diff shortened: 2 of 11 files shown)


jntrnr:
Nice!

You'll probably want a few more tests, as there are some corner cases:
* What if you only pass some, but not all, of the defaults?
* What if the user makes some parameters required and others default?
* If the user puts the default params before the required params, does it error?
* Does this work for methods also?

Questions for future PRs:
* do default params work for closures?
* If so, what happens if one of the default values closes over a variable outside of the closure?

adleroliveira:
> 

> Nice!
> 
> You'll probably want a few more tests, as there are some corner cases:
> 
> * What if you only pass some, but not all, of the defaults?
> * What if the user makes some parameters required and others default?
> * If the user puts the default params before the required params, does it error?
> * Does this work for methods also?
> 
> Questions for future PRs:
> 
> * do default params work for closures?
> * If so, what happens if one of the default values closes over a variable outside of the closure?


1. What if you only pass some, but not all, of the defaults?

This is addressed in this test:
samples/functions/default_arguments_mixed.jakt
```
function say(text: String, number: i64 = 5, boolean: bool, optional: String?, byte: u8 = b'a') {
    println("{} {} {} {}", text, number, boolean, optional, byte)
}

function main() {
    say(text: "WHF", boolean: true, optional: None)
}
```

2. What if the user makes some parameters required and others default?

Not sure if I understand the case, but I think this test address it:
samples/functions/default_arguments_some_defaults.jakt
```
function say(text: String = "WHF", number: i64 = 5, boolean: bool = false, optional: String? = None) {
    println("{} {} {} {}", text, number, boolean, optional)
}

function main() {
    say(text: "Hello", optional: "World")
}
```

3.  If the user puts the default params before the required params, does it error?

It does not, I address it in the following test:
samples/functions/default_arguments_default_before_required.jakt
```
/// Expect:
/// - output: "Hello 5 false WHF 121\n"

function say(text: String = "Hello", number: i64 = 5, boolean: bool = false, optional: String?, byte: u8) {
    println("{} {} {} {} {}", text, number, boolean, optional, byte)
}

function main() {
    say(optional: "WHF", byte: b'y')
}
```

4. Does this work for methods also?

It does. I address it in the following test
samples/functions/default_arguments_method.jakt
```
/// Expect:
/// - output: "Well, hello friends. My name is Jane and I am 36 years old.\n"

class Person {
    public function greet(this, name: String = "Jane", age: i64 = 36) {
        println("Well, hello friends. My name is {} and I am {} years old.", name, age)
    }
}

function main() {
    let p = Person()
    p.greet()
}
```

jntrnr:
Cool, thanks for the clarifications.

We just landed a massive PR, so you'll need to fix a few conflicts before we can finish up.

jntrnr:
good job - looks good!

PR #1046, opened 2022-08-06
Everywhere: Compiletime Function Execution, you get a consteval, you get a const, everyone gets CTFE

alimpfard:
TLDR: Replace `function` in function decls with `comptime` to have calls to that function execute at compile time.

Moar details:
- Jakt interpreter
- Value/AST synthesis
- Lots of things implemented, lots more not implemented
- Also contains a REPL (need to link against Lagom to work - see commit details)

TODO:
- ~~Moar tests~~
- ~~Moar documentation~~
- ~~Moar tidy commits~~


Hendiadyoin1:
For atomicity it might be nice to update the bootstrap before using the new feature in the compiler

CMakeLists.txt
```diff
@@ -75,9 +75,9 @@ set(SELFHOST_SOURCES
   selfhost/ide.jakt
```
Hendiadyoin1, line 75: At this point stage0 will not know about `comptime`
So when bisecting this commit would fail

alimpfard: There's not much else I can do besides squashing the two commits.

PR #1051, opened 2022-08-07
tests+jakttest: Minor fixes and changes

CountBleck:
This PR changes a few small things:
1. I added back a missing slash in a test's header.
2. I removed every reference to "selfhost-only", because we don't use the selfhost compiler.
3. I removed the "Build Dependencies" section in Jakttest's README, since it's now built with the rest of Jakt.

### DIFF
--- a/jakttest/README.md
+++ b/jakttest/README.md
@@ -5,16 +5,11 @@ It's also written in Jakt.
 
 ## Build dependencies
 
-To build Jakttest you should use the `ninja` build system which will use the provided `build.ninja` file.
-It needs the same `clang++` that is required to build any other Jakt file, and `python` 3.6+ to run the
-patch script that patches the generated C++ from Jakt code to accept the external C++ code.
+Jakttest is built along with Jakt and therefore needs the same dependencies.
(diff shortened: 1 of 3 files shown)


jakttest/README.md
```diff
@@ -3,18 +3,7 @@
...
-## Build dependencies
-
-To build Jakttest you should use the `ninja` build system which will use the provided `build.ninja` file.
-It needs the same `clang++` that is required to build any other Jakt file, and `python` 3.6+ to run the
-patch script that patches the generated C++ from Jakt code to accept the external C++ code.
-
-As a list:
-
-- `ninja` (or `ninja-build` on some package managers)
-- `clang++`
```
jntrnr, line 15: This should be changed to saying we require `cmake` and `clang++` rather than being removed, with the text in the paragraph changed to match.

CountBleck: That seems redundant, because Jakttest is built with Jakt anyway.

jntrnr: just following up from discord chat - the README here should change to point people to the main README for build instructions

CountBleck: How do the current changes look?

PR #1052, opened 2022-08-07
Pull in the repl as `jakt --repl`

alimpfard:
Enough said.

Make sure to use `rlwrap` or face the wrath of stdin.

### DIFF
--- a/selfhost/repl.jakt
+++ b/selfhost/repl.jakt
@@ -6,43 +6,86 @@ import utility { Span }
 import error { JaktError }
 import codegen { CodeGenerator, CodegenDebugInfo, ControlFlowState, AllowedControlExits }
 
-import extern "LibLine/Editor.h" {
-    namespace AK {
-        extern struct String {
-            function characters(this) -> raw c_char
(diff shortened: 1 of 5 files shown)


selfhost/repl.jakt
```diff
@@ -6,43 +6,86 @@ import utility { Span }
...
+    function create(prompt: String) throws -> Editor {
+        mut std_in = fopen("/dev/stdin".c_string(), "r".c_string())
+        if std_in == null<FILE>() {
+            eprintln("Could not open /dev/stdin for reading")
+            throw Error::from_errno(42)
+        }
+
+        let editor = Editor(
+            standard_input_file: std_in
+            line_pointer: allocate<c_char>(count: 4096)
```
ADKaster, line 54: Did we not carry ByteBuffer over from AK?

Alternatively, is there like an "unsafe" way to get the .data() of a Jakt::Array of bytes?

alimpfard: we did not, and there is not.

PR #1057, opened 2022-08-08
selfhost: Make '--run' and '-r' execute a file with the interpreter

alimpfard:
For faster dev cycles...when the interpreter can run your code :P

cc @jntrnr, you asked for dis.


jntrnr:
oh yes, this is awesome

PR #1061, opened 2022-08-09
codegen+ide+typechecker: Refactor methods to be part of enums

robryanx:
Methods that operate on an enum that were previously passed as an
argument to a function due to compiler limitations have been refactored 
to be part of the enum that they are operating on.

`expression_type(expr)` changed to `expr.type()`
`expression_can_throw(expr)` changed to `expr.can_throw()`
`expression_span(expr)` changed to `expr.span()`
`expression_is_mutable(expr)` changed to `expr.is_mutable(program)`
`promote(num_const, type_id, program)` changed to
`num_const.promote(type_id, program)`


awesomekling:
Much nicer :)

PR #1065, opened 2022-08-10
runtime: Add join function on Array and ArraySlice

robryanx:
This takes advantage of the existing string formatters to allow an Array
or ArraySlice of any type to be joined into a string.

### DIFF
--- a/runtime/prelude.jakt
+++ b/runtime/prelude.jakt
@@ -22,6 +22,7 @@ extern struct Array<T> {
     function iterator(this) -> ArrayIterator<T>
     function first(this) -> T?
     function last(this) -> T?
+    function join(this, anon separator: String) throws -> String
 }
 
 extern struct ArraySlice<T> {
@@ -32,6 +33,7 @@ extern struct ArraySlice<T> {
     function to_array(this) throws -> Array<T> 
     function first(this) -> T?
     function last(this) -> T?
+    function join(this, anon separator: String) throws -> String
 }
 
 extern struct String {
--- a/runtime/Builtins/Array.cpp
+++ b/runtime/Builtins/Array.cpp
@@ -1,0 +1,43 @@
+#pragma once
+
+#include <Builtins/Array.h>
+#include <errno.h>
+
+namespace JaktInternal {
+using namespace Jakt;
+
+template<typename T>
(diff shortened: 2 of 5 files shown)


runtime/prelude.jakt
```diff
@@ -22,6 +22,7 @@ extern struct Array<T> {
     function iterator(this) -> ArrayIterator<T>
     function first(this) -> T?
     function last(this) -> T?
+    function join(this, anon s: String) throws -> String
```
networkException, line 25:
```suggestion
    function join(this, anon separator: String) throws -> String
```
Would be more consistent and expressive

runtime/prelude.jakt
```diff
@@ -31,6 +32,7 @@ extern struct ArraySlice<T> {
...
```
networkException, line 35: Same as above

runtime/prelude.jakt
robryanx: Yeah, if preferable could change this. There is kind of a mix currently in descriptive names and one letter names for anon parameters in prelude currently.
In effect no one would actually use the parameter name in this case.

networkException: still adding more code that violates the code style also isn't great

robryanx: Sure makes sense, I have updated them.

alimpfard (commented):
🤔 StringBuilder has a `join` in Serenity, should we do that instead?
```c++
builder.join(Iterable, separator)
```

I'm somewhat reluctant to add overly generic things to the prelude that can be provided by default via (the soon to be ready hopefully) traits

robryanx:
> StringBuilder has a `join` in Serenity, should we do that instead?
> 
> ```c++
> builder.join(Iterable, separator)
> ```
> 
> I'm somewhat reluctant to add overly generic things to the prelude that can be provided by default via (the soon to be ready hopefully) traits

I looked into this first, I think it ends up making the syntax a lot more verbose unless I am missing something
```
let collection = ["A", "B"]
mut sb = StringBuilder::create()
sb.join(",", collection)
    
println("{}", sb.to_string())
```
How would traits interact with the prelude classes? They would just be added on the jakt side?

robryanx:
Will close this and revisit once we can use traits with prelude classes.

PR #1066, opened 2022-08-10
typechecker: Typecheck format-strings and don't allocate them as Strings

Hendiadyoin1:
This will stop us from writing illegal format-strings, which would
otherwise cause segfaults at runtime.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -5257,6 +5257,70 @@ struct Typechecker {
         return None
     }
 
+    function typecheck_format(mut this, args: [(String, CheckedExpression)]) throws {
+        guard args[0].1 is QuotedString(val: format_string, span) else{
+            eprintln("Internal Compiler Error: `typecheck_format` called with dynamic format string: `{}`", args[0].1)
+            abort()
+        }
+        
+        mut n_args = 1uz
+        mut index = 0uz
+        mut in_format_specifier = false
+
+        while index < format_string.length() {
+            let current_span = Span(file_id: span.file_id, start: span.start + index + 1, end: span.start + index + 1)
+            if in_format_specifier {
+                guard format_string.byte_at(index) != b'{' else {
+                    .error("Found '{' inside of format specifier", current_span)
+                    return
+                }
+
+                // FIXME: check for valid formatting specifiers, like ':x'
+
+                if format_string.byte_at(index) == b'}' {
+                    in_format_specifier = false
+                }
+
+            } else {
+                guard format_string.byte_at(index) != b'}' else {
+                    guard index + 1 == format_string.length() else {
+                        guard format_string.byte_at(index + 1) == b'}' else {
+                            .error("Unmatched '}' in format string", current_span)
+                            return
+                        }
+                        index+=2
+                        continue
+                    }
+                .error("Unmatched '}' in format string", current_span)
+                    return
+                }
+                guard format_string.byte_at(index) == b'{' else {
+                    index++
+                    continue
+                }
+
+                guard index + 1 != format_string.length() else{
+                    .error("Unmatched '{' in format string", current_span)
+                    return
+                }
+
+                if format_string.byte_at(index + 1) == b'{' {
+                    index++
+                } else {
+                    n_args++
+                    in_format_specifier = true
+                }
+            }
+            index++
+        }
+
+        guard n_args == args.size() else {
+            .error(format("Wrong number of arguments to format (expected: {}, got: {})", n_args - 1, args.size() - 1), span)
+            return
+        }
+
+    }
+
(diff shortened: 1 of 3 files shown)


jntrnr:
Hmm, I'm trying to remember - I thought we didn't want indexing of strings because we wanted to move to utf8?

Hendiadyoin1:
Made is use `String::byte_at` now

Hendiadyoin1:
Update, now works on `println` etc and the format string on its own should not allocate anymore

selfhost/typechecker.jakt
```diff
@@ -915,6 +915,7 @@ boxed enum CheckedExpression {
     Boolean(val: bool, span: Span)
     NumericConstant(val: CheckedNumericConstant, span: Span, type_id: TypeId)
     QuotedString(val: String, span: Span)
+    FormatString(val: String, span: Span)
```
alimpfard, line 918: This is _far_ too "special" to be useful, especially as the current format of these functions is temporary until we get variadic generics
Let's keep the check for literals, and not overfit to a temporary state (split this commit at least)

Hendiadyoin1: I made it its own commit for now

Hendiadyoin1:
I think I fixed it up with the new layout and made it work with the interpreter,
@alimpfard can you check this again?

alimpfard:
First commit lgtm, but as I said before, I don't like the entirely separate expression type, it's a hack on top of a hack (specialised typecheck for format) meant to bypass another hack (no overloaded strings or literal "optimisation") - that's too many hacks.

Hendiadyoin1:
Well then I will drop that part for now...

selfhost/typechecker.jakt
```diff
@@ -5257,6 +5257,70 @@ struct Typechecker {
...
 
+    function typecheck_format(mut this, args: [(String, CheckedExpression)]) throws {
+        guard args[0].1 is QuotedString(val: format_string, span) else{
+            eprintln("Internal Compiler Error: `typecheck_format` called with dynamic format string: `{}`", args[0].1)
+            abort()
+        }
+        
+        mut n_args = 1uz
+        mut index = 0uz
+        mut in_format_sepcifier = false
```
networkException, line 5268:
```suggestion
        mut in_format_specifier = false
```

Hendiadyoin1: I cant write..

Good catch, will fix it up later this evening

selfhost/typechecker.jakt
```diff
@@ -5257,6 +5257,70 @@ struct Typechecker {
...
+                guard format_string.byte_at(index) != b'}' else {
+                    guard index + 1 == format_string.length() else {
+                        guard format_string.byte_at(index + 1) == b'}' else {
+                            .error("Unmatched '}' in format string", current_span)
+                            return
+                        }
+                        index+=2
+                        continue
+                    }
+                .error("Unmatched '}' in format string", current_span)
```
alimpfard, line 5294: Indent this

selfhost/typechecker.jakt
```diff
@@ -5257,6 +5257,70 @@ struct Typechecker {
...
+        
+        mut n_args = 1uz
+        mut index = 0uz
+        mut in_format_specifier = false
+
+        while index < format_string.length() {
+            let current_span = Span(file_id: span.file_id, start: span.start + index + 1, end: span.start + index + 1)
+            if in_format_specifier {
+                guard format_string.byte_at(index) != b'{' else {
+                    .error("Found '{' inside of format specifier", current_span)
```
alimpfard, line 5274: This is not right,
```
format("{: >{}}", ...)
```
Pads with spaces to the length given.

selfhost/typechecker.jakt
```diff
@@ -5257,6 +5257,70 @@ struct Typechecker {
...
+                }
+
+            } else {
+                guard format_string.byte_at(index) != b'}' else {
+                    guard index + 1 == format_string.length() else {
+                        guard format_string.byte_at(index + 1) == b'}' else {
+                            .error("Unmatched '}' in format string", current_span)
+                            return
+                        }
+                        index+=2
```
alimpfard, line 5291: Spaces around += pls

selfhost/typechecker.jakt
```diff
@@ -5257,6 +5257,70 @@ struct Typechecker {
...
+
+        while index < format_string.length() {
+            let current_span = Span(file_id: span.file_id, start: span.start + index + 1, end: span.start + index + 1)
+            if in_format_specifier {
+                guard format_string.byte_at(index) != b'{' else {
+                    .error("Found '{' inside of format specifier", current_span)
+                    return
+                }
+
+                // FIXME: check for valid formatting specifiers, like ':x'
```
alimpfard, line 5278: Also add a fixme for explicit indices like `"{1}"`, I'd suggest looking at AK's comptime checker, that's been tested enough to be reasonably correct

PR #1067, opened 2022-08-11
all: Support Range shorthands

robryanx:
For example `1..` `..2` `..`

In the context of general assignment
`..2` is shorthand for `0..2`
`1..` is shorthand for `1..9223372036854775807` (LLONG_MAX)
`..` is shorthand for `0..9223372036854775807`

In the context of creating an ArraySlice
`[1..]` is shorthand for `[1..arr.size()]`
`[..2]` is shorthand for `[0..2]`
`[..]` is shorthand for `[0..arr.size()]`

Also refactor to remove `IndexedRangeExpression` (As `IndexedExpression` can just contain a range.
Also fix a bug where it wasn't possible to use `Is` to match an Enum variant that also matched by name to types in parent scopes.


samples/arrays/array_bad2.jakt
```diff
@@ -1,5 +1,5 @@
 /// Expect:
-/// - error: "Index is not an integer"
+/// - error: "Index must be an Integer or a Range"
```
jntrnr, line 2: these shouldn't be capitalised, unless that's the name of the type

jntrnr:
A few questions:

* Can ranges be passed as values into functions, held in variables, and held in structs?
* If so, what does that look like? We probably want some tests for it
* Can ranges have expressions on each end? We'll probably want tests for that as well

jntrnr:
Oh, I just noticed this isn't adding ranges, it's adding the open-ended bit of ranges. Still probably worth a thought about how open-ended ranges and the cases I mentioned.

samples/arrays/array_bad2.jakt
robryanx: Sure, have updated this.

robryanx:
> Oh, I just noticed this isn't adding ranges, it's adding the open-ended bit of ranges. Still probably worth a thought about how open-ended ranges and the cases I mentioned.

I have added some more test cases around what you mentioned, let me know if you can think of anything else that should be considered.
Originally I thought it would be a parsing issue to add support for open ended endings outside of the array context but it seems fine so I added that eg.
`1..` is shorthand for `1..9223372036854775807` (LLONG_MAX)
Probably not overly useful but good for completeness.

runtime/Jakt/StringBuilder.h
```diff
@@ -153,6 +153,23 @@ struct Formatter<JaktInternal::Set<T>> : Formatter<StringView> {
     }
 };
 
+template<typename T>
+struct Formatter<JaktInternal::Range<T>> : Formatter<StringView> {
+    ErrorOr<void> format(FormatBuilder& builder, JaktInternal::Range<T> const& range)
+    {
+        if (m_alternative_form)
+            JaktInternal::_pretty_print_enabled = true;
```
alimpfard, line 161: Who sets this back to its original value?

selfhost/parser.jakt
```diff
@@ -3614,7 +3640,20 @@ function parsed_expression_equals(anon lhs_expression: ParsedExpression, anon rh
         else => false
     }
     Range(from: lhs_from, to: lhs_to) => match rhs_expression {
-        Range(from: rhs_from, to: rhs_to) => parsed_expression_equals(lhs_from, rhs_from) and parsed_expression_equals(lhs_to, rhs_to)
+        Range(from: rhs_from, to: rhs_to) => {
+            mut equal = false
+            if lhs_from.has_value() == rhs_from.has_value() {
+                if lhs_to.has_value() == rhs_to.has_value() {
+                    if lhs_from.has_value() && lhs_to.has_value() {
+                        equal = parsed_expression_equals(lhs_from!, rhs_from!) and parsed_expression_equals(lhs_to!, rhs_to!)
```
alimpfard, line 3648: Maybe for a future PR: sounds like an excuse to move expression_equals into ParsedExpression, and do `lhs_from?.equals(rhs_from ?? ParsedExpression::Garbage)` or something (maybe even expose Optional::map)

runtime/Jakt/StringBuilder.h
robryanx: Yeah good catch, seems it wasn't reset in any cases, I had just copied it across from the array formatter.
Added a commit to fix it in runtime and codegen.

selfhost/parser.jakt
robryanx: Yeah would be a lot cleaner. How does Optional::map work?

alimpfard: Takes a function of `T -> U`, if the optional has a value, invokes the function and returns the value, otherwise returns none, overall yields `U?`.

```
extern struct Optional<T> {
    function map<U>(anon mapper: function(anon x: T) -> U) -> U?
}
```

should model it in jakt (it can handle throwing functions too, but Jakt can't overload like that yet).

alimpfard (approved):
C++ bits LGTM.
New syntax feels odd, but I could get used to it, so LGTM.
Indexed(Range)Expression merge :+1:

selfhost/parser.jakt
robryanx: Thanks, will take a look in a separate PR.

PR #1072, opened 2022-08-12
README: Improve the cast semantics description + reference casts

kleinesfilmroellchen:
The cast semantics for floats and ints is refined with this description, as was discussed on Discord.

Reference casts should be just another way of using reference and dereference operators, but they would allow for some convenient combination of referencing and casting in one operation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,8 @@ Null pointers are not possible in safe mode, but pointers can be wrapped in `Opt
 
 Note that **weak** pointers must always be wrapped in `Optional`. There is no **weak T**, only **weak T?**.
 
+Additionally, there are the reference types (see below) `&T` and `&mut T`, providing safe non-null pointers.
+
 ## Math safety
 
 - [x] Integer overflow (both signed and unsigned) is a runtime error.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -385,21 +387,47 @@ There are two built-in casting operators in **Jakt**.
...
-    - A parent can be cast to a child, but this will check the type at runtime and fail if the object was not of the child type or one of its subtypes.
-- If the types are incompatible, a user-defined cast is attempted to be used. The details here are not decided yet.
-- If nothing works, the cast will not even compile.
-
-Additional casts are available in the standard library. Two important ones are `as_saturated` and `as_truncated`, which cast integral values while saturating to the boundaries or truncating bits, respectively.
+- [ ] Casts to the same type are infallible and pointless, so might be forbidden in the future.
+- [ ] If both types are primitive, a safe conversion is done.
+    - [ ] Integer casts will fail if the value is out of range. This means that promotion casts like i32 -> i64 are infallible.
+    - [ ] Float -> Integer casts truncate the decimal point, if the float is a normal numeric value. If the float is any NaN or either Infinity, the cast fails.
+    - [ ] Integer -> Float casts resolve to the closest value to the integer representable by the floating-point type. If the integer value is too large, the cast fails.
```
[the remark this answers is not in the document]
CountBleck: `i64`/`u64` casts to `f32` could be too large.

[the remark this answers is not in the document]
Hendiadyoin1: The maximum value of a float32 is `(2 − 2−23) × 2127` which way is larger than `2^64 - 1`, so that does not hold true, the maximum accurately representable integer is 2^24 or so, so rounding would occur

README.md
```diff
@@ -385,21 +387,46 @@ There are two built-in casting operators in **Jakt**.
...
+- If nothing works, the cast will not compile.
+
+Some examples of possible and not possible casts (the left side shows the data type of a `value`, not the `value` itself):
+```
+u32 as? u16  // might not succeed
+u16 as! i32  // infallible
+f32 as? u32  // normally truncates, but values like NaN cause failure
+u64 as! bool // only false if the value is 0
+A as! B      // infallible assuming A is a subclass of B
+&mut A as! A // creates a copy assuming A is a struct type; equivalent to (*value)
```
alimpfard, line 418: does this actually compile? `verify_cast<T>(T&)` sounds vaguely possible, but I'm not sure

README.md
```diff
@@ -385,21 +387,46 @@ There are two built-in casting operators in **Jakt**.
...
+i32 as! &i32 // equivalent to (&value)
```
alimpfard, line 419: :yakfused: this sounds very not possible, what if the lhs is a temporary?

README.md
```diff
@@ -385,21 +387,46 @@ There are two built-in casting operators in **Jakt**.
...
+    - [ ] Casts between infallible pointer types can happen alongside casts within the contained type, if the pointer type doesn't change or is cast strictly upwards on the pointer type hierarchy. This hierarchy goes (from bottom to top) `&mut T` -> `&T` -> `T`.
+- If the types are part of the same type hierarchy (i.e. one is a child type of another), casts are possible according to these rules. Note that this also goes if the types are the same pointer type and the contained type is part of the same type hierarchy (which applies recursively), so `&mut A` and `&mut B` are part of the same type hierarchy if `A` and `B` are.
+    - [ ] A child can be cast to its parent infallibly.
+    - [ ] A parent can be cast to a child, but this will check the type at runtime and fail if the object was not of the child type or one of its subtypes.
+- [ ] If the types are incompatible, a user-defined cast is attempted to be used. The details here are not decided yet, but it will involve implementing some generic trait.
+- If nothing works, the cast will not compile.
+
+Some examples of possible and not possible casts (the left side shows the data type of a `value`, not the `value` itself):
```
alimpfard, line 411: but nothing indicates what's possible and what's not?

README.md
```diff
@@ -385,21 +387,46 @@ There are two built-in casting operators in **Jakt**.
...
+    - [ ] Any primitive -> bool will create `true` for any value except 0, which is `false`.
+    - [ ] bool -> any primitive will do `false -> 0` and `true -> 1`, even for floats.
+- If the types are two different pointer types of the same underlying type `T`, the cast is generally allowed and infallible.
+    - [ ] A cast to `T` from any pointer type, if `T` is reference counted, will increment the reference count as expected; that's the preferred way of creating a strong reference from a weak or temporary reference.
+    - [ ] A cast to `T` from any pointer type, if `T` is *not* reference counted, will copy the value.
+    - [ ] Casts from `&T` to `&mut T` is forbidden. The other way around is allowed.
+    - [ ] A cast from and to `raw T` is unsafe but infallible.
+    - [ ] A cast to `&T` or `&mut T` is allowed for orthagonality, but has the same effect as directly taking the reference with the `&` or `&mut` operators, respectively.
+    - [ ] Casts from `&T` or `&mut T` are allowed for orthagonality and follow the same rules as dereferencing.
```
alimpfard, line 403: have we discussed this?

README.md
```diff
@@ -385,21 +387,46 @@ There are two built-in casting operators in **Jakt**.
...
-
-Additional casts are available in the standard library. Two important ones are `as_saturated` and `as_truncated`, which cast integral values while saturating to the boundaries or truncating bits, respectively.
+- [ ] Casts to the same type are infallible and pointless, so might be forbidden in the future.
+- [ ] If both types are primitive, a safe conversion is done.
+    - [ ] Integer casts will fail if the value is out of range. This means that promotion casts like i32 -> i64 are infallible.
+    - [ ] Float -> Integer casts truncate the decimal point, if the float is a normal numeric value. If the float is any NaN or either Infinity, the cast fails.
+    - [ ] Integer -> Float casts resolve to the closest value to the integer representable by the floating-point type. If the integer value is not exactly representable, the cast fails.
+    - [ ] Any primitive -> bool will create `true` for any value except 0, which is `false`.
+    - [ ] bool -> any primitive will do `false -> 0` and `true -> 1`, even for floats.
+- If the types are two different pointer types of the same underlying type `T`, the cast is generally allowed and infallible.
```
alimpfard, line 397:
```
enum Foo: i32 {
    A
    B
}

enum Bar: i32 {
    A = 1
    B
}

Foo::A as! Bar
```

Not safe, definitely not correct, and most certainly not infallible - we don't check them right now, but casting to an enum type should be guaranteed to produce one of the variants.

alimpfard (commented):
nitpicks:

README.md
kleinesfilmroellchen: Not all of these compile yet; that's why the above list is a checklist.

README.md
kleinesfilmroellchen: In a simple implementation the C++ would probably not compile. In Rust this is allowed, you could think of an invisible `let` being introduced in the expression's statement's scope (or a smaller scope; doesn't matter). We could compile it to exactly this in C++, but it's not a priority. However, it should be allowed as I find it a useful Rust feature that stops variables from being special just because they have a name; after all you can always extract a temporary to a variable and take a reference to that. There's an argument to be had about mutability of the temporary through mutable references, but I don't think that leads to weird situations which cannot be thought of as "the temporary behaves as an implicit local variable". In most cases the mutation will just not have any effects as the pseudo-variable becomes inaccessible right after the expression evaluation.

README.md
kleinesfilmroellchen: The comments do?

README.md
kleinesfilmroellchen: Yeah no, but do you disagree?

README.md
kleinesfilmroellchen: `Foo::A` is an `i32`. `Bar` is not `i32`. So this does not compile, and this is not what I mean with "pointer types", that term is reserved for `class`, `&`, `&mut`. What you're looking at is probably type aliases, but neither Foo nor Bar are a type alias of i32 like `using` would be, as (and you correctly identified this) not all instances of i32 are valid instances of Foo and Bar. It is just the case that instances of Foo and Bar are either of type i32 or of a type that we can implicitly treat as i32 (no clue how *exactly* we currently implement that).

README.md
alimpfard: It's not so much about names as it is about scopes, c++ defines the scope of a value differently than that of a bound variable.

That said, I don't think we should add docs for things we haven't decided on yet?

README.md
alimpfard: Kinda, I'd have to play with the idea a bit to have a conclusive answer, but it seems like pointless feature duplication to me
`&T` -> `&U` being a separate matter of course

README.md
alimpfard: Ah, I misread that, my bad.

README.md
alimpfard: All of them read as "this is possible" to me,
> i32 as! &i32 // equivalent to (&value)

kleinesfilmroellchen: Ah right, what I mean by that is that "semantics of `&value` apply", i.e. it does not compile if `&value` does not compile.

kleinesfilmroellchen:
@alimpfard I have specifically worded the things you challenged to say "we haven't decided this yet", I hope that is okay as there's plenty of totally fine things in the list we can implement first.

PR #1074, opened 2022-08-12
parser: enable destructuring assignment on for statement

adleroliveira:
By allowing the parser to parse destructuring assignment
on `for` statement, we can make use of the already existing
ParsedStatement::DestructuringAssignment statement and just
pass it to the typechecker.

```
/// Expect:
/// - output: "1, Hello\n2, World\n"

function main() {
    mut list = [("Hello", 1), ("World", 2)]

    for (word, idx) in list.iterator() {
        println("{}, {}" idx, word)
    }
}
```


samples/tuples/destructuring_assignment_for.jakt
```diff
@@ -0,0 +1,10 @@
+/// Expect:
+/// - output: "1, Hello\n2, World\n"
+
+function main() {
+    mut list = [("Hello", 1), ("World", 2)]
+
+    for (word, idx) in list.iterator() {
```
alimpfard, line 7: 🤨 why is the index after the element?
Not a problem for the test of course, just seemed odd

selfhost/parser.jakt
```diff
@@ -2166,14 +2160,30 @@ struct Parser {
...
+                destructured_var_decls = .parse_destructuring_assignment(is_mutable: false)
+                mut tuple_var_name = ""
+                for var in destructured_var_decls.iterator() {
+                    tuple_var_name += var.name
+                    tuple_var_name += "__"
+                }
+                iterator_name = tuple_var_name
+            }
+            else => {
+                .error("Expected iterator name", .current().span())
```
alimpfard, line 2181: Iterator name or destructuring pattern

selfhost/parser.jakt
```diff
@@ -2166,14 +2160,30 @@ struct Parser {
...
+            Identifier(name) => {
+                iterator_name = name
+                .index++
+            }
+            LParen => {
+                destructured_var_decls = .parse_destructuring_assignment(is_mutable: false)
+                mut tuple_var_name = ""
+                for var in destructured_var_decls.iterator() {
+                    tuple_var_name += var.name
+                    tuple_var_name += "__"
```
alimpfard, line 2176: Uhh...just add a fresh_var() or something?

adleroliveira: I could add a fresh var, for sure. The way I did was becuase it is easier to understand it in the generated c++ code:

```
Tuple<i64, String> const jakt__idx__word__ = idx__word__;
const i64 idx = ((jakt__idx__word__).get<0>());
 const String word = ((jakt__idx__word__).get<1>());
```

I can change it if it is really necessary

samples/tuples/destructuring_assignment_for.jakt
adleroliveira: I've changed the order to be more accurate

selfhost/parser.jakt
adleroliveira: I updated the error message to better represent the need

selfhost/parser.jakt
alimpfard: if that helped you or you think it helps, sure, let's keep it.

jntrnr: I believe this should be any pattern, rather than explicitly checking for `(`.

@alimpfard - for the matching, can we match on a pattern? I'm on a different machine, so it's a little tricky to check. But feels like we should allow the same pattern here we allow with `is` and `match`

alimpfard: 🤔 I mean if the match is guaranteed to work, why not?
But it does seem rather awkward, what should
```
enum Foo { A, B }
...
for A() in [Foo::A].iterator() {}
```
do here? 
what about
```
enum Foo { A }
...
for A() in [Foo::A].iterator() {}
```

adleroliveira: @jntrnr I would have to think better about how pattern matching in a `for` statement would work if it were for any pattern instead of just `(`. Maybe investigate how other languages (Rust maybe?) do it.

I would just go ahead and land pattern matching on enums and then build on that on future PRs.

What do you think?

alimpfard: fwiw currently the only things allowed after `let` are a name and an open paren for destructuring
I'd suggest putting that in a separate function, and using it here (because ultimately, this for loop binding desugars to a `let`).

Then we wouldn't have to answer these questions in the future, whatever `let` accepts, `for` accepts as well.

PR #1075, opened 2022-08-13
selfhost/main: Add -p argument from Rust compiler

SmellyJay:
Simply re-adds the '-p' argument from the Rust compiler, as a shorter way to prettify the CPP source.


jntrnr:
This was fixed as part of https://github.com/SerenityOS/jakt/pull/1077 I believe

PR #1078, opened 2022-08-14
Editors: Add package header and footer for jakt-mode

Walheimat:
Adds a package header and footer so that the mode can also be installed using a MELPA recipe.

Using `use-package` and `quelpa`, for example, you can then do this:

```elisp
(use-package jakt-mode
  :mode "\\.jakt\\'"
  :quelpa
  (jakt-mode
   :fetcher url
   :url "https://raw.githubusercontent.com/Walheimat/jakt/main/editors/emacs/jakt-mode.el"))
```

A complete header requires a version string but I wasn't sure what to set there.

### DIFF
--- a/editors/emacs/jakt-mode.el
+++ b/editors/emacs/jakt-mode.el
@@ -1,3 +1,18 @@
+;;; jakt-mode.el --- A major-mode for editing Jakt source code -*-lexical-binding: t-*-
+
+;; Url: https://github.com/SerenityOS/jakt
+;; Keywords: languages
+;; Package-Requires: ((emacs "25.1"))
+
+;; This file is distributed under the terms of the BSD 2-Clause
+;; "Simplified" license.
+
+;;; Commentary:
+
+;; This package implements a major-mode for editing Jakt source code.
+
+;;; Code:
+
 (defvar jakt-mode-syntax-table nil "Syntax table for `jakt-mode`.")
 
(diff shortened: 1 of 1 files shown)


jntrnr:
This looks like you're putting someone else's name on it? As best I can tell this person was not the original author.

In general, we discourage people from signing other people's names to files. You may want to reach out to the other authors to see if you can get them to sign their names to it.

Walheimat:
You're right, I was just going off the author GitHub was showing me.

@draxil  @veera-sivarajan  @lanmonster  @alimpfard  @0verse

Could you add your names? I removed the line for now, the version string as well (although it is necessary for a package install.

veera-sivarajan:
Hi @Walheimat, sure I'll add my name. Should I add `Author: $my_name` in the header?

draxil:
At this stage all I did was rename a file. I feel I'd only want to stick my name on if I do something more significant :)

Walheimat:
@veera-sivarajan yes, sorry, here's a link to the official docs: https://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers

Basically we want this:

```elisp
;; Author: Your Name <yourname@example.com>
;;      Someone Else <someone@example.com>
;;      Another Person <another@example.com>
```

@draxil That's cool. I actually I have some ideas to add but would be cool to have all the meta data first.

PR #1081, opened 2022-08-15
parser: Refactor to move equals methods into the subject struct/ enum

robryanx:


AtkinsSJ:
For that first commit, it's probably worth adding that as a test to make sure it doesn't break in the future.

robryanx:
> For that first commit, it's probably worth adding that as a test to make sure it doesn't break in the future.

Yeah, have added that now.

AtkinsSJ (approved):

PR #1088, opened 2022-08-18
Fix a number of fixmes

robryanx:


jntrnr:
@robryanx - does this fix any outstanding issues? If so, can we add unit test(s) to cover them?

robryanx:
> @robryanx - does this fix any outstanding issues? If so, can we add unit test(s) to cover them?

No issues, just some "FIXME"s in the code.
One commit is just replacing `0` with `builtin(BuiltinType::Void)`.
The other is filling in a bunch of missing type equality checks.

PR #1090, opened 2022-08-18
runtime: Use correct size bound for array slice iterator

### LINKED ISSUE #1089: array slice iterator() causes SIGILL on run
--- program
function main() {
    for c in ["one", "two", "three"][1..].iterator() {
        println("{}", c)
    }
}
--- gdb output
Starting program: /home/rich/Prog/jakt/array_slice
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/libthread_db.so.1".
two
three

Program received signal SIGILL, Illegal instruction.
0x0000555555578478 in JaktInternal::ArrayStorage<Jakt::String>::at(unsigned long)::{lambda()#1}::operator()() const ()
(gdb) bt
#0  0x0000555555578478 in JaktInternal::ArrayStorage<Jakt::String>::at(unsigned long)::{lambda()#1}::operator()() const ()
#1  0x00005555555783e8 in JaktInternal::ArrayStorage<Jakt::String>::at(unsigned long) ()
#2  0x000055555556d5dc in JaktInternal::ArrayIterator<Jakt::String>::next()
    ()
#3  0x0000555555561d2e in Jakt::main(JaktInternal::Array<Jakt::String>) ()
#4  0x0000555555561a36 in main ()

robryanx:
Previously when using a range with no size specified the iterator
wouldn't be correctly limited to the size of the slice.
Fixes #1089

### DIFF
--- a/runtime/Builtins/Array.h
+++ b/runtime/Builtins/Array.h
@@ -264,7 +264,7 @@ public:
 
     ArraySlice<T> slice_range(size_t from, size_t to) const
     {
-        return slice(from, to-from);
+        return slice(from, to - from);
     }
 
     ArraySlice<T> slice(size_t offset, size_t size) const;
@@ -347,6 +347,9 @@ public:
     {
         VERIFY(m_storage);
         VERIFY(m_offset < m_storage->size());
+
+        if (m_offset + m_size >= m_storage->size())
+            m_size = m_storage->size() - m_offset;
     }
 
     ErrorOr<Array<T>> to_array() const
@@ -356,13 +359,13 @@ public:
         for (size_t i = 0; i < size(); ++i) {
             TRY(array.push(at(i)));
         }
-            
+
         return array;
     }
 
     ArrayIterator<T> iterator() const
     {
-        return ArrayIterator<T> { *m_storage, m_offset, m_size };
+        return ArrayIterator<T> { *m_storage, m_offset, size() };
     }
 
     bool is_empty() const { return size() == 0; }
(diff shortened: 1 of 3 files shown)


runtime/Builtins/Array.h
```diff
@@ -362,7 +362,7 @@ class ArraySlice {
 
     ArrayIterator<T> iterator() const
     {
-        return ArrayIterator<T> { *m_storage, m_offset, m_size };
+        return ArrayIterator<T> { *m_storage, m_offset, size() };
```
jntrnr, line 365: is this because m_size doesn't encode that there is no limit to the range on that side?

robryanx: Yeah, I guess the larger question would be:
```
mut arr = [1, 2, 3, 4]
let slice = arr[..]
arr.push(5)
```
Should `slice` in that case be `[1, 2, 3, 4, 5]`
If yes I think it is correct after this fix.
If no, then it would be best to check the bounds when first setting `m_size`.

jntrnr: imho, slice should be viewing `[1, 2, 3, 4]`. I think of it as a view, and the indices get filled in when you create the view.

robryanx: Yeah, I was thinking about this some more and agree. I will update it.

robryanx: This is updated now.

runtime/Builtins/Array.h
```diff
@@ -347,6 +347,10 @@ class ArraySlice {
     {
         VERIFY(m_storage);
         VERIFY(m_offset < m_storage->size());
+
+        if(m_offset + m_size >= m_storage->size()) {
+            m_size = m_storage->size() - m_offset;
+        }
```
ADKaster, line 353: Do we not have a clang-format check in CI?

This should be
```suggestion
        if (m_offset + m_size >= m_storage->size())
            m_size = m_storage->size() - m_offset;
```

robryanx: Not currently in CI. I have updated to remove the braces/ fix the spacing.

PR #1091, opened 2022-08-18
typechecker: A couple of fixes for functions with function parameters

robryanx:


selfhost/typechecker.jakt
```diff
@@ -2215,6 +2216,51 @@ struct Typechecker {
...
+            Function(params: lhs_params, can_throw: lhs_can_throw, return_type_id: lhs_return_type_id) => {
+                guard rhs_type is Function(params: rhs_params, can_throw: lhs_can_throw, return_type_id: rhs_return_type_id) else {
+                    .error(
+                        format("Type mismatch: expected ‘{}’, but got ‘{}’", .type_name(lhs_type_id), .type_name(rhs_type_id))
+                        span
+                    )
+                    return false
+                }
+
+                if not (lhs_params.size() == rhs_params.size()) {
```
alimpfard, line 2217: Let's be nice and say where they differ a bit more precisely, something like "expected function... got function... (the types differ in argument number N, expected i32 got i64)"

selfhost/typechecker.jakt
```diff
@@ -2215,6 +2216,51 @@ struct Typechecker {
                     }
                 }
             }
+            Function(params: lhs_params, can_throw: lhs_can_throw, return_type_id: lhs_return_type_id) => {
```
alimpfard, line 2208: Also check if they can throw?

alimpfard (commented):
Also need to support it in substitute type vars.

selfhost/typechecker.jakt
alimpfard: Or maybe say that as a hint?

robryanx: Yeah, added as a hint, it is still using the generic name at the function level as `check_types_for_compat` just returns a bool:
`The parameter types differ at argument 2: expected ‘T’, but got ‘i32’`

There is also the error at the parameter level:
`Type mismatch: expected ‘i64’, but got ‘i32’`

The spans could also probably be improved a bit in future, as there are no spans passed in for the actual parameters.

selfhost/typechecker.jakt
robryanx: Yeah forgot that initially, added now.

robryanx:
> Also need to support it in substitute type vars.

Added, I think it is correct but not sure exactly how I test that case.

PR #1095, opened 2022-08-18
codegen+runtime: Fix pretty printing in exceptional paths 

MacDue:
This PR replaces the `_pretty_print_level++/--` and enabling/disabling the pretty printing with RAII helper classes.
This means that even if one of the many `TRY()`s in the `debug_description()` fails pretty printing will still return to a valid state.

(Updated bootstrap due to issues with now incompatible runtimes)


ADKaster:
Overall this seems like a reasonable change, but I'm a bit concerned by the +4k line diff in the bootstrap compiler. 

What impact does this change have on compile times?

alimpfard:
@ADKaster that's just the snapshot not having been updated for a while :sweat_smile:

MacDue:
Yep, this change alone actually will decrease the generated lines a little bit

PR #1096, opened 2022-08-19
main: Exit with non-zero exit if clang fails

jntrnr:
With this PR, if clang exits with a non-zero exit code, we'll exit jakt with a non-zero exit code. This will allow shell scripts, CI, etc to pick up on compiler errors that are coming from clang and not just directly from jakt.

### DIFF
--- a/selfhost/main.jakt
+++ b/selfhost/main.jakt
@@ -397,7 +397,13 @@ function main(args: [String]) {
         )
 
         if run_executable and compiler_status == 0 {
-            system(output_filename.c_string())
+            return system(output_filename.c_string())
+        } else {
+            if compiler_status == 0 {
+                return 0
+            } else {
+                return 1
+            }
         }
     }
 }


selfhost/main.jakt
```diff
@@ -397,7 +397,13 @@ function main(args: [String]) {
...
 
         if run_executable and compiler_status == 0 {
-            system(output_filename.c_string())
+            return system(output_filename.c_string())
+        } else {
+            if compiler_status == 0 {
+                return 0
+            } else {
+                return 1
+            }
```
alimpfard, line 406: 🤔 
```
        if ... {
            return system(...)
        }
        return compiler_status
```
?

jntrnr: Doesn't work, unfortunately. `compiler_status` gets set to -1 because of `system`, but returning this directly on macOS turns it into a `0` 🤷 . So I use an additional step to normalise to 0 or 1.

ADKaster: If we set this to a unique error status from jakt compiler error then the rationale for keeping run-one.sh in jakttest decreases 🤔

jntrnr: Correction `system` was returning "status: 256"

alimpfard: ah I see. okay then :shrug:

PR #1098, opened 2022-08-21
typechecker: Add type hint for match statement return type

robryanx:
Previously for an Optional match you would need
```
let b = match a {
    Bar => Some(2)
    Baz => {
        let ret: i64? = None

        yield ret
    }
}
```

With a type hint it can now be written as
```
let b: i64? = match a {
    Bar => 2
    Baz => None
}
```

Also adds numeric inference on optional types.

### DIFF
--- a/samples/match/match_optional_inference.jakt
+++ b/samples/match/match_optional_inference.jakt
@@ -1,0 +1,33 @@
+/// Expect:
+/// - output: "None\nNone\n"
+
+enum Foo {
+    Bar
+    Baz
+}
+
+function main() {
+    let a = Foo::Baz
+
+    let b = match a {
+        Bar => {
+            yield Some(2)
+        }
+        Baz => {
+            yield None
+        }
+    }
+
+    println("{}", b)
+
+    let c:i64? = match a {
+        Bar => {
+            yield 2
+        }
+        Baz => {
+            yield None
+        }
+    }
+
+    println("{}", c)
+}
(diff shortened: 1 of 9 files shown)


alimpfard:
:thinking: this should be allowed currently as well afaik?
```
let b = match a {
    Bar => Some(2)
    Baz => None
}
```

Propagating a type hint is nice though, +1 for making it work the expected way.

robryanx:
> thinking this should be allowed currently as well afaik?
> 
> ```
> let b = match a {
>     Bar => Some(2)
>     Baz => None
> }
> ```
> 
> Propagating a type hint is nice though, +1 for making it work the expected way.

Yeah interesting, it does.
What I was seeing was that it doesn't work when yielding from a block
```
let b = match a {
        Bar => Some(2)
        else => {
            yield None
        }
    }
```
I'll look into what the difference is there.

alimpfard:
Ah! Yield doesn't get the type hint as it's in a block, which doesn't accept one

jntrnr:
@alimpfard - is the fix here that the type hint should go into the block for a match so any `yield` in it can see it?

alimpfard:
I guess, we could pass it in as `yield_type_hint: TypeId?`

robryanx:
> I guess, we could pass it in as `yield_type_hint: TypeId?`

Yeah, looking into something like that to see if that works with everything else.

robryanx:
Updated now to include yield type hint.

PR #1100, opened 2022-08-21
editors/vscode: Small highlighting improvements

0GreenClover0:
This PR fixes "mut" highlighting in function argument lists (before only the "mut this" argument was highlighted) and adds highlighting for the "is" keyword.

Before:

obraz

After:

obraz

### DIFF
--- a/editors/vscode/syntaxes/jakt.tmLanguage.json
+++ b/editors/vscode/syntaxes/jakt.tmLanguage.json
@@ -786,7 +786,7 @@
       "patterns": [
         {
           "name": "keyword.operator.logic.jakt",
-          "match": "\\b(not|and|or)\\b"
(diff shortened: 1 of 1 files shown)


editors/vscode/syntaxes/jakt.tmLanguage.json
```diff
@@ -522,7 +522,7 @@
           "include": "#this"
         },
         {
-          "match": "\\b(mut)\\s+(this)\\b",
+          "match": "\\b(mut)\\b",
```
alimpfard, line 525: Capture 2 of this now points to the void, please remove that too (just below in captures.2)

0GreenClover0: Indeed, fixed

PR #1104, opened 2022-08-23
runtime: Add enumerate method on Array/ArraySlice

robryanx:
Just putting this out there. Maybe we want to do some bigger refactor of iterators later once we have traits.

Returns an iterator `ArrayIteratorIndexed` with a `next` function that
returns a `Tuple` of `index, item`

### DIFF
--- a/runtime/Builtins/Array.h
+++ b/runtime/Builtins/Array.h
@@ -3,6 +3,7 @@
 #include <Jakt/Checked.h>
 #include <Jakt/Error.h>
 #include <Jakt/RefCounted.h>
+#include <Jakt/Tuple.h>
 #include <Builtins/Range.h>
 #include <initializer_list>
 #include <stdlib.h>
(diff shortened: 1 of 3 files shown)


jntrnr:
Rather than adding this to the C++ side, can we make this functionality part of the prelude instead? Since it seems like it could be written in jakt rather than requiring C++.

Adding more to the C++ side makes it harder to merge AK back from the original Serenity one and since it's a better fit in jakt, probably the better choice.

robryanx:
We can, just not sure exactly where it would go?
The current prelude is basically just a list of signatures for things that exist in the runtime?

jntrnr:
pinging @alimpfard for ideas here. Seems like we might want the prelude to be a hybrid?

alimpfard:
:thinking: we currently don't codegen for the prelude at all, and I'm pretty sure it wouldn't even generate something that can be compiled if we did...

But it does seem sensible to keep the C++ library requirements as small as possible, so...I guess fixing that and using a hybrid prelude would be nice

robryanx:
Yeah, I am guessing once we have traits we would end up with wanting something like this in the prelude in the enumerate case (using the syntax from the traits PR):

```
trait Enumerable {
    function enumerate(this) -> Enumerator
}

struct Enumerator<T> {
    index: usize
    iterator: ArrayIterator<T>

    function next(mut this) -> (usize, T)? {
        .index++

        let item: T? = .iterator.next()
        if item.has_value() {
            let ret: (usize, T) = (.index - 1, item!)
            return Some(ret)
        } else {
            return None
        }
    }
}

type Array {
    implements(Enumerable) function enumerate(this) -> Enumerator {
        return Enumerator(
            index: 0
            iterator: .iterator()
        )
    }
}
```

robryanx:
Will close this and revisit once we have support for traits on prelude classes.

PR #1105, opened 2022-08-23
Everywhere: Changes unrelated to static reflection made while implementing static reflection

alimpfard:
Changes are mainly around evaluating, storing and calling lambdas.


alimpfard:
Looks like I've broken something about generics, drafting until that's resolved.

PR #1106, opened 2022-08-23
typecheker+codegen: Add basic support for ranges in match cases

adleroliveira:
```
function main() {
    let x = 42

    let y = match x {
        0..30 => "foo"
        30..100 => "bar"
        else => "baz"
    }

    println("{}", y)
}
```

### DIFF
--- a/samples/match/match_ranges.jakt
+++ b/samples/match/match_ranges.jakt
@@ -1,0 +1,14 @@
+/// Expect:
+/// - output: "bar\n"
+
+function main() {
+    let x = 42
+
+    let y = match x {
+        0..30 => "foo"
+        30..100 => "bar"
+        else => "baz"
+    }
+
+    println("{}", y)
+}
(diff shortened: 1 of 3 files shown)


alimpfard:
I'd prefer to hold out on this (+ generating specialised code for it) until we can implement it via traits...but if it helps simplify something in the meantime, we could have the specialised implementation

adleroliveira:
> I'd prefer to hold out on this (+ generating specialised code for it) until we can implement it via traits...but if it helps simplify something in the meantime, we could have the specialised implementation

I have no problem waiting for a trait based implementation, if it is agreed that is best to wait. 

I am implementing a Gameboy Emulator (Inspired by JT's idea of implementing a NES emu). And there is lots and lots of places where I use ranges + match cases. This is the main reason why I created this PR (make my life easier).

The alternative is that I myself can do a trait based implementation of this same feature once traits are available.

alimpfard:
We can definitely have the feature before traits are in, no problem.

PR #1107, opened 2022-08-23
typechecker: Report errors in more unreachable cases

### LINKED ISSUE #151: Return type inference doesn't reject incompatible returns

lanmonster:
We used to only check for reachability after "never return" statements.
Now, we check after every statement that can divert control flow.

fixes #151

### DIFF
--- a/selfhost/types.jakt
+++ b/selfhost/types.jakt
@@ -325,7 +325,6 @@ boxed enum Type {
                     return false
                 }
             }
-            return false
         }
     }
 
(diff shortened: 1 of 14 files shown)


selfhost/types.jakt
```diff
@@ -1038,6 +1046,7 @@ boxed enum CheckedExpression {
 
     function control_flow(this) -> BlockControlFlow => match this {
         Match(expr, match_cases, span, type_id, all_variants_constant) => {
+            // eprintln("control flow for expr: {}", expr)
```
AtkinsSJ, line 1049: This looks like it shouldn't be here. :^)

lanmonster: good catch!

PR #1110, opened 2022-08-25
codegen+runtime: Module resolution fixes/ basic formatters for Array/Set/Dictionary iterators

robryanx:


selfhost/typechecker.jakt
```diff
@@ -2719,7 +2719,7 @@ struct Typechecker {
                     is_instantiated: false
                     parsed_function: None
                     is_comptime: false
-                    is_closure: false
+                    is_closure: true
```
alimpfard, line 2722: Isn't this a function expression without any captures? that's not what this flag is for.
Maybe replace this field to be an enum?

robryanx: Ah, I have updated it to add to `FunctionType`
`Expression` and `Closure`
Let me know if that is okay, it seems that there are no `is_closure=true` cases currently so I am not sure that it is fine for that case.

jntrnr: pinging @alimpfard - does the above fix the issue you saw?

alimpfard: Looks right, sorry I forgot about this one :sweat_smile:

robryanx:
@jntrnr Just noticed this had conflicts, fixed now.

PR #1111, opened 2022-08-25
codegen: Expand match control flow resolution in-place

cybergsus:
Removes the `JAKT_RESOLVE_CONTROL_FLOW` macros and outputs their contents inside
codegen, so that we can have better line detection from clang in generated C++
code from matches.

Updates bootstrap since it removes an otherwise needed macro from `runtime`.


bootstrap/stage0/runtime/lib.h
```diff
@@ -280,42 +280,12 @@ struct ExplicitValueOrControlFlow {
         if constexpr (IsVoid<Value>)
             return;
         else
-            return move(value).template get<ExplicitValue<Value>>().value;
+            return move(move(value).template get<ExplicitValue<Value>>().value);
```
AtkinsSJ, line 283: I don't understand this `move()` you've added.

cybergsus: I haven't added that move. I haven't updated bootstrap manually, I just copied `runtime/lib.h` over.

cybergsus: Anyway, I've removed that change since it's irrelevant to the PR.

PR #1113, opened 2022-08-25
typechecker: Remove redundant errors that come from bad lookups

cybergsus:
Removes errors of the kind "no methods available on type", "assignment between
incompatible types", "attempt to access member on a non-struct type" etc. that propagate when some lookup yields "unknown" as the type of the expression. This avoids cluttering the error output and thus gives a clearer message to the user: "I got stuck at this point because I don't know what you're doing".

Failed lookups on functions now return a dummy expression that has "unknown" as the return type instead of "void". Failed lookups on method calls because of an unknown expression type id run the arguments through the typechecker but otherwise accept the call name and make it have "unknown" as a result.

Since the test proves that those redundant errors *don't* exist by requiring only one error to exist (the one lookup error that would've chained the redundant ones), I've modified Jakttest so it can check for this condition as well, under a `exact-error` flag after the `Expect: `.

### DIFF
--- a/jakttest/jakttest.jakt
+++ b/jakttest/jakttest.jakt
@@ -183,14 +183,41 @@ function compare_error(bytes: [u8], expected: String) throws -> bool {
     let builder_output = builder.to_string()
     let stripped_expected = strip_line_breaks(expected)
 
-    // println("builder output: {}", builder_output)
-    // println("stripped_expected: {}", stripped_expected)
-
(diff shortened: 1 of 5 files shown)


cybergsus:
I'm in the process of preventing error propagation on a for loop as well, if this is merged I'll submit another PR.

cybergsus:
OK, now it's ready.

PR #1114, opened 2022-08-26
all: Add support for basic class inheritance

jntrnr:
co-authored with @awesomekling

This adds simple form of class-based inheritance to get us started. For this first PR, the focus is on adding an inheritance hierarchy and calling methods in your parent class.

eg)
```
class Parent {
    public function greet(this) {
        println("I am the parent")
    }
}

class Child: Parent {
}

function main() {
    let child = Child()
    child.greet()
}
```

### DIFF
--- a/samples/classes/inheritance.jakt
+++ b/samples/classes/inheritance.jakt
@@ -1,0 +1,16 @@
+/// Expect:
+/// - output: "I am the parent\n"
+
+class Parent {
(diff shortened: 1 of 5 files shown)


AtkinsSJ:
Very very cool!

I'd suggest swapping the two commits around though, since the test will fail until the second commit.

PR #1117, opened 2022-08-27
typechecker: Add FunctionGenerics class to store specializations

robryanx:
This avoids the issue where currently infinite recursion will occur in
typechecking as typecheck_and_specialize_generic_function will be
recursively called.

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -695,10 +695,10 @@ struct CodeGenerator {
     function codegen_function_generic_parameters(mut this, function_: CheckedFunction) throws -> String {
         mut output = ""
 
-        if not function_.generic_params.is_empty() {
+        if not function_.generics.params.is_empty() {
             output += "template <"
             mut first = true
-            for generic_parameter in function_.generic_params.iterator() {
+            for generic_parameter in function_.generics.params.iterator() {
                 if first {
                     first = false
                 } else {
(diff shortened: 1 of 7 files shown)


jntrnr:
By once, do you mean only once for the given types? If you mean that, seems good. If you mean once and only once, regardless of types, then this wouldn't be correct.

robryanx:
> By once, do you mean only once for the given types? If you mean that, seems good. If you mean once and only once, regardless of types, then this wouldn't be correct.

Once in that it doesn't specialise again if it is already specialised in a parent scope.

We create a new child scope for each specialisation so the only case that wouldnt work with this if a generic function called itself again with different types, not sure if that is something we want to support?

robryanx:
So for example, from the test
```
function recurse<T>(anon i: T) -> T {
    if i > 100 {
        return i
    }

    return recurse(i + 1)
}

function main() {
    let a = recurse(1)
    let b = recurse('a')
}
```
This works.

If I change `recurse(i + 1)` to `recurse((i + 1) as! i32)` I will get
`Type mismatch: expected ‘i64’, but got ‘i32’`
As the parent specialisation of `recurse` that it is being called from was specialised to `i: i64`

So the question here would be, should that be allowed?
If not this works fine.
If so it probably needs a fairly different solution that instead holds a list of specialisations that have already been typechecked.

alimpfard:
other specialisations should also be valid (and checked), so this should go with a specialisation cache instead of forcing all instantiations of that generic entity to be the currently instantiated one.

robryanx:
Sure. I will have a look into that approach.

robryanx:
@alimpfard Updated this now to store the specialisations.

alimpfard (approved):

PR #1118, opened 2022-08-27
typechecker: Allow calling a function passed by reference

robryanx:
Previously when resolving a call we didn't check to see if we needed to
dereference a variable in scope before checking if it is a function.

This also adds checking for references in `check_types_for_compat`

### DIFF
--- a/tests/typechecker/function_pass_by_reference.jakt
+++ b/tests/typechecker/function_pass_by_reference.jakt
@@ -1,0 +1,16 @@
+/// Expect:
+/// - output: "2\n"
+
+function add(i: i64, adder: &function(anon i: i64) -> i64) -> i64  {
+    return adder(i)
+}
+
+function main() {
+    let adder = function(anon i: i64) -> i64 {
+        return i + 1
+    }
+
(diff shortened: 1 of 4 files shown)


tests/typechecker/function_pass_by_reference.jakt
```diff
@@ -0,0 +1,16 @@
...
+function add(i: i64, adder: &function(anon i: i64) -> i64) -> i64  {
+    return adder(i)
+}
+
+function main() {
+    let adder = function(anon i: i64) -> i64 {
+        return i + 1
+    }
+
+    let a = add(i: 1, adder: &adder)
```
jntrnr, line 13: now that we have https://github.com/SerenityOS/jakt/pull/1115, can this be `let a = add(i: 1, &adder)`?

robryanx: Yeah, I have updated that now.

robryanx: Just to note I have also updated the PR to allow for checking for reference types in `check_types_for_compat`

PR #1127, opened 2022-08-29
typechecker+codegen+tests: Accept never as a branch result

cybergsus:
This affects matches specially but another test is added to make sure
that assigning never is still reported as an error since it doesn't make
sense to store something you're not getting.

EDIT: cc @alimpfard


alimpfard:
I'm gonna copypaste what I said on discord:
> imo `=> { panic() }` makes it clear that this branch isn't yielding values, unlike `=> panic()` which sounds like it's evaluating to the value of that call

But if this sounds like a good idea to @jntrnr & @awesomekling, I wouldn't mind having it

PR #1129, opened 2022-08-30
jakttest: Better output for test failure cases

robryanx:
When creating a test to expect "error:" and no error occurs the test
result stage will be later than the expected stage.

### DIFF
--- a/jakttest/jakttest.jakt
+++ b/jakttest/jakttest.jakt
@@ -345,15 +345,24 @@ enum TestStage {
         TestRun => "Test binary run"
     }
 
-    function from_exit_code(exit_code: i32) => match exit_code {
-        0i32 => Some(TestStage::TestRun)
-        1i32 => Some(TestStage::TestRun)
-        2i32 => Some(TestStage::CompileCpp)
-        3i32 => Some(TestStage::TranspileJakt)
-        else => {
-            let nothing: TestStage? = None
-            yield nothing
-        }
+    function to_order(this) => match this {
+        TranspileJakt => 1
+        CompileCpp => 2
+        TestRun => 3
+    }
+
+    function output_filenames(this) => match this {
+        TranspileJakt => ("", "compile_jakt.err")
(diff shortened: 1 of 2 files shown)


jakttest/jakttest.jakt
```diff
@@ -727,6 +734,11 @@ function main(args: [String]) {
                     output +=  failed_stage + " failed with output:\n"
                     output += error
                 }
+                ErroredAtLaterStage(failed_stage, error) => {
+                    output += "Test failed at an later stage than expected:\n"
```
AtkinsSJ, line 738:
```suggestion
                    output += "Test failed at a later stage than expected:\n"
```

robryanx: Thanks, fixed that.

jntrnr:
@robryanx - is there a test case we could add to test this?

robryanx:
Not really I guess, as it would be testing jakttest itself.

I am going to look into changing it a bit as while this works the error is still a bit unhelpful.
I think what it really needs is something along the lines of `Test was expecting an error but succeeded with output: `

robryanx:
@jntrnr Okay I have updated this to be a bit broader.
Some test examples which would show new behavior:
```
/// Expect:
/// - error: "Error here\n"

function main() {
    println("Output success")
}
```
would give
```
|	Test successfully executed when error was expected:
|	Expected Error:
|	Error here
|	Got Output:
|	Output success
```

```
/// Expect:
/// - output: "Test\n"

function main() {
    mut arr = [1, 2, 3, 4]
    mut slice = arr[1..2]
    mut slice_2 = slice[1..1]
}
```
would give
```
|	Test failed at an earlier stage than expected:
|	Clang++ compilation of generated C++ source failed with output:
|       {long cpp error here}
|	Test binary run expected with success output:
|	Test
```

or for later instead of earlier the same test as above but changing the expectation to `/// - error: "Test\n"` which would be an error expected at `TranspileJakt` before `CompileCpp`

PR #1133, opened 2022-08-31
typechecker: Don't infer return type for function blocks

lanmonster:
Not specifying a return type for a function block results in that block
returning void. We will still infer the return type for fat arrow
functions.


jntrnr:
I'm surprised you didn't have to change more test cases for this one. I guess maybe we don't use return type inference very much?

PR #1135, opened 2022-08-31
typechecker: Make checking calls more fine-grained

cybergsus:
Instead of relying on a single resolve_call and typecheck_call to do all
the job, I've split them so that functions can be resolved differently
depending on the context (method call vs. regular call). This will make
space for traits as well where, similar to the superclass situation, it
has to lookup in multiple scopes to find the requested method.

This is what I've done, but first let's have some context:

`typecheck_call` does three main things:
- (1) Resolve the parent scope to resolve the function that's being called if
     we have a parent structure as the parent scope.
- (2.1) if the function is `eprintln` or `print` or `println` or `eprintln`, then
  we take a custom path where we just run the typechecker over the
  arguments.
- (2.2) if the function is `format`, then we do the same as with the printing
  function AND we make sure that we're catching the exception (since
  `String::formatted` can throw in the runtime).
- (2.3) otherwise, call `resolve_call` to get the function ID for the user
  function being called:

    - (2.4) if we don't get a function ID back, we just visit the arguments to
      check them and make the function return `void` (really its return type
      is unknown, which I've substituted).

- (4) if we DO get a function ID back:
    - instantiate generics
    - typecheck the arguments against the argument labels
    - ... more things might apply here, since I just copypasted the parts I
      was interested in.
    - set the return type of the function to the resolved return type of the
      (instantiated) function.

- (5) if we had a function ID, and the function is marked `comptime`, and our
  current function is not `comptime`, then we'll run the interpreter through
  the function call. If the interpreter had an error then we'll return a
  call like we'd do if we hadn't run the interpreter at all.

- (6) typecheckers for both regular calls and method calls call into
  `typecheck_call`.

What I've done:

- (4) is now the method `typecheck_resolved_call`, that accepts a
  **known** function ID.
- (5) is now the method `execute_call_if_comptime`, that is only called
  when the function ID is known.
- (2.1) and (2.2) are now the method `detect_and_typecheck_builtin_call`, which is
  called only when typechecking a call expression, and returns `None` when it's
  not calling a builtin function.
- (2.4) is now the method `typecheck_unresolved_call`.
- (6) dissapears. Calls are handled by `typecheck_call_expr` and
  `typecheck_method_call_expr` differently:
  - (1) is done by `typecheck_method_call_expr` without calling `resolve_call`. `typecheck_call` uses the old untouched `resolve_call`. `typecheck_method_call_expr` uses `resolve_method_call`, where we already know that we have an enum/struct.
  - only `typecheck_call_expr` calls `try_typecheck_builtin_call`, since a
    method call can't be to `eprintln`.

- Both `typecheck_method_call_expr` and `typecheck_call_expr` now only call
  `execute_call_if_comptime` when they have a function ID, since the rest of the
  times there's no chance it's going to do something.

- `typecheck_call` dissapears, since it's been divided in multiple chunks that
  are used or not depending on the context.


selfhost/typechecker.jakt
```diff
@@ -4184,11 +4184,47 @@ struct Typechecker {
             yield CheckedExpression::QuotedString(val, span)
         }
         Call(call, span) => {
-            yield .typecheck_call(call, caller_scope_id: scope_id, span, this_expr: None, parent_id: None, safety_mode, type_hint, must_be_enum_constructor: false)
+            let maybe_builtin_call = .try_typecheck_builtin_call(call, span, scope_id, safety_mode, type_hint)
```
jntrnr, line 4187: this is the wrong direction as it's making the containing function much larger. If you want to refactor, you'll want to take this logic and move it into its own method, and have that method call into the steps you need.

cybergsus: sure. :)

cybergsus: I've also removed the optional arguments, it was not adding any value.

cybergsus: And did the same thing with the method call.

selfhost/typechecker.jakt
```diff
@@ -6061,148 +5991,296 @@ struct Typechecker {
...
+            return .checked_call_to_method_call(
+                expr: checked_expr
+                checked_call: .typecheck_unresolved_call(call, scope_id, safety_mode)
+                is_optional, span)
+        }
+
+        let callee = .get_function(function_id!)
+
+        let checked_call = .typecheck_resolved_call(
+            call, caller_scope_id: scope_id, span
```
robryanx, line 6191: In cases where arguments are being split to multiple lines I think one argument per line is best.

selfhost/typechecker.jakt
```diff
@@ -6061,148 +5991,296 @@ struct Typechecker {
...
+
+        return .execute_call_if_comptime(
+            checked_call, span,
+            caller_scope_id: scope_id
+            resolved_namespaces: []
+            this_expr: checked_expr
+        ) ?? .checked_call_to_method_call(expr: checked_expr, checked_call, is_optional, span)
+    }
+
+    function checked_call_to_method_call(mut this, expr: CheckedExpression, checked_call: CheckedCall, is_optional: bool, span: Span) throws -> CheckedExpression =>
```
robryanx, line 6215: Arguably for readability multiline arrow function like this would be better as a regular function.

selfhost/typechecker.jakt
```diff
@@ -6314,3 +6392,9 @@ struct Typechecker {
         return false
     }
 }
+
+function checked_call_to_call_expr(checked_call: CheckedCall, span: Span) throws -> CheckedExpression =>
```
robryanx, line 6396: Same here and no reason I think this can't be a method like `checked_call_to_method_call`

robryanx (commented):
This all looks fine, just added a few minor formatting comments.

selfhost/typechecker.jakt
cybergsus: that one isn't a method because there was no dependency on the typechecker, and I'm used to having helper "filling" functions be free functions, but can make it a static method if that would really help.

cybergsus:
@robryanx I've been fixing your formatting nitpicks. The second one about converting it to a "regular" function, all it did was add noise with all the braces and the `return` keyword. I'll push in the changes hoping this PR gets merged soon, since without this I cannot continue working on traits.

cybergsus:
after rebasing turns out that the `guard` that I introduced conflicts with the current algorithm for checking builtins. Since this PR is not supposed to change the algorithm, I removed that extra check.

selfhost/typechecker.jakt
```diff
@@ -4463,77 +4463,8 @@ struct Typechecker {
...
-                        is_optional
-                        type_id: result_type)
-                }
-                else => {
-                    .compiler.panic("typecheck_call should return `CheckedExpression::Call()`")
-                }
-            }
-        }
+        Call(call, span) => .typecheck_call_expr(call, span, must_be_enum_constructor: false, scope_id, safety_mode, type_hint)
+        MethodCall(expr, call, span, is_optional) => .typecheck_method_call_expr(expr, call, span, is_optional, scope_id, safety_mode, type_hint)
```
maddanio, line 4467: is "call" supposed to be a superset of "method_call"? from the naming I would guess that, but given they are called from distinct enum cases contradicts that

cybergsus: No, "call" is a qualified call (e.g `foo()`), methodcall is the one with a dot `foo.bar()`. They have different ways of resolving, hence this refactor.

PR #1141, opened 2022-09-02
Don't infer return type in function/method blocks

### LINKED ISSUE #151: Return type inference doesn't reject incompatible returns

### LINKED ISSUE #483: Returning enum named `c_int` causes C++ error

### LINKED ISSUE #1092: Return type not correctly inferred when returning newly constructed instance
To reproduce:
```
class C {

    public function creat() throws {

        let s = true

        return C()
    }

}

function main() {
    println("{}", C::creat())
}
```
When compiling, the following error from the C++ is  produced:
```
error: no viable conversion from returned value of type 'Jakt::NonnullRefPtr<Jakt::C>' to function return type 'ErrorOr<void>'
return (TRY((C::create())));
       ^~~~~~~~~~~~~~~~~~~~
```
Workaround: The error goes away and the `main()` function works are expected if the `creat()` function is changed to explicitly state the return type:
```
    public function creat() throws -> C {
         ...
    }
```

### LINKED ISSUE #1093: Function return type inference changes when statement is removed from block
To reproduce:
```
class C {

    public function creat() throws {

        let s = true

        return C()
    }

}

function main() {
    println("{}", C::creat())
}
```
Error message:
```
error: no viable conversion from returned value of type 'Jakt::NonnullRefPtr<Jakt::C>' to function return type 'ErrorOr<void>'
return (TRY((C::create())));
       ^~~~~~~~~~~~~~~~~~~~
```

The problem goes away if the `match` statement is removed as follows:
```
class C {
    public function creat() throws {
        return C()
    }
}

function main() {
    println("{}", C::creat())
}
```

0GreenClover0:
A spiritual ancestor of #748 and #215. Fixes #151 in case incompatible return types are actually reachable (which seemed to be the issue, given the title). #1133 follow-up.

Resolves #1092 and resolves #1093 because we no longer infer the return type for function blocks. Fixes #483.

Not sure about the generic stuff, please check :^)

I don't know how to put a commit description without commit linter complaining, so I'll just paste it here:
> This refers to functions as well as methods that are not fat-arrow.
> Parser and a couple of tests needed to be updated in some cases, to specify the return type.
> When the return type is not specified for not-fat-arrow functions, it is set to void.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -1729,18 +1729,31 @@ struct Typechecker {
         .current_function_id = method_id
 
         let VOID_TYPE_ID = builtin(BuiltinType::Void)
+        mut function_return_type_id = .typecheck_typename(
+            parsed_type: func.return_type,
+            scope_id: function_scope_id,
+            name: None
+        )
 
-        let block = .typecheck_block(parsed_block: func.block, parent_scope_id: function_scope_id, safety_mode: SafetyMode::Safe)
-        let function_return_type_id = .typecheck_typename(parsed_type: func.return_type, scope_id: function_scope_id, name: None)
+        if not func.is_fat_arrow and func.return_type is Empty {
+             function_return_type_id = VOID_TYPE_ID
+        }
+
+        checked_function.return_type_id = function_return_type_id
+
+        let block = .typecheck_block(
+            parsed_block: func.block,
+            parent_scope_id: function_scope_id,
+            safety_mode: SafetyMode::Safe
+        )
 
         mut return_type_id = function_return_type_id
         if function_return_type_id.equals(unknown_type_id()) and not block.statements.is_empty() {
-            // If the return type is unknown, and the function starts with a return statement,
+            // If the return type is unknown and function is a fat arrow function,
             // we infer the return type from its expression.
-            if block.statements[0] is Return(val) and val.has_value() {
-                return_type_id = val!.type()
-            } else {
-                return_type_id = VOID_TYPE_ID
+            // The only statement in fat arrow function block.statements is a Return statement.
+            if func.is_fat_arrow {
+                return_type_id = .infer_function_return_type(block)
             }
         } else if function_return_type_id.equals(unknown_type_id()) {
             return_type_id = VOID_TYPE_ID
@@ -1929,9 +1942,16 @@ struct Typechecker {
             )
             .ignore_errors = old_ignore_errors
 
-            let return_type_id = match function_return_type_id.equals(unknown_type_id()) {
-                true => .infer_function_return_type(block)
-                else => .resolve_type_var(type_var_type_id: function_return_type_id, scope_id: parent_scope_id)
+            mut return_type_id = function_return_type_id
+            if function_return_type_id.equals(unknown_type_id()) {
+                // If the return type is unknown and function is a fat arrow function,
+                // we infer the return type from its expression.
+                // The only statement in fat arrow function block.statements is a Return statement.
+                if parsed_function.is_fat_arrow {
+                    return_type_id = .infer_function_return_type(block)
+                }
+            } else {
+                return_type_id = .resolve_type_var(type_var_type_id: function_return_type_id, scope_id: parent_scope_id)
             }
 
             checked_function.block = block
@@ -2107,14 +2127,15 @@ struct Typechecker {
             name: None
         )
 
-        // Infer return type if necessary
-        // If the return type is unknown, and the function starts with a return statement,
-        // we infer the return type from its expression.
-        let UNKNOWN_TYPE_ID = unknown_type_id()
         let VOID_TYPE_ID = void_type_id()
         mut return_type_id = VOID_TYPE_ID
-        if function_return_type_id.equals(UNKNOWN_TYPE_ID) {
-            return_type_id = .infer_function_return_type(block)
+        if function_return_type_id.equals(unknown_type_id()) {
+            // If the return type is unknown and function is a fat arrow function,
+            // we infer the return type from its expression.
+            // The only statement in fat arrow function block.statements is a Return statement.
+            if parsed_function.is_fat_arrow {
+                return_type_id = .infer_function_return_type(block)
+            }
         } else {
             return_type_id = .resolve_type_var(
                 type_var_type_id: function_return_type_id,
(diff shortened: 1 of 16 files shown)


0GreenClover0:
Might have fixed that for methods too. I'll test a little bit and push when I'm done

selfhost/typechecker.jakt
```diff
@@ -1729,18 +1729,34 @@ struct Typechecker {
...
+        let block = .typecheck_block(
+            parsed_block: func.block,
+            parent_scope_id: function_scope_id,
+            safety_mode: SafetyMode::Safe
+        )
 
         mut return_type_id = function_return_type_id
         if function_return_type_id.equals(unknown_type_id()) and not block.statements.is_empty() {
-            // If the return type is unknown, and the function starts with a return statement,
+            // If the return type is unknown, function is a fat arrow function and it starts with a return statement,
```
jntrnr, line 1752: I'm a little confused with this statement. If a function is a fat arrow, then the thing it's pointing to should be the value it's returned.

Is this talking about the case where there are additional returns inside of the fat arrow?

tests/typechecker/auto_deref.jakt
```diff
@@ -47,8 +47,8 @@ function deref_match(anon foo: &i32) -> i32 {
 }
 
 struct Foo {
-    function const_method(this) => 69
-    function mutable_method(mut this) => 69 + 1
+    function const_method(this) -> i32 => 69i32
```
jntrnr, line 50: I think we should be able to infer the type of fat arrow functions, as we're immediately returning what they point to

0GreenClover0: The inferred type here is i64 but the function was used as it was i32 in deref_access function, which returns i32. Before we would just silently accept the overflow and wrap the number. Not sure what the correct behavior is...?

```cpp
struct Foo {
    function const_method(this) -> i32 => 69i32
    function mutable_method(mut this) -> i32 => 69i32 + 1i32
}

function deref_access(anon foo: &mut Foo) -> i32 {
    return foo.const_method() + foo.mutable_method()
}
```

0GreenClover0: Actually I was wrong, after my changes the overflow of i32 in checked addition here would cause a runtime panic.

selfhost/typechecker.jakt
0GreenClover0: I've changed the comment, added a copy of it to the typecheck_function and unified the logic of that snippet between the typecheck_method and typecheck_function

0GreenClover0: Also: unified this logic for functions with generic parameters

tests/typechecker/auto_deref.jakt
0GreenClover0: Unless, you meant that we only need to say `=> 69i32` instead of `-> i32 => 69i32`, then yes you are right 😅 I wrote the return type out of habit

jntrnr: Right, the last one. It should just be `=> 69i32` and infer `i32` as the return type.

0GreenClover0:
@jntrnr I've added some comments and deleted some debug variables / checks that I forgot about (oops...). Sorry for the mess, it must have been a pain to understand the changes.

I've also added one fixme, because even after this PR we still don't check properly the type of make_memo() in
`/tests/typechecker/generic_across_scopes.jakt`. The tests itself is unfortunately way above my head and I don't know what the exact problem is.
If the PR gets merged I'll probably add an issue for that

alimpfard:
> we still don't check properly the type of make_memo()

That has an explicit return type, and is correctly typechecked on main - what fails?

selfhost/typechecker.jakt
```diff
@@ -3759,20 +3780,65 @@ struct Typechecker {
...
+        or type_hint.value().equals(unknown_type_id())
+        or .get_type(checked_expr.type()) is TypeVariable
+        // FIXME: Not sure how, but we should deal with GenericInstance differently.
+        // Ex. in tests/typechecker/generic_across_scopes.jakt we don't check the return type of make_memo properly
+        or .get_type(type_hint!) is GenericInstance {
+            // We don't yet exactly know the type of what we need to return / type of what are we returning
+            return CheckedStatement::Return(val: checked_expr, span)
+        }
+
+        if not type_hint.value().equals(checked_expr.type()) {
```
alimpfard, line 3833: Don't do this, please never use `TypeId::equals` to check for type compatibility - there's a function for that, it even logs the errors for you.

0GreenClover0:
> > we still don't check properly the type of make_memo()
> 
> That has an explicit return type, and is correctly typechecked on main - what fails?

Nah, it doesn't fail the test itself, but when you change
```cpp
function make_memo<R, S>() throws -> Memo<R, S> {
    let value: [R:S] = [:]
    return Memo(value)
}
```

to... for example this:

```cpp
function make_memo<R, S>() throws -> Memo<R, S> {
    let value: [R:S] = [:]
    return ""
}
```

It doesn't complain and generates invalid cpp code (even on main)

selfhost/typechecker.jakt
0GreenClover0: I can see why :P
After I changed it to `check_types_for_compat`, the early return in case line 3828 `or .get_type(type_hint!) is GenericInstance` is true, is no longer needed. Unfortunately it doesn't solve the problem of make_memo so the FIXME belongs somewhere else (maybe TypeVariable?).

This allows us to properly check the return type of the function "find" in `samples/closures/generic_closure_with_parameter.jakt`

```cpp
function find<T>(anon arr: [T], anon cb: function(a: T) -> bool) -> T? {
    for i in 0..arr.size() {
        if cb(a: arr[i]) {
            return i
        }
    }
    return None
}

function main() {
    mut arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
    
    println("{}", find(arr, function(a: i64) -> bool => a == 5).value_or(-1))
}
```

`return i` gives an error now, because we try to return `usize` instead of `i64?`. I'll change it to `return i as! i64` and add a test case for the error.

0GreenClover0:
Pushed up the changes. If I remove the `or .get_type(checked_expr.type()) is TypeVariable` on line 3824 in the early return, the only failing test now is generic_across_scopes.jakt with the error

```Error: Type mismatch: expected ‘i64’, but got ‘V’
----- ./tests/typechecker/generic_across_scopes.jakt:28:13
 27 |         if memo!.value.contains(x) { 
 28 |             return memo!.value[x] 
                  ^- Type mismatch: expected ‘i64’, but got ‘V’
 29 |         } 
-----
```

0GreenClover0:
It might not even be related to the `is TypeVariable` early return
Am I dumb or this error is getting swallowed? This does not print any Jakt errors and gets past the codegen stage, where invalid C++ is emitted

obraz

0GreenClover0:
You learn something everyday
obraz

Not sure if this is the way, but I set ignore_errors to `false` before checking for type compatibility in `typecheck_return()` and after the check I restore its value. Thanks to this, we properly error when the return type of `make_memo` is wrong.
I've added a test case for it as well.

I don't think there are any other known issues.

alimpfard:
The first pass over generic functions _should_ ignore errors - that's expected
The second specialised pass (on call) should not be ignoring it though

0GreenClover0:
So is there a bug somewhere? It seems to be set to true both times we call typecheck_return for make_memo

0GreenClover0:
There seems to be a bug for sure.

```cpp
function make_memo<R, S>() throws -> Memo<R, S> {
    let x: String = 32 
    let value: [R:S] = [:]
    return Memo(value)
}
```

Assigning i64 to String gives us...

root@CloverLaptop:/home/jakt# ./build/jakt ./tests/typechecker/generic_across_scopes.jakt 
internal error: Garbage statement in codegen
Aborted

0GreenClover0:
We don't really typecheck functions like make_memo at all currently, so I'll remove the hack with setting `.ignore_errors = false`, I don't think it would work in all cases regardless.

PR #1145, opened 2022-09-03
selfhost: Set default runtime path based on location of jakt binary 

ADKaster:
Now we look in current_executable_directory/../include/runtime for the
runtime files instead of $PWD/runtime by default. This makes it so that
one doesn't have to pass -R in most cases. Windows is still a bit iffy
because the FilePath utility doesn't understand that Windows paths have
backslashes instead of forward slashes.

At the same time, mess with the structure of the build directory a bit to match the structure of the install tree.

This means the binaries are in build/bin and the runtime is in
./build/include/runtime. Any future libraries we create will be dropped
in ./build/lib. This enables future work to let the jakt compiler find
its resources automagically from the current executable's location.


CMakeLists.txt
```diff
@@ -25,6 +25,19 @@ else()
...
+  set_property(TARGET ${target} PROPERTY "ARCHIVE_OUTPUT_DIRECTORY" "${CMAKE_CURRENT_BINARY_DIR}/lib")
+  set_property(TARGET ${target} PROPERTY "LIBRARY_OUTPUT_DIRECTORY" "${CMAKE_CURRENT_BINARY_DIR}/lib")
+  set_property(TARGET ${target} PROPERTY "RUNTIME_OUTPUT_DIRECTORY" "${CMAKE_CURRENT_BINARY_DIR}/bin")
+  foreach (build_config ${CMAKE_CONFIGURATION_TYPES})
+    string(TOUPPER ${build_config} CFG_TYPE)
+    set_property(TARGET ${target} PROPERTY "ARCHIVE_OUTPUT_DIRECTORY_${CFG_TYPE}" "${CMAKE_CURRENT_BINARY_DIR}/lib")
+    set_property(TARGET ${target} PROPERTY "LIBRARY_OUTPUT_DIRECTORY_${CFG_TYPE}" "${CMAKE_CURRENT_BINARY_DIR}/lib")
+    set_property(TARGET ${target} PROPERTY "RUNTIME_OUTPUT_DIRECTORY_${CFG_TYPE}" "${CMAKE_CURRENT_BINARY_DIR}/bin")
+  endforeach()
+  set_property(TARGET ${target} PROPERTY "BUILD_INCLUDE_DIRECTORY" "${CMAKE_CURRENT_BINARY_DIR}/include")
```
ADKaster, line 38: @friendlyanon Is there a reasonable way to do this in a way that doesn't destroy multi-config generators?

What I was hoping to create with this change was to have it so that whether a developer invokes the jakt compiler via `jakt my_file.jakt` from the runtime directory or an installed location, they can expect it to find any built-in includes or runtime libraries.

Using both the change tagged here and the `copy_include_to_build.cmake` file generation below the goal is, for single config:
```
build/
   bin/
      jakt
      jakttest
      etc
   include/
      runtime/
         All the runtime headers needed to compile a jakt program
   lib/
     future location of std lib, runtime lib, etc
```
And then for each config of a multi-config generator, it would have 
```
build/
   Debug/
      bin/
      include/
      lib/
   Release/
      bin/
      include/
      lib/
```

I looked around and it seems like LLVM uses `${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/bin`
https://github.com/llvm/llvm-project/blob/a46154cb1cd09aa26bc803d8696e6e9283aac6a9/llvm/CMakeLists.txt#L383-L398

But doing that makes `Ninja Multi-Config` very angry. I assume it makes MsBuild less angry, though I haven't tried it yet on my Windows PC with VS 2022.

What tripped me up with just setting the [non-config-suffixed properies](https://cmake.org/cmake/help/latest/prop_tgt/RUNTIME_OUTPUT_DIRECTORY.html) is that now Ninja Multi-Config puts the files in
`build/bin/Debug` or, sometimes, `build/bin/Debug/bin` which is super not what I wanted. I want the application to find its resources via the equivalent of `$(dirname $(dirname <current exe path>))/include` by default, while keeping the switches there for people with weird setups.

Trying to use things like `$<CONFIG>` in the file copy for the include files also tripped me up. I couldn't figure out how to get it to create the structure above using Ninja Multi-Config, in a way that re-copies any runtime/ headers into a per-config include directory when they change. The fact that `$<CONFIG>` expands to a lower-cased name makes it super confusing how you're supposed to use it to extract these properties back out later...

friendlyanon: I would just do this:

```cmake
  set(prefix "")
  get_cmake_property(is_multi_config GENERATOR_IS_MULTI_CONFIG)
  if(is_multi_config)
    set(prefix "$<CONFIG>/")
  endif()

  set_property(TARGET "${target}" PROPERTY ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/${prefix}lib")
  set_property(TARGET "${target}" PROPERTY LIBRARY_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/${prefix}lib")
  set_property(TARGET "${target}" PROPERTY RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/${prefix}bin")
```

---

> What tripped me up with just setting the [non-config-suffixed properies](https://cmake.org/cmake/help/latest/prop_tgt/RUNTIME_OUTPUT_DIRECTORY.html) is thatnow Ninja Multi-Config puts the files in `build/bin/Debug` or, sometimes, `build/bin/Debug/bin` which is super not what I wanted.

To quote the link:

> Multi-configuration generators ([Visual Studio](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#visual-studio-generators), [Xcode](https://cmake.org/cmake/help/latest/generator/Xcode.html#generator:Xcode), [Ninja Multi-Config](https://cmake.org/cmake/help/latest/generator/Ninja%20Multi-Config.html#generator:Ninja%20Multi-Config)) append a per-configuration subdirectory to the specified directory unless a generator expression is used.

and `$<CONFIG>` indeed prevents that from happening.

---

I got nothing for the header situation. Maybe try [`cmake -E create_symlink`](https://cmake.org/cmake/help/latest/manual/cmake.1.html#:~:text=multiple%20input%20files.-,create_symlink,-%3Cold%3E%20%3Cnew%3E)?

ADKaster: aha, I was indeed over-thinking the heck out of it. The hint for using create_symlink for the headers seems much easier than what I was trying to do before with copying whenever anything changes. Thanks!

CMakeLists.txt
```diff
@@ -68,14 +85,30 @@ if (FINAL_STAGE GREATER_EQUAL 2)
...
+    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/runtime>
+  )
   add_executable(Jakt::jakt_stage2 ALIAS jakt_stage2)
+  apply_output_rules(jakt_stage2)
+endif()
+
+# Link runtime into build directory(ies) for relative pathing usage
+#    Note: "If a sequential execution of multiple commands is required, use multiple execute_process() calls with a single COMMAND argument."
+if (CMAKE_CONFIGURATION_TYPES)
+  foreach (build_type IN LISTS CMAKE_CONFIGURATION_TYPES)
```
friendlyanon, line 99: A `POST_BUILD` command might be better here imo, so the things are linked when necessary. The presence of `CMAKE_CONFIGURATION_TYPES` is not the determining factor for a multi config generator.

alimpfard:
Moving the commits into #1187 to work around the silly `#include <File.cpp>` in `lib.h` which is causing windows CI to fail.

alimpfard:
Now part of #1187.

PR #1149, opened 2022-09-03
runtime: Make runtime a separately compiled static library

ADKaster:
This prevents us from compiling all of Format, String, File, etc along with every single Jakt executable.
  
  Depends on #1145


ADKaster:
OBE by https://github.com/SerenityOS/jakt/pull/1187

PR #1161, opened 2022-09-07
all: Allow destructuring into existing variables/ from nested Tuples

robryanx:
If destructure is not preceded by `let`/`mut` existing variables will be
destructured into rather than new variables be declared.

Allow destructuring recursively for nested Tuples.

Introduced `ParsedExpression::Destructure` to be used for the 
destructure pattern to allow easier pattern matching destructures on 
enums in a future PR. Currently it takes the same form as 
`ParsedExpression::JaktTuple`.

Updated naming scheme for the Tuple being destructured to work like
anonymous lambdas to avoid naming ambiguity when the same variables are 
destructured into multiple times.

### DIFF
--- a/samples/tuples/destructuring_assignment.jakt
+++ b/samples/tuples/destructuring_assignment.jakt
@@ -1,11 +1,42 @@
 /// Expect:
-/// - output: "1 hello\n"
+/// - output: "1 hello\n1 hello\n2, 3, 4\nTest(a: 1, b: 2, c: 3)\n"
 
 function foo() -> (i64, String) {
     return (1, "hello")
 }
 
+struct Test {
+    a: i64
+    b: i64
+    c: i64
+}
+
 function main() {
     let (x1, x2) = foo()
+
     println("{} {}", x1, x2)
+
+    mut bar = 0
+    mut baz = ""
+
+    (bar, baz) = foo()
+
+    println("{} {}", bar, baz)
+
+    let tuple = (1, 2, 3)
+
+    mut (a, b, c) = tuple
+    a++
+    b++
+    c++
+
+    println("{}, {}, {}", a, b, c)
+
+    let nested_tuple = (1, (2, 3))
+
+    mut test = Test(a: 5, b: 6, c: 7)
+
+    (test.a, (test.b, test.c)) = nested_tuple
+
+    println("{}", test)
 }
--- a/samples/tuples/destructuring_assignment.jakt
+++ b/samples/tuples/destructuring_assignment.jakt
@@ -1,5 +1,5 @@
 /// Expect:
-/// - output: "1 hello\n1 hello\n2, 3, 4\nTest(a: 1, b: 2, c: 3)\n"
+/// - output: "1 hello\n1 hello\n2, 3, 4\n1, 2, 3\n"
 
 function foo() -> (i64, String) {
     return (1, "hello")
@@ -14,14 +14,16 @@ struct Test {
 function main() {
     let (x1, x2) = foo()
 
-    println("{} {}", x1, x2)
+    print("{} ", x1)
+    println("{}", x2)
 
     mut bar = 0
     mut baz = ""
 
     (bar, baz) = foo()
 
-    println("{} {}", bar, baz)
+    print("{} ", bar)
+    println("{}", baz)
 
     let tuple = (1, 2, 3)
 
@@ -30,7 +32,9 @@ function main() {
     b++
     c++
 
-    println("{}, {}, {}", a, b, c)
+    print("{}, ", a)
+    print("{}, ", b)
+    println("{}", c)
 
     let nested_tuple = (1, (2, 3))
 
@@ -38,5 +42,7 @@ function main() {
 
     (test.a, (test.b, test.c)) = nested_tuple
 
-    println("{}", test)
+    print("{}, ", test.a)
+    print("{}, ", test.b)
+    println("{}", test.c)
 }
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -5860,50 +5860,7 @@ struct Typechecker {
                 }
             }
         }
-        IndexedTuple(expr, index, is_optional, span) => {
-            let checked_expr = .typecheck_expression_and_dereference_if_needed(expr, scope_id, safety_mode, type_hint: None, span)
-
-            let tuple_struct_id = .find_struct_in_prelude("Tuple")
-            let optional_struct_id = .find_struct_in_prelude("Optional")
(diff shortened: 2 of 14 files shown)


samples/tuples/destructuring_assignment.jakt
```diff
@@ -1,11 +1,42 @@
...
+    b++
+    c++
+
+    println("{}, {}, {}", a, b, c)
+
+    let nested_tuple = (1, (2, 3))
+
+    mut test = Test(a: 5, b: 6, c: 7)
+
+    (test.a, (test.b, test.c)) = nested_tuple
```
alimpfard, line 39: :thinking: strong python-style "WAT" vibes here, not sure if I'm totally in love with this tbh

selfhost/typechecker.jakt
```diff
@@ -4560,7 +4690,7 @@ struct Typechecker {
         Function(captures, params, can_throw, return_type, block, span) => .typecheck_lambda(captures, params, can_throw, return_type, block, span, scope_id, safety_mode)
         Try(expr, catch_block, catch_name, span) => .typecheck_try(expr, catch_block, catch_name, scope_id, safety_mode, span, type_hint)
         TryBlock(stmt, catch_block, error_name, error_span, span) => .typecheck_try_block(stmt, error_name, error_span, catch_block, scope_id, safety_mode, span)
-        Operator => {
+        else => {
```
alimpfard, line 4702: Please don't add an else here, let's just have the compiler yell at us when we add new ones

selfhost/typechecker.jakt
```diff
@@ -3494,19 +3494,28 @@ struct Typechecker {
...
-
-        checked_var_decl = .typecheck_var_decl(var: var_decl, init: expr, scope_id, safety_mode, span)
-        if checked_var_decl! is VarDecl(var_id) {
-            tuple_variable = .program.get_variable(var_id)
+        if checked_expr is Var(var) {
+            let maybe_var = .find_var_in_scope(scope_id, var: var.name)
+            if maybe_var.has_value() {
+                tuple_variable = maybe_var!
+            } else {
+                .error(format("Tuple `{}` not found", var.name), span)
```
alimpfard, line 3502: err...`variable {} not found`?

selfhost/typechecker.jakt
```diff
@@ -3494,19 +3494,28 @@ struct Typechecker {
...
+                is_mutable: false,
+                inlay_span: None,
+                span
+            )
+
+            checked_var_decl = .typecheck_var_decl(var: var_decl, init: expr, scope_id, safety_mode, span)
+            if checked_var_decl! is VarDecl(var_id) {
+                tuple_variable = .program.get_variable(var_id)
+            } else {
+                .error(format("Internal destructure varible `{}` not found", var_decl.name), span)
```
alimpfard, line 3517: This seems like a strange detail to expose as an error, given that it's not the user's fault, maybe panic instead?

alimpfard (commented):
TLDR: Not sure if I love the destructure-into-random-expressions bit, but here's some reviews regardless.

samples/tuples/destructuring_assignment.jakt
robryanx: The extension on what we have currently here is specifically for `IndexedStruct`.
For example in the compiler it would allow:
```
namespace_.name = name!.0
namespace_.name_span = name!.1
```
To be rewritten as
```
(namespace_.name, namespace_.name_span) = name!
```
Happy to remove it though if you think it is too much.

alimpfard: We can keep it and see if it does anything I guess, I get the feeling that it's going to be mildly surprising behaviour.

samples/tuples/destructuring_assignment.jakt
```diff
@@ -1,11 +1,48 @@
...
+
+    print("{}, ", a)
+    print("{}, ", b)
+    println("{}", c)
+
+    let nested_tuple = (1, (2, 3))
+
+    mut test = Test(a: 5, b: 6, c: 7)
+
+    (test.a, (test.b, test.c)) = nested_tuple
```
jntrnr, line 43: IMHO destructuring should go into bindings not expressions. In the above you're overloading the concept of assignment and binding, but I think you want to keep binding a bit more pure. For example, this would be fine:

```
let (e, (f, g)) = nested_tuple
```

tests/typechecker/destructure_missing_variable.jakt
```diff
@@ -0,0 +1,8 @@
+/// Expect:
+/// - error: "Variable 'bar' not found"
+
+function main() {
+    mut foo = 0
+
+    (foo, bar) = (1, 2)
```
jntrnr, line 7: Same here, the destructuring should go into bindings not assignments. `foo` already exists so unless you shadow it, you can't rebind it.

tests/typechecker/destructure_private.jakt
```diff
@@ -0,0 +1,19 @@
...
+        return TestClass(x, y)
+    }
+}
+
+function main() {
+    let tuple = (8, 9)
+
+    let test_class = TestClass::make(x: 6, y: 7)
+
+    (test_class.x, test_class.y) = tuple
```
jntrnr, line 18: Same here.

I'll stop repeating it, you get the idea.

jntrnr:
Left some comments. Just to echo some of @alimpfard's thoughts above, I think it's important to keep bindings separate from other types of things (eg expressions or assignments). This makes it easier to read and a bit less surprising.

robryanx:
> Left some comments. Just to echo some of @alimpfard's thoughts above, I think it's important to keep bindings separate from other types of things (eg expressions or assignments). This makes it easier to read and a bit less surprising.

I think Ali was specifically questioning the case on destructing into struct properties (IndexedStruct), which mainly seemed nice to me for completeness and could certainly be removed.

The idea of the PR was to allow for both assignments and bindings so:
`let (a, b) = (3, 4)`
Would we what we have currently.

But we could then have:
```
mut a = 1
mut b = 2

(a, b) = (3, 4)
```
Without a `mut`/`let` keyword for assigning new values to existing variables

The primary use case would be conditional destructing
```
mut a = 1
mut b = 2

if some_condition {
    (a, b) = (3, 4)
}
```

Whereas currently we might do
```
mut a = 1
mut b = 2

if some_condition {
    let (temp_a, temp_b) = (3, 4)

    a = temp_a
    b = temp_b
}
```
or just
```
mut a = 1
mut b = 2

if some_condition {
    let tuple = (3, 4)

    a = tuple.0
    b = tuple.1
}
```

So I guess the question is do we want destructing into existing variables. If so with this syntax or something else?

PR #1168, opened 2022-09-09
typechecker+codegen: Add basic support for inherited class fields

awesomekling:
Co-Authored-By: JT <jt@serenityos.org>

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -2959,19 +2959,30 @@ struct CodeGenerator {
 
             if not function_.params.is_empty() {
                 output += ": "
-                mut first = true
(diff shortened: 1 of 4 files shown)


selfhost/codegen.jakt
```diff
@@ -2939,14 +2939,37 @@ struct CodeGenerator {
...
+                        }
+                        first = false
+
+                        output += "move(a_"
+                        output += param.variable.name
+                        output += ")"
+                    }
+                    output += ")"
+
+                    first = false
```
lanmonster, line 2962: instead of juggling `first` around, you can use `join` from utility.jakt

PR #1174, opened 2022-09-10
lexer+formatter: Store raw number value along with NumericConstant

robryanx:
This allows the formatter to use the raw number value in the source
rather than the parsed `NumericConstant` when formatting.

Also fixed bitwise-and spacing.

### DIFF
--- a/selfhost/lexer.jakt
+++ b/selfhost/lexer.jakt
@@ -19,7 +19,7 @@ enum Token {
     SingleQuotedString(quote: String, span: Span)
     SingleQuotedByteString(quote: String, span: Span)
     QuotedString(quote: String, span: Span)
-    Number(number: NumericConstant, span: Span)
+    Number(number: NumericConstant, raw_number: String, span: Span)
     Identifier(name: String, span: Span)
     Semicolon(Span)
     Colon(Span)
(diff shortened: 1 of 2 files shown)


selfhost/lexer.jakt
```diff
@@ -332,10 +332,10 @@ enum NumericConstant {
     }
 }
 
-function make_float_token(number: f64, suffix: LiteralSuffix, span: Span) throws -> Token => match suffix {
+function make_float_token(number: f64, suffix: LiteralSuffix, raw_number: String, span: Span) throws -> Token => match suffix {
     // FIXME: consider unifying with make_integer_token() to a single function
-    LiteralSuffix::F32 => Token::Number(number: NumericConstant::F32(f64_to_f32(number)), span)
-    LiteralSuffix::F64 => Token::Number(number: NumericConstant::F64(number), span)
+    LiteralSuffix::F32 => Token::Number(number: NumericConstant::F32(f64_to_f32(number)), raw_number, span)
+    LiteralSuffix::F64 => Token::Number(number: NumericConstant::F64(number), raw_number, span)
```
jntrnr, line 338: this makes me wonder if we should just keep Token::Number a string and let the parser handle actually parsing it.

robryanx: Yeah it probably does make more sense, was thinking that while making the changes.

robryanx:
Closing this and submitting a new PR as moving number parsing to the parser is a significantly different approach.

PR #1176, opened 2022-09-11
all: Add fancy quotes support for Windows

jntrnr:
This does two things:

* Reverts a previous PR that moved from fancy quotes to backticks because of a Windows limitation
* Enabled UTF-8 support for Windows console apps so fancy quotes work properly. This happens as part of the C++ we codegen, so all apps built in Jakt should get this feature on Windows.

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -248,6 +248,7 @@ struct CodeGenerator {
         )
         mut output = ""
         output += "#include <lib.h>\n"
+        output += "#ifdef _WIN32\nextern \"C\" __cdecl void SetConsoleOutputCP(unsigned long code_page);\nconst unsigned long CP_UTF8 = 65001;\n#endif\n"
         let sorted_modules = generator.topologically_sort_modules()
         for idx in sorted_modules.size()..0 {
             let i = sorted_modules[idx - 1].id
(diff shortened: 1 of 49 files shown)


selfhost/codegen.jakt
```diff
@@ -248,6 +248,7 @@ struct CodeGenerator {
         )
         mut output = ""
         output += "#include <lib.h>\n"
+        output += "#ifdef WINDOWS\n#include <Windows.h>\n#endif\n"
```
ADKaster, line 251: This seems to be a reasonable fix... except that Windows.h is very, very heavy. 

I suspect that this will fail in CI because one of the headers pulled in by Windows.h  has a `#define` for `Yield`, which is used as a type in the compiler, and we didn't explicitly `#define NOMINMAX` here, so `min` and `max` will be macros in any generated code. 

I did quite the dance for this in #1149 . One of the main motivations to use a separate library for the runtime was to shield normal jakt code from platform-specifics and awkward headers. Side note, `<vxWorks.h>` defines `OK` and `ERROR` as integer constants :).

From what i've seen, conventional windows wisdom would be to actually forward declare the `SetConsoleOutputCP()` method the way it shows up in the Windows SDK headers. Without swapping over to my windows box, something like `extern "C" __cdecl void SetConsoleOutputCP(DWORD code_page);`

ADKaster: hm. I might be wrong about the compile failure? 

Unless `WINDOWS` isn't actually defined? I thought the OS detection macro was normally `_WIN32`

jntrnr: @ADKaster not sure but it did work locally with WINDOWS. I'll move it to `_WIN32` and do the forward declare like you suggest

ADKaster: thanks! side note on the forward declare, I think you can technically forward declare a global method at function scope. though that might be more awkward than forward declaring it up next to lib.h.

Not pulling in Windows.h also means either a local define or hard coded constant for CP_UTF8, which is `65001` iirc. 

Ah but you've already pushed it optimistically. :) The compiler shall be the source of truth on this one I think.

PR #1179, opened 2022-09-12
all: Move `NumericConstant` parsing to the parser

robryanx:
This changes the lexer to output a `Number` token with: `prefix: LiteralPrefix, number: String, suffix: LiteralSuffix` instead of
`number: NumericConstant`

The parser then parses this token into a `NumericConstant`

The primary motivation for this is so that the Number token carries the information required to reconstruct the source for formatting. It also makes for a better separation of concerns though as this was parsing that the lexer was doing.

This also changes the lexing slightly in that for example for an octal: `0o78` would end up as `Garbage` due to the 8 being an invalid digit. It would now lex as `0o7` and then a number `8`

`0o8` would lex as `0o` and `8`, with `0o` giving an error in the parser as having no digits.

### DIFF
--- a/selfhost/formatter.jakt
+++ b/selfhost/formatter.jakt
@@ -108,20 +108,7 @@ struct FormattedToken {
         SingleQuotedString(quote) => format("'{}'", quote)
         SingleQuotedByteString(quote) => format("b'{}'", quote)
         QuotedString(quote) => format("\"{}\"", quote)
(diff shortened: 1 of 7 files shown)


dangrie158:
Great, I had this change in mind when trying to parse Format specifiers. Although this specific issue is now obsolete I also think the lecker was the wrong place to parse the numeric constant

PR #1181, opened 2022-09-12
vscode: Fix file:// URI for gotodef

jntrnr:
Looks like we weren't quite prepending a whole `file://` URI, which caused goto-def to struggle to find the files when you would run goto-def.

This seems like a new-ish issues, I'm not sure if it's a regression in jakt or a difference in how vscode runs in recent versions.

### DIFF
--- a/editors/vscode/server/src/server.ts
+++ b/editors/vscode/server/src/server.ts
@@ -144,7 +144,7 @@ async function goToDefinition(
                 ? (await fs.promises.readFile(obj.file)).toString()
                 : document.getText() ?? ""
         );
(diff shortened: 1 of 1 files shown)


AtkinsSJ:
The git history says it's been like that since https://github.com/SerenityOS/jakt/commit/6da79b2d08da14ad1249432eeef7eeb49a10718a, so a while!

jntrnr:
I don't know why it broke recently, then. I know vscode got updated in the last two weeks, so it's possible it no longer lets us cheat on the number of forward slashes 🤔

treuks:
Works directly by using F12. No links to click though

image

PR #1183, opened 2022-09-14
lexer+parser+types: Add else bindings

robryanx:
This allows us to avoid writing representative mapping for enum variants
by allowing the default mapping to be done in the `CatchAll` match 
pattern.

`CatchAll` with bindings is always typechecked last and will typecheck 
all missing variants with those bindings and codegen each one 
separately.

For example things like
```
match this {
    Boolean(span) => span
    NumericConstant(span) => span
    QuotedString(span) => span
    ByteConstant(span) => span
    CharacterConstant(span) => span
    ...
}
```
can become
```
match this {
    else(span) => span
}
```

This also allows for a mix of specific cases and a catch all such as
```
function type(this) -> TypeId => match this {
    Boolean => builtin(BuiltinType::Bool)
    NumericConstant(type_id) => type_id
    EnumVariantArg(arg) => arg.type_id
    IndexedExpression(type_id) => type_id
    IndexedDictionary(type_id) => type_id
}
```
becoming
```
function type(this) -> TypeId => match this {
    Boolean => builtin(BuiltinType::Bool)
    EnumVariantArg(arg) => arg.type_id
    else(type_id) => type_id
}
```


samples/match/match_enum_else_binding.jakt
```diff
@@ -0,0 +1,26 @@
...
+    match foo {
+        else(x) => {
+            println("{}", x)
+        }
+    }
+
+    let bar = Foo::B(x: 2)
+
+    match bar {
+        else(x) => {
```
jntrnr, line 22: can we add another test where we match some, but not all, of the variants and then use `else(x)` to catch the rest of the cases and require that the remainder have the common field?

robryanx: Sure, I added it to the example and added a test for the failure case where a variant is missing the common field and a test for failing when attempting to use it with a generic match.

jntrnr: I meant something more like this:

```
enum Foo {
    A(x: i64)
    B(x: i64)
    C(x: i64)
}

function main() {
    let a = Foo::A(x: 1)

    match a {
        A(x) => println("I found an A {}", x)
        else(x) => println("Just {}", x)
    }
}
```

robryanx: Added that case as well, reversed the cases so that it also tests that it doesn't match A in the else case when it appears first.

samples/match/match_enum_else_binding.jakt
```diff
@@ -0,0 +1,46 @@
...
+    A(y) => y
+    else(x) => x
+}
+
+function main() {
+    let a = Foo::A(x: 1)
+
+    match a {
+        else(x) => println("Just {}", x)
+        A(x) => println("I found an A {}", x)
```
jntrnr, line 26: Wouldn't this one just always print the first case? `else(x)` would bind to all the cases.

jntrnr:
Recommendation here: add one test per file instead of multiple tests in the same file, so that you don't make the unit test harder to read and that it's immediately obvious what's intended for the given output.

samples/match/match_enum_else_binding.jakt
robryanx: Current behavior is that match will work irrespective of order, so you can have else first and it would still match to A, so I am keeping that same behavior for else bindings.

robryanx:
> Recommendation here: add one test per file instead of multiple tests in the same file, so that you don't make the unit test harder to read and that it's immediately obvious what's intended for the given output.

Yeah, I can split this up a bit. I think at the moment it can be a bit unclear with samples vs tests, I see the samples as being "here is what you can do with the feature" and the tests being "here are more specific success/failure cases".

jntrnr:
Yeah true, though with this one because it's such a new kind of feature that people might not expect or understand when they first see it, keeping the samples clear I think will give us a lot of benefit

samples/match/match_enum_else_binding.jakt
```diff
@@ -0,0 +1,24 @@
+/// Expect:
+/// - output: "I found an A 1\nJust 2\n"
```
jntrnr, line 2: This should be "Just 1" rather than "I found an A 1"

robryanx: Do you mean because if the else is first it should override the A match?
The current behavior without else bindings is:
```
let a = Foo::A

match a {
    else => println("else")
    A => println("I found an A")
}
```
Will print `I found an A`

samples/match/match_enum_else_binding.jakt
jntrnr: I think that'd be too surprising since order matters in a match traditionally.

samples/match/match_enum_else_binding.jakt
jntrnr: That looks like it's a bug to me, since there isn't a way for `else` to be a catchall if your cases overlap in some way.

Pinging @alimpfard for thoughts here too. My assumption was that `match` would work like other languages, where we start matching at the top and go down the list.

alimpfard: 🤔 I originally made "else" just a name for "all the cases not mentioned in the match" as that was easier to implement, so the order never really mattered (if that part of the IDE extension is still working, hovering else would then show you which cases it covers)

Interestingly, `else` doesn't actually feel like an irrefutable pattern to me, where e.g. `_` does...

alimpfard: Both behaviours seem sensible in a way to me, "else" being a name for "everything else" and "else" being an irrefutable pattern, I kinda like the O(1)-ness of the match, but it also isn't consistent with value matches (we do go down the list checking for equality there)

jntrnr: @robryanx - if that's how it already works, that shouldn't block this PR, I just wanted to have a clear understanding of where we want to go

alimpfard: I'll make an issue for that so we can discuss it there.

PR #1185, opened 2022-09-15
selfhost: Allow namespaced sub-modules, and create Standard:: module

ADKaster:
Few things here:

- Allow namespaced modules, with `::` as the namespace delimeter.
  -  `import Foo::Bar` will search for "Foo/Bar.jakt" in each of the configured include paths. It will then search under ``$(basename current-file.jakt)`` and under ``$(basename main-file.jakt)``
  
- Special case the module namespace `Standard::` to point to ``runtime/Standard``. 

- Wrestle with the bootstrap compiler to try to keep the commits atomic :sob:  If needed I can squash these up and only update bootstrap once in the PR, at the cost of atomicity. 

- Update jakttest to allow testing sub-modules with local cpp includes they want to use, and add a test that tests that capability.


jntrnr:
Weird, the CI error looks like the bootstrap is the older bootstrap

robryanx:
> Weird, the CI error looks like the bootstrap is the older bootstrap

I assume that this bootstrap would have been generated before the else bindings one, so it errors on that.

ADKaster:
Yeah, I rebased and copied over the bootstrap from my build dir, but I need to re-bootstrap my changes from main instead

ADKaster:
OBE

PR #1187, opened 2022-09-17
Multi~~track~~file ~~drifting~~ codegen

alimpfard:
- Splits each module into a .h and a .cpp
- `import fn() { ... }`
- `import extern "foo" { ... } before_include define { foo = "bar" } after_include undefine { foo }`
- idk, probably more stuff


jntrnr (approved):

selfhost/main.jakt
```diff
@@ -472,65 +573,166 @@ function main(args: [String]) {
...
-            extra_include_paths
-            extra_lib_paths
-            extra_link_libs
-            optimize
-        )
+            total_jobs += 1
+        }
 
-        if run_executable and compiler_status == 0 {
-            return system(output_filename.c_string())
```
robryanx, line 522: Seems that currently the option to run the compiled executable has been dropped.

selfhost/main.jakt
```diff
@@ -472,65 +573,166 @@ function main(args: [String]) {
...
 
-    mut cpp_filename: String = ""
-    mut output_filename: String = ""
+    mut depfile_builder = StringBuilder::create()
 
-    if (write_source_to_file or build_executable or run_executable) {
-        
-        if set_output_filename.has_value() {
-            cpp_filename = binary_dir + "/" + set_output_filename! + ".cpp"
-            output_filename = binary_dir + "/" + set_output_filename!
```
robryanx, line 484: Custom output filename isn't being used currently.

selfhost/main.jakt
alimpfard: Oh _that's_ the commented stuff I deleted and forgot about!

selfhost/main.jakt
alimpfard: It's a meaningless flag now, will remove.

robryanx: If it is removed how would we give the binary a custom name now?

alimpfard: yep, used it for the executable file name.

alimpfard:
TODO items:
- Update install rules and fixup the default value for `--runtime-library-path`
- Investigate runtime issue on windowz

selfhost/main.jakt
```diff
@@ -210,6 +310,9 @@ function main(args: [String]) {
     //     return 1
     // }
 
+    let guessed_output_filename = binary_dir + "/" + file_path.basename(strip_extension: true)
+    let output_filename = set_output_filename ?? guessed_output_filename
```
robryanx, line 314: I think set_output_filename should also be relative to the build directory.

jntrnr: also makes me think we should replace all the `+ "/" + ` with `+ path_sep() +` and have `path_sep` be a compile-time function

robryanx:
There is some issue at the moment that I don't fully understand where if I build jakt with ninja, then rebuild jakt with jakt, then rebuild again I end up with an `unknown-unknown-unknown-unknown` triple string so it will use unknown_process.jakt

selfhost/main.jakt
alimpfard: >I think set_output_filename should also be relative to the build directory.
And not the directory the compiler is invoked in? that's highly unusual

> path_sep
sure, why not

alimpfard:
Not with cmake? cmake is supposed to set a "basically correct" triple for the build, you'll have to pass `-T $(clang -dumpmachine)` if you're not using cmake to configure the build

selfhost/main.jakt
robryanx: I guess this is fine.
Before for example would be `jakt -o jakt selfhost/main.jakt` which would now be `jakt -o build/jakt selfhost/main.jakt`

robryanx:
> Not with cmake? cmake is supposed to set a "basically correct" triple for the build, you'll have to pass `-T $(clang -dumpmachine)` if you're not using cmake to configure the build

Thanks, that works fine.

alimpfard:
Final state:
- Making the compiler build the executable directly fails on windows, because [c1c365e7f7d3cc20b1a5d079df3eb78ae306522b](https://github.com/SerenityOS/jakt/commit/c1c365e7f7d3cc20b1a5d079df3eb78ae306522b) uses different compiler arguments, and windows is exceptionally nitpicky about some of those. (CMake works though)
- Other supported platforms are okay.
- CMake rebuilds everything if a module changes, but rebuilds are much faster, so I don't think this PR needs to address that - maybe a future PR.
- The compiler _can_ build in parallel, defaults to 2 jobs (pass `-J N` to build with N jobs)

I consider this PR done, if someone wants to make linking work on windows, make a PR :^)

jntrnr:
Tested on Windows and seems to be working for the cases we want to work. There's some autoconfiguration stuff we'll want to add in the future to cut down on the number of necessary compiler flags, but the fundamentals look good.

jakttest/run_one.py
```diff
@@ -0,0 +1,111 @@
...
+            )
+        except subprocess.CalledProcessError as e:
+            sys.exit(2)
+
+    # Run the executable inside the parent directory of the test file,
+    # dump stderr into runtest.err and stdout into runtest.out in temp_dir
+    # exit with status == 1 on failure
+    with open(temp_dir / "runtest.out", "w") as stdout, open(
+        temp_dir / "runtest.err", "w"
+    ) as stderr:
```
ADKaster, line 97: this formatting is kinda sus, we should add python related CI later.

meta/update_bootstrap.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+set -e
+
+: "${CURRENT_JAKT_COMPILER:=build/jakt}"
+: "${JAKT_RUNTIME_DIR:=build/runtime}"
```
ADKaster, line 6: this script assumes the selfhost built jakt rather than cmake? intended or an artifact of 28 commits?

runtime/lib.h
```diff
@@ -378,6 +367,32 @@ ALWAYS_INLINE constexpr OutputType as_truncated(InputType input)
...
+inline String ___jakt_get_target_triple_string()
+{
+#ifdef __JAKT_BUILD_TARGET
+    return String(__JAKT_BUILD_TARGET);
+#else
+// Pure guesswork.
+#   if defined(WIN32)
+    return String("i686-pc-windows-msvc");
+#   elif defined(WIN64)
+    return String("x86_64-pc-windows-msvc");
```
ADKaster, line 379: Is WIN64 and WIN32 still correct here? I remember mentioning _WIN64 and _WIN32

selfhost/utility.jakt
```diff
@@ -279,4 +223,24 @@ function interpret_escapes(anon s: String) throws -> String {
     }
 
     return builder.to_string()
-}
\ No newline at end of file
+}
+
+function null<T>() -> raw T {
```
ADKaster, line 228: this and allocate are duplicated in posix_process.jakt

selfhost/windows_process.jakt
```diff
@@ -0,0 +1,293 @@
...
+        throw Error::from_errno(errno_value())
+    }
+}
+
+function wait_for_some_set_of_processes_that_at_least_includes(processes: &[usize:Process]) throws -> (usize?, ExitPollResult) {
+    if processes.is_empty() {
+        throw Error::from_errno(38)
+    }
+
+    mut handles = allocate<raw void>(count: processes.size())
```
ADKaster, line 254: this allocation is leaked

ADKaster (commented):
Reviewed the fancy parts, not the compiler part.

Some of the list manip in the cmake looks funny but I can fix that up later.

jakttest/run_one.py
alimpfard: black 🤷

meta/update_bootstrap.sh
alimpfard: intended, but I'll change the default.

selfhost/utility.jakt
alimpfard: `allocate` there is calloc, this one is malloc.
will dedup null though.

PR #1189, opened 2022-09-18
codegen: Move match else case to the end for generic/value match

robryanx:
This allows an else case to be placed at any position in a value match and only match if no other case matches. The other cases will still be matched in order.

This makes the behavior closer to that of an enum match with the else case only matching if no other case matches.


jntrnr:
We chatted a bit about this and it seems generally there's a preference for `else` to come last rather than coming anywhere in the list.

robryanx:
Sure that is fine. I can update the approach to give an error if it doesn't appear last.

Issue #973, opened 2022-07-25
Selfhost Compiler Crashed with Segmentation Fault printing a LinkedList


horenso:
WHF :) On the most recent `main` branch (commit `9c52bee6`) I tried to compile the following progam:

```
struct LinkedList {
    value: i64
    next: LinkedList?
}

function main() {
    mut node1 = LinkedList(value: 1, next: None)
}
```
and the compiler (not the resulting program) crashed with a Segmentation Fault.|

Steps I took:
```
cargo install --path .
jakt ./selfhost/main.jakt && mv ./build/main ./jakt_selfhost
./jakt_selfhost test.jakt # test.jakt is the program shown above
```

The Rust-based compiler doesn't crash and produces some Cpp which clang then rejects with
`error: invalid application of 'sizeof' to an incomplete type 'Jakt::LinkedList'`

robryanx:
It is basically this line: `if struct_.record_type is Class and not top_level {` in codegen.jakt
As the record type is Struct it ends up with infinite recursion.
If the code is changed from `struct` to `class` it works, I am not sure if there should be an error about recursion in structs or that the codegen should be changed to handle it.

horenso:
Ah interesting, since `Struct` is a value type it makes sense that it cannot contain itself as a value.

maddanio:
Basically the same as with recursive enums which have to be boxed (aka class)

bytebaendiger:
So the compiler should reject code in which a value type references itself directly, as in:
```jakt
struct S {
    field: S
}
```
and indirectly through other value types such as tuples, arrays or other structs (e. g. `Optional`):
```jakt
struct S {
    field: S?
}
```
whilst allowing non-value types such as pointers to reference it safely?

maddanio:
only if it is a struct, it should work for classes.

maddanio:
i.e. this should be ok:
```
class S {
    field: S
}
```

PR #1191, opened 2022-09-19
Typechecker: Reject recursive structs/value types

laubsammler:
Previously, recursive value types, more precisely structs, used to crash the compiler.
Now it is able to detect them by walking the struct's dependency hierarchy (in terms of types), checking every field's type wether it is or contains the "root structs" type. :^)


bytebaendiger:
Sorry, forgot to mention that this PR addresses issue  #973.

PR #1195, opened 2022-09-24
Template Literals

dangrie158:
WHF, 

After #1151 got the feedback that it would be good to differentiate string interpolation from standard string literals I decided to give template literals a go.

This PR introduces template literals with a syntax closely following [ES6 template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). Also inspired by the ES6 design, if no tag function is prepended to the literal, string interpolation is performed by default:

```
let world = "friends"
`Hello ${world}`
```

But that's only the start :) You can also define a custom tag function to control during comp time and runtime how to transform the template literal, e.g. you could define a `json` tag function to type check json during compiletime and create a valid representation during runtime with interpolated values:
```
println(json`{"${ 123 }": ${ string },
"float": ${ float_ },
"array": ${ array },
"dict": ${ dict }
}`) 
```
would output:
```
{"123": "foo",
"float": 12.5,
"array": [{"can": "contain a set"}]
"dict": {"This": ["Contains a set"]}
}
```

However, you're not limited to return strings in your tag function but can return any type (e.g. parse an object tree into another representation 🤷 )

The tag function is any comp time function that expects a single `quotes` parameter that contains the string parts of the literal and returns a function that gets called during runtime and gets all the evaluated embedded expressions, so
```
tag`foo ${bar}`
```  
is roughly equal to:
```
tag(["foo"])(bar)
```
However to be able to pass arbitrary values during runtime (as the evaluated expressions can have different types) the need to get wrapped into a generic container. This container is the enum `JaktValue` that contains different entries for all possible primitive types (plus Optionals, Arrays, Sets and Dicts). This type is defined in the runtime and has a corresponding declaration in the prelude to make it usable by any user program. The Implementation of this Type is partially autogenerated from the definition in the prelude and extended by some convenience constructors.
As I'm absolutely new to C++, I'm sure there are many improvements possible especially there :)

In `samples/template_literals/` are 2 working examples.

Let me know what you think.

Known Limitations:

- [ ] No Interpreter support yet
- [ ] The JaktValue stuff needs to go
- [ ] Need support for stdlib first to factor out implementation of interpolate_string from the prelude


jntrnr:
A few thoughts - 

Can you split off the IDE work from this PR? It looks unrelated and could probably be reviewed and handled separately
For the compile time stuff, let me tag @alimpfard for thoughts.

alimpfard:
I think the value stuff can probably wait for a static reflection MVP (which is just waiting for a standard library split).
Re the implementation, seems fine methinks

dangrie158:
Alright I split out the first three and the 6th commit. Should I create separate PRs for each logical change?

For the reflection stuff: I have some ideas I will write down tonight and post on discord. I think it should be doable without having any intermediary value type or anything else in the stdlib but maybe I just don’t possess the c++ foo to see why this won’t be doable

jntrnr:
> separate PRs

yes please

PR #1196, opened 2022-09-26
typechecker: Add better error for calling an instance method statically

robryanx:
Previously this would just get the "Wrong number of arguments" error because an instance method has the `this` argument.

### DIFF
--- a/tests/typechecker/call_instance_method_statically.jakt
+++ b/tests/typechecker/call_instance_method_statically.jakt
@@ -1,0 +1,16 @@
+/// Expect:
+/// - error: "Cannot call an instance method statically\n"
+
+class Foo {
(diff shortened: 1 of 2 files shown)


tests/typechecker/call_instance_method_statically.jakt
```diff
@@ -0,0 +1,16 @@
+/// Expect:
+/// - error: "Cannot call an instance method statically\n"
```
jntrnr, line 2: Can we add some help text too?

Something like "did you mean to type `.b()`?"

robryanx: Currently I added the hint "Add a dot before the method name to call an instance method"

PR #1199, opened 2022-09-27
formatter: Additional fixes

robryanx:


jntrnr:
Looks good. There's a conflict with some of the other formatting work, but after that I think we're probably good to land

robryanx:
> Looks good. There's a conflict with some of the other formatting work, but after that I think we're probably good to land

Rebased now.

PR #1200, opened 2022-09-27
parser+typechecker: Update lambda return type inference

robryanx:
This updates lambdas to match the rules now for other types of functions
- Only infer type if it is an arrow function.
- For non arrow functions without a return type specified the return type becomes void instead of unknown.


selfhost/parser.jakt
```diff
@@ -3142,17 +3142,19 @@ struct Parser {
...
                 .index++
                 let expr = .parse_expression(allow_assignments: true, allow_newlines: false)
                 let span = expr.span()
                 yield ParsedBlock(stmts: [ParsedStatement::Return(expr, span)])
             }
             else => .parse_block()
         }
-
-        return ParsedExpression::Function(captures, params, can_throw, return_type, block, span: merge_spans(start, .current().span()))
+
```
jntrnr, line 3156: stray space?

robryanx:
> Do we have any fat-arrow lambda tests that show that they still infer the type?

Yeah, `samples/closures/function_as_variable.jakt` covers that.

PR #1202, opened 2022-09-29
Function overloading

dangrie158:
This adds support for function overloading, so you can do stuff like: 

```
function to_string(anon x: i64) throws -> String {
    return format("{}i64", x)
}
function to_string(anon x: String) throws => x
function to_string<T>(anon z: T) throws => format("{}", z)


function main() {
    println(to_string(["Array"]))
    println(to_string("String"))
    println(to_string(64i64))
}
```

The type checker chooses the call to use based on the sum of specificities of all its arguments where an argument has a specificity based on its type:
- Any "direct" Type (primitives, structs, enums, classes) have 1 "unit" of specificity.
- A Type variable has a specificity of 0.
- Generic types and functions have their specificity calculated recusively based on the arguments / parameters so that a fully defined generic (i.e. <String, i64> compared to <K, V>) has a specificity slightly below 1 unit, so a primitive is always most specific.

Or, to match some wording from [cppreference](https://en.cppreference.com/w/cpp/language/overload_resolution):

For each pair of viable function F1 and F2 F1 is determined to be a better function than F2
1) if F1 is a non-template function while F2 is a template specialization
2) or, if not that, F1 and F2 are both template specializations and F1 is more specialized 
3) or, if not that, F1 and F2 are non-template functions with the same parameter-type-lists, and F1 is more constrained than F2

:)

### DIFF
--- a/runtime/prelude.jakt
+++ b/runtime/prelude.jakt
@@ -19,6 +19,7 @@ extern struct Array<T> {
     function resize(mut this, anon size: usize) throws
     function shrink(mut this, anon size: usize) -> void
     function push(mut this, anon value: T) throws
+    function push_values(mut this, anon other: &Array<T>) throws
     function pop(mut this) -> T?
     function iterator(this) -> ArrayIterator<T>
     function first(this) -> T?
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -614,24 +614,26 @@ struct CodeGenerator {
                 output += "\n"
             }
 
-            for (_, function_id) in scope.functions {
-                if not function_id.module.equals(current_module.id) {
-                    continue
-                }
+            for (_, overload_set) in scope.functions {
+                for function_id in overload_set {
+                    if not function_id.module.equals(current_module.id) {
+                        continue
+                    }
 
-                let function_ = .program.get_function(function_id)
+                    let function_ = .program.get_function(function_id)
 
-                let previous_function = .current_function
-                .current_function = function_
-                defer .current_function = previous_function
+                    let previous_function = .current_function
+                    .current_function = function_
+                    defer .current_function = previous_function
 
-                if function_.linkage is External or function_.type is ImplicitConstructor or function_.type is ImplicitEnumConstructor {
-                    continue
-                }
+                    if function_.linkage is External or function_.type is ImplicitConstructor or function_.type is ImplicitEnumConstructor {
+                        continue
+                    }
 
-                if not function_.generics.params.is_empty() {
-                    output += .codegen_function(function_)
-                    output += "\n"
+                    if not function_.generics.params.is_empty() {
+                        output += .codegen_function(function_)
+                        output += "\n"
+                    }
                 }
             }
 
@@ -654,24 +656,26 @@ struct CodeGenerator {
         }
 
         mut output = ""
-        for (_, function_id) in scope.functions {
-            if not function_id.module.equals(current_module.id) {
-                continue
-            }
+        for (_, overload_set) in scope.functions {
+            for function_id in overload_set {
+                if not function_id.module.equals(current_module.id) {
+                    continue
+                }
 
-            let function_ = .program.get_function(function_id)
+                let function_ = .program.get_function(function_id)
 
-            let previous_function = .current_function
-            .current_function = function_
-            defer .current_function = previous_function
+                let previous_function = .current_function
+                .current_function = function_
+                defer .current_function = previous_function
 
-            if function_.linkage is External or function_.type is ImplicitConstructor or function_.type is ImplicitEnumConstructor {
-                continue
-            }
+                if function_.linkage is External or function_.type is ImplicitConstructor or function_.type is ImplicitEnumConstructor {
+                    continue
+                }
 
-            if function_.generics.params.is_empty() {
-                output += .codegen_function(function_)
-                output += "\n"
+                if function_.generics.params.is_empty() {
+                    output += .codegen_function(function_)
+                    output += "\n"
+                }
             }
         }
 
@@ -692,18 +696,20 @@ struct CodeGenerator {
             output += .codegen_debug_description_getter(struct_)
 
             let scope = .program.get_scope(struct_.scope_id)
-            for (_, function_id) in scope.functions {
-                let function_ = .program.get_function(function_id)
-                let previous_function = .current_function
-                .current_function = function_
-                defer .current_function = previous_function
+            for (_, overload_set) in scope.functions {
+                for function_id in overload_set {
+                    let function_ = .program.get_function(function_id)
+                    let previous_function = .current_function
+                    .current_function = function_
+                    defer .current_function = previous_function
 
-                if function_.type is ImplicitConstructor {
-                    output += .codegen_constructor(function_)
-                    output += "\n"
-                } else if not function_.type is ImplicitEnumConstructor and not function_.is_comptime and function_.generics.params.is_empty() {
-                    output += .codegen_function_in_namespace(function_, containing_struct: struct_.type_id)
-                    output += "\n"
+                    if function_.type is ImplicitConstructor {
+                        output += .codegen_constructor(function_)
+                        output += "\n"
+                    } else if not function_.type is ImplicitEnumConstructor and not function_.is_comptime and function_.generics.params.is_empty() {
+                        output += .codegen_function_in_namespace(function_, containing_struct: struct_.type_id)
+                        output += "\n"
+                    }
                 }
             }
         }
@@ -727,14 +733,17 @@ struct CodeGenerator {
             }
 
             let scope = .program.get_scope(enum_.scope_id)
-            for (_, function_id) in scope.functions {
-                let function_ = .program.get_function(function_id)
-                let previous_function = .current_function
-                .current_function = function_
-                defer .current_function = previous_function
-                if not function_.type is ImplicitConstructor and not function_.type is ImplicitEnumConstructor and not function_.is_comptime {
-                    output += .codegen_function_in_namespace(function_, containing_struct: enum_.type_id)
-                    output += "\n"
+            for (_, overload_set) in scope.functions {
+                for function_id in overload_set {
+                    let function_ = .program.get_function(function_id)
+                    let previous_function = .current_function
+                    .current_function = function_
+                    defer .current_function = previous_function
+
+                    if not function_.type is ImplicitConstructor and not function_.type is ImplicitEnumConstructor and not function_.is_comptime {
+                        output += .codegen_function_in_namespace(function_, containing_struct: enum_.type_id)
+                        output += "\n"
+                    }
                 }
             }
         }
@@ -789,22 +798,24 @@ struct CodeGenerator {
             output += .codegen_namespace_predecl(scope: .program.get_scope(child), current_module)
         }
 
-        for (_, function_id) in scope.functions {
-            if not function_id.module.equals(current_module.id) {
-                continue
-            }
-            let function_ = .program.get_function(function_id)
-            if function_.type is ImplicitEnumConstructor {
-                continue
-            }
+        for (_, overload_set) in scope.functions {
+            for function_id in overload_set {
+                if not function_id.module.equals(current_module.id) {
+                    continue
+                }
+                let function_ = .program.get_function(function_id)
+                if function_.type is ImplicitEnumConstructor {
+                    continue
+                }
 
-            let previous_function_id = .current_function
-            .current_function = Some(function_)
-            defer .current_function = previous_function_id
+                let previous_function_id = .current_function
+                .current_function = Some(function_)
+                defer .current_function = previous_function_id
 
-            if not function_.type is ImplicitConstructor and function_.name != "main" {
-                output += .codegen_function_predecl(function_)
-                output += "\n"
+                if not function_.type is ImplicitConstructor and function_.name != "main" {
+                    output += .codegen_function_predecl(function_)
+                    output += "\n"
+                }
             }
         }
 
@@ -905,10 +916,10 @@ struct CodeGenerator {
             output += param.variable.name
         }
         output += ")"
-        
+
         if not function_.is_static() and not function_.is_mutating() {
             output += " const"
-        } 
+        }
         if function_.is_override {
             output += " override"
         }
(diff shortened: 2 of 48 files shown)


dangrie158:
Its probably necessary to rebase on #1135 once it is merged

maddanio:
what are the resolution rules though? or are they too obvious to document :)?

maddanio:
can I change the argument number/labeling between overloads?

samples/functions/default_arguments_in_method.jakt
```diff
@@ -0,0 +1,13 @@
+/// Expect:
+/// - output: "42\n"
+
+enum S {
+    X
+    function my_func(this, anon def: i64 = 42) {
```
robryanx, line 6: I guess that the change is specifically fixing enum default arguments?
As there is `samples/functions/default_arguments/method.jakt` that seems to work currently.
If so probably best to mention enum in the commit message/ sample name.

dangrie158: True, just noticed that as well when writing on discord

dangrie158:
> can I change the argument number/labeling between overloads?

Sure :). I had a bug when using different number of arguments but I just pushed the fix for that :)



> what are the resolution rules though? or are they too obvious to document :)?

Those are defined in the commit messages, I will copy them into the first post here for convenience :)

samples/functions/default_arguments_in_method.jakt
dangrie158: Updated the commit message now, thanks

maddanio:
> > can I change the argument number/labeling between overloads?
> 
> Sure :). I had a bug when using different number of arguments but I just pushed the fix for that :)

and what about only changing argument labels? is that allowed?

dangrie158:
Yep. However, if the labels don't match (for non-anon labels), it is not considered a match, so this:

```
function to_string(x: i64) throws -> String {
    return format("{}i64", x)
}
function to_string<T>(z: T) throws => format("{}", z)

function main() {
    println(to_string(z: 64i64))
}
```

would print `64` instead of `64i64` since the label of the int variant does not match and is thus not considered an overload

maddanio:
Interesting, so you can overload on labels. Wondering if that could be useful or should better be forbidden.

dangrie158:
> Yep. However, if the labels don't match (for non-anon labels), it is not considered a match, so this:

Scratch that. Currently that is an error during compilation, the above sample worked because one has a template parameter, but with twice the same signature it won't work. This could be fixable by mangling the function names during code gen based on the labels but before I fix this I think we should define if we want this at all or not.

I could think of use cases where it would be useful, e.g.
```
to_int(from_string: "123")
to_int(from_octal_string: "123")
```
but it may also be weird in other cases. I'll hop on discord and throw this in there.

dangrie158:
> I'll hop on discord and throw this in there.

Alright, consensus was pretty clearly towards disallowing overloading based on the labels alone, so I added a check to the type checker that throws an error if you attempt to add a function whose signature matches (i.e.has the same name, throws, comptime, and arguments, where arguments that are not generic types need to match the other function exactly while generic parameters always match.

dangrie158:
Just pushed the changes to get method overloading fixed and added a sample that uses it.

selfhost/typechecker.jakt
```diff
@@ -463,48 +468,74 @@ struct Typechecker {
         return true
     }
 
-    function add_function_to_scope(mut this, parent_scope_id: ScopeId, name: String, function_id: FunctionId, span: Span) throws -> bool {
+    function add_function_to_scope(mut this, parent_scope_id: ScopeId, name: String, function_ids: [FunctionId], span: Span) throws -> bool {
```
maddanio, line 471: is this actually adding an overload set? if so it should probably be called add_overload_set. well, really just the scope-local part of a potential overload set, right?

selfhost/main.jakt
```diff
@@ -480,9 +480,9 @@ function main(args: [String]) {
                     continue
                 }
 
-                for function_ in scope.functions.iterator() {
-                    if function_.0 == "main" {
-                        main_function_id = function_.1
+                for (function_name, function_ids) in scope.functions.iterator() {
```
maddanio, line 483: again, confusing that its called function when its really an overload set....not sure about a good name. overload set? function family? is this bikeshedding?

maddanio:
in general I am curious: how does the actual compile time resolution work, ist it actually performed by the c++ compiler in the end? that would mean that jakt has to completely match c++ overload rules. that would also mean that jakt would inherit c++s [ADL](https://en.cppreference.com/w/cpp/language/adl).
the same question could be posed with templates and SFINAE I guess...

dangrie158:
Thank you very much for the review:

> in general I am curious: how does the actual compile time resolution work, ist it actually performed by the c++ compiler in the end?

You're completely right. I thought it would be good to match the c++ resolution rules since the serenity community is probably familiar with them and we don't have to name mangle the generated calls if overloaded functions do exist.

However I guess the c++ rules have too many edge cases, many of which probably don't affect us (yet) and make it difficult to ensure the jakt compiler and c++ compiler always match up.

I will add name mangling to the code generator if the function set (yet, that name checks out :)) contains multiple values based on the parameter types, this way only the jakt compiler decides which function gets called in the end.

selfhost/typechecker.jakt
dangrie158: good point. I guess overload_set is a good name, although I then should probably change the store to a set just to make sure the name matches.

selfhost/main.jakt
dangrie158: > is this bikeshedding?
Nah, the current name is moot. I don't know if any canonical name exists but [CPP Reference](https://en.cppreference.com/w/cpp/language/overload_resolution) talks about sets of candidate functions, so I think overload set accurately describes the intent. I would still name the member of the Scope `functions` but this is then a mapping of names -> overload sets

dangrie158: ok, overload_set already reads a lot nicer 👍

samples/functions/overloading_different_return_types.jakt
```diff
@@ -0,0 +1,15 @@
+/// Expect:
+/// - output: "true\ntrue\ntrue\n"
+
+function identity(anon x: i64) throws => x
+function identity(anon x: String) => x
+function identity<T>(anon x: T) => x
```
ADKaster, line 6: What happens if you declare the generic function before the concrete overloads? Does it still prefer the concrete functions before instantiating the generic?

jntrnr: I'm a wee bit tempted to say we should try them in the order you declare them, and the first one that succeeds should win. Might be worth chatting through on discord (if we haven't already)

dangrie158: Currently all matching functions get scored by the function described in https://github.com/SerenityOS/jakt/pull/1202/commits/8185aac078223f97c30fa0e401895f74a0ac5ec1 and the highest score wins, so yes, the concrete function gets picked over the generic version regardless of declaration order.

maddanio: Is that implemented via name mangling?

I think using declaration order is not a good idea, declaration order can also be changed by shuffling imports or just cleaning up code to make it more readable, seems awfully fragile

maddanio: What is the use case for allowing such overloads?

dangrie158: > Is that implemented via name mangling?

The current implementation does not mangle any names. Consent on discord was that matching the (relevant) c++ rules is worth the tradeoff for increased compatibility instead of name mangling and doing our own thing (for now?)

> What is the use case for allowing such overloads?

One concrete example would be a catch-all method for variants like this:

```
boxed enum Box {
    Void
    Bool(value:bool)
    U8(value: u8)
    U16(value: u16)
    ..
    JaktString(value: String)
    JaktArray(value: Array<Box>)
    JaktDictionary(value: Dictionary<Box, Box>)
    JaktSet(value: Set<Box>)
    JaktOptional(value: Box)
    JaktNone
    Unknown

    function create(value: String) => Box::JaktString(value)
    function create(value: u8) => Box::U8(value)
    function create(values: Array<T>) {
        let jakt_values: Box = []
        for value in values.iterator() {
            jakt_values.push(Box::create(value))
        }
        Box::Array(value: jakt_values)
    }
    function create(value: T) => Box::Unknown
}
```

selfhost/types.jakt
```diff
@@ -593,6 +593,50 @@ class CheckedFunction {
...
+
+        for param_index in 0..this.params.size() {
+            let lhs_param = this.params[param_index]
+            let rhs_param = other.params[param_index]
+
+            let lhs_param_id = lhs_param.variable.type_id.to_string()
+            let rhs_param_id = rhs_param.variable.type_id.to_string()
+
+            if not lhs_param.variable.type_id.equals(rhs_param.variable.type_id)
+            and not (lhs_generic_type_ids.contains(lhs_param_id) and rhs_generic_type_ids.contains(rhs_param_id))
```
alimpfard, line 667: style nit: indent this

selfhost/codegen.jakt
```diff
@@ -613,24 +613,26 @@ struct CodeGenerator {
                 output += "\n"
             }
 
-            for (_, function_id) in scope.functions.iterator() {
```
alimpfard, line 616: commit description typo: repreent -> represent.

selfhost/typechecker.jakt
```diff
@@ -6146,15 +6291,49 @@ struct Typechecker {
...
+                                args = []
+                                for resolved_arg in resolved_args.iterator() {
+                                    args.push((call.name, resolved_arg))
+                                }
+                            }
+                        }
+                        MatchError(errors) => {
+                            for error in errors.iterator() {
+                                errors_while_trying_to_find_matching_function.push(error)
+                            }
```
alimpfard, line 6312: Note for later: Array::extend() would be nice.

runtime/prelude.jakt
```diff
@@ -19,6 +19,7 @@ extern struct Array<T> {
     function resize(mut this, anon size: usize) throws
     function shrink(mut this, anon size: usize) -> void
     function push(mut this, anon value: T) throws
+    function push_values(mut this, anon other: &Array<T>) throws
```
alimpfard, line 22: Discard the previous note lol.
Sidenote: Should we call this `extend` or keep it as `push_values`?

dangrie158: I wanted to name it `extend` originally but then saw that Array actually already has a push_values function that does kind of the same thing, so I thought this would be the more appropriate name, but I now see that `AK::Vector` uses `extend` as well, so happy to rename it. Should I "rectify" and rename the original `push_values` as well or would it be better to not touch this one as they are expecting different parameters?

tests/typechecker/class_private_field_default.jakt
```diff
@@ -2,12 +2,16 @@
 /// - error: "Can't access field ‘muda_amount’, because it is marked private\n"
 
 class GoodEncapsulation {
+    public function create(muda_amount: u32, ora_amount: u32) throws {
+        return GoodEncapsulation(muda_amount, ora_amount)
+    }
```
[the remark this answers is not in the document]
dangrie158: Nah. It actually fails with ```
build/Root Module.h:11:50: error: class member cannot be redeclared
static ErrorOr<NonnullRefPtr<GoodEncapsulation>> create(u32 muda_amount, u32 ora_amount);
``` 
if I mark both fields public.

I'll update it to actually compile if it wan't for the error that is expected

File: runtime/prelude.jakt (hunk 8 above)
Comment (alimpfard) [reply runtime/prelude.jakt:22]: Let's just keep it `push_values` for now.

**Comment by jntrnr** [comment]
This is close to ready.

Something I'd like to see before we land it is better unit tests. The current ones, using things like the identity function or returning a simple string, don't tell us if overloading is actually working. We could, for example, be falling into one of the cases and never touching the other ones, and it'd be difficult to know by looking at the test results.

Instead, can you write unit tests so that it's clear which of the cases matched?

Also, it'd be good to land a larger number of tests. Overloading is a pretty complicated feature generally, so we want to know it's going to work in different cases. Here are a few that came to mind:

* What happens if you use overloading in a class hierarchy?
* What happens if you import a function and then create an overload in the same scope as the import?
* Do we give an error if functions only differ by return type?

These are mostly just to keep things from breaking in the future, but it also helps us get a sense for what to expect in the different cases.

**Comment by dangrie158** [comment]
> Instead, can you write unit tests so that it's clear which of the cases matched?

Not sure I understood you correctly. The identity function checks that each overload defined is called, otherwise the return type should not match and the equality check should fail typechecking.
I did however rewrite the tests to match the style of other tests and also moved them from samples to tests since that's what they actually are :). I also added some affixes to the to_string sample so it is possible to ensure the correct method got called. 


> Also, it'd be good to land a larger number of tests. Overloading is a pretty complicated feature generally, so we want to know it's going to work in different cases. Here are a few that came to mind:

> - What happens if you use overloading in a class hierarchy?
Added a test. Name hiding prevents overloading of a method in the parent. Currently there is no syntax to explicitly call into the parent scope but I want to implement this in an extra PR as it requires even more changes (to the parser at least) and this one is already big enough :)

> - What happens if you import a function and then create an overload in the same scope as the import?
Added a test. Currently you can overload function in the scope no matter if imported or explicitly declared. I don't yet know if this is good or if you also would like to use name hiding. Let's discuss this on discord but I think this is definitely out of scope for this PR.

> - Do we give an error if functions only differ by return type?
Added a test for that as well: Yep, at least one parameter needs to differ

I also added tests to exercise the overload rules in specificity for generic parameters and a test to make sure you can't overload main. Let me know if you can think of anything else that's missing.

**Comment by jntrnr** [comment]
Hey @dangrie158 - we've got another PR that's been in the works for months to add traits that we're going to land before this one. It'll probably cause some waves. https://github.com/SerenityOS/jakt/pull/1242 (a continuation of https://github.com/SerenityOS/jakt/pull/972).

We still want to be able to do the function overloading, just a bit of an unfortunate case where both PRs are pretty extensive changes to similar areas of the code.

**Comment by alimpfard** [comment]
:sweat_smile: sorry for the conflicts...
on the bright side, looks like it's just a single file

**Comment by dangrie158** [comment]
Should be good to go

File: selfhost/types.jakt [hunk 10]
```diff
@@ -638,6 +638,50 @@ class CheckedFunction {
     public is_virtual: bool
     public is_override: bool
 
+    public function signature_matches(this, anon other: CheckedFunction) throws -> bool {
```
alimpfard, line 678: Would `Typechecker::signatures_match()` work for this?

alimpfard (commented):
Impl looks alright, I just have the same question as I did before: are there any differences between our overload resolution alg and C++'s one? because codegen could end up picking the wrong one without us noticing if so.

Also, how does this mesh with traits? not a requirement to get right in this PR, but I'd like to know if they're okay with each other so to speak.

selfhost/types.jakt
dangrie158: As far as I understand the implementation: no since those serve different purposes
- `CheckedFunction::signature_matches ` checks if all arguments are the same i.e. `f<T>(x: T)` is **not the same** as `f(x: String)`
- `Typechecker::signatures_match` checks if all arguments are compatible `f<T>(x: T)` is **compatible with** `f(x: String)`

but since this function is used to check if a function with the same signature is already defined we want the former behaviour here, otherwise we would not allow more specific overrides of generic functions.

dangrie158:
> Impl looks alright, I just have the same question as I did before: are there any differences between our overload resolution alg and C++'s one? because codegen could end up picking the wrong one without us noticing if so.

To start with: I would consider any differences a bug that needs fixing. That being said I tried my best to match the c++ rules where they matter (which is not too much with the current jakt features). Since I don't want to claim hate match I instead describe which corner cases I considered and then move the ball back to you. If you find any way to get a bogus resolution where jakt and clang differ I'm happy to fix them

1. Since in jakt there is no implicit integer conversion, we don't need to consider any candidates with different int types to be valid candidates for overload resolution. if there are multiple candidates for a call which only differ for an int argument, jakt selects the one exactly matching the call, and so will c++, so implicit int casts ✅
2. if multiple candidates match because there are multiple implementations using generics, the most specific matching the call is picked by jakt, which matches the c++ behaviour. This is tested in [function_overloading_calls_most_specific_generic_signature.jakt](https://github.com/SerenityOS/jakt/blob/57cc540982a7ca3f219c0ad11ea40da95c792512/tests/typechecker/function_overloading_calls_most_specific_generic_signature.jakt), so generic argument resolution ✅
3. When we search for candidate functions, we do not traverse the inheritance chain in the case for methods, so we match the c++ behaviour by definition until we invent some more syntax. Therefore: name hiding ✅
4. ??? I guess that's all the corners we have where we need to match c++, but let me know if I missed anything.

> Also, how does this mesh with traits? not a requirement to get right in this PR, but I'd like to know if they're okay with each other so to speak.

Well, if we assume we both have tests for all cases, which I would assume, all tests still pass, so I don't think we're stepping on our toes :) I guess we would need to define how both concepts would interop, But since this PR basically only allows to define a function name multiple times, the problems I see (default trait implementations getting resolved in a scope) are already handled?

alimpfard:
Alright, so if matching C++ resolution rules is the goal, this LGTM entirely.
I'd be happy to merge once the conflict is fixed :^)

dangrie158:
Should be fixed

alimpfard (approved):
LGTM. Let's just make sure CI is happy.

PR #1208, opened 2022-09-29
all: Warnings Cleanup

robryanx:
Much nicer to look at :^)

Includes a fix for yielding in a block after a try statement.


selfhost/codegen.jakt
```diff
@@ -1814,12 +1814,9 @@ struct CodeGenerator {
         Try(expr, catch_block, catch_name, span, type_id, inner_type_id) => {
             mut output = ""
             let fresh_var = .fresh_var()
-            let fresh_label = .fresh_label()
             let is_void = inner_type_id.equals(void_type_id())
             let try_var = .fresh_var()
 
-            .entered_yieldable_blocks.push((fresh_var, fresh_label))
-
```
alimpfard, line 1822: Could we have a test for this too?

robryanx: Yeah good idea, turns out that yielding from the catch block didn't work before or after this change as the result of the block wasn't handled as an expression. Fixed that and added a test both for yielding from catch and for the original fix here of yielding after a try expression.

PR #1210, opened 2022-09-30
Fix Jakttest on Apple Silicon

lanmonster:
and support all but one string function in comptime
and fix an ide error


alimpfard:
Now that this is using the arch, please swap the WIN32 and WIN64 branches to unbork the windows CI :upside_down_face:

Issue #1053, opened 2022-08-07
Implement all expressions and statements for the comptime interpreter


alimpfard:
## Statements
- [x] Expression
- [x] Defer
- [ ] DestructuringAssignment
- [x] VarDecl
- [x] If
- [ ] Block
- [x] Loop
- [ ] While
- [x] Return
- [x] Break
- [x] Continue
- [x] Throw
- [ ] Yield
- [x] InlineCpp (disallowed)
- [x] Garbage

## Expressions
- [ ] Boolean
- [x] NumericConstant
- [x] QuotedString
- [x] ByteConstant
- [x] CharacterConstant
- [ ] UnaryOp
    - [ ] PreIncrement
    - [ ] PostIncrement
    - [ ] PreDecrement
    - [ ] PostDecrement
    - [ ] Negate
    - [ ] Dereference
    - [ ] RawAddress
    - [ ] Reference
    - [ ] MutableReference
    - [x] LogicalNot
    - [ ] BitwiseNot
    - [ ] TypeCast
    - [ ] Is
    - [ ] IsEnumVariant
- [ ] BinaryOp
    - [x] Add
    - [x] Subtract
    - [x] Multiply
    - [x] Divide
    - [x] Modulo
    - [x] LessThan
    - [ ] LessThanOrEqual
    - [x] GreaterThan
    - [ ] GreaterThanOrEqual
    - [x] Equal
    - [ ] NotEqual
    - [ ] BitwiseAnd
    - [ ] BitwiseXor
    - [ ] BitwiseOr
    - [ ] BitwiseLeftShift
    - [ ] BitwiseRightShift
    - [ ] ArithmeticLeftShift
    - [ ] ArithmeticRightShift
    - [x] LogicalAnd
    - [x] LogicalOr
    - [ ] NoneCoalescing
    - [ ] Assign
    - [ ] BitwiseAndAssign
    - [ ] BitwiseOrAssign
    - [ ] BitwiseXorAssign
    - [ ] BitwiseLeftShiftAssign
    - [ ] BitwiseRightShiftAssign
    - [ ] AddAssign
    - [ ] SubtractAssign
    - [ ] MultiplyAssign
    - [ ] ModuloAssign
    - [ ] DivideAssign
    - [ ] NoneCoalescingAssign
    - [ ] Garbage
- [ ] JaktTuple
- [ ] Range
- [ ] JaktArray
- [ ] JaktSet
- [ ] JaktDictionary
[...]

PR #1215, opened 2022-10-04
More comptime fun

### LINKED ISSUE #1212: No warning about needing 'throws' when initializing a Set
```
function foo() {
    let set: Set<u8> = {}
}
```
this will fail at the cpp stage with:
```
build/Root Module.cpp:26:36: error: no viable conversion from returned value of type 'Jakt::Error' to function return type 'bool'
const JaktInternal::Set<u8> set = (TRY((Set<u8>::create_with_values({}))));
                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/lanmon/Development/jakt/build/include/runtime/Jakt/Try.h:16:20: note: expanded from macro 'TRY'
            return _temporary_result.release_error(); \
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
Error: Compilation failed
```

lanmonster:
This patch exposes a function from the underlying c++, `shrink`, and implements all methods except for hash.
Checks a bunch of boxes in #1053 and fixes #1212


jntrnr:
Cool - looks good for what I could tell

Looks like the bootstrap got updated since this one went up. You'll probably need a rebuild.

jntrnr:
@lanmonster did you mean to check in the `selfhost_build` directory?

lanmonster:
> @lanmonster did you mean to check in the `selfhost_build` directory?

no. fixed now!

PR #1221, opened 2022-10-11
selfhost: Allow namespaced sub-modules, and create jakt:: module (again)

demizer:
This PR replaces https://github.com/SerenityOS/jakt/pull/1185

@ADKaster requested I create a new PR

A new PR was necessary because the old PR was severely out-of-date due to the multi-file compilation improvements. This PR requires running `update_bootstrap.sh` in three different commits (already done).

There is still one issue with this PR that needs to be addressed:

**Cmake STDLIB_SOURCES kludge**

Unfortunately, I could not figure out how to get `jakt-executable.cmake` to optionally include generated `jakt__<foo>.cpp` files only if they exist at build time in .

Perhaps we can always generate Standard modules output in the compiler:

1. This can possibly be achieved using the same mechanism in typechecker.jakt that is used for prelude.jakt, but that feels like a hack.
1. The entire Standard library would be included with every program whether it is needed or not.

Original PR message:

> Few things here:
> 
> * Allow namespaced modules, with :: as the namespace delimeter.
> 
>    `import Foo::Bar` will search for "Foo/Bar.jakt" in each of the configured include paths. It will then search under `$(basename current-file.jakt)` and under `$(basename main-file.jakt)`
> 
> * Special case the module namespace Standard:: to point to runtime/Standard.
> 
> * Wrestle with the bootstrap compiler to try to keep the commits atomic sob If needed I can squash these up and only update bootstrap once in the PR, at the cost of atomicity.
> 
> * Update jakttest to allow testing sub-modules with local cpp includes they want to use, and add a test that tests that capability.


demizer:
Working on fixing the windows issue

selfhost/compiler.jakt
```diff
@@ -136,7 +136,8 @@ class Compiler {
         .get_file_id_or_register(file_name)
     }
 
-    public function search_for_path(this, anon module_name: String) throws -> Path? {
+    public function search_for_path(this, anon input_module_name: String) throws -> Path? {
+        let module_name = input_module_name.replace(replace: "::", with: "/")
         for include_path in .include_paths.iterator() {
             let candidate_path = include_path + "/" + module_name + ".jakt"
```
alimpfard: It'd be nice if we could use the path abstraction layer instead of building unixy paths manually

selfhost/compiler.jakt
```diff
@@ -138,6 +139,14 @@ class Compiler {
 
     public function search_for_path(this, anon input_module_name: String) throws -> Path? {
         let module_name = input_module_name.replace(replace: "::", with: "/")
+        // Note: Checking this after the replacement!
+        let standard_module_name = "Standard/"
+        if module_name.length() > standard_module_name.length() and standard_module_name == module_name.substring(start: 0, length: standard_module_name.length()) {
```
alimpfard, line 144: uhh...`if module_name.starts_with(standard_module_name)`?

selfhost/compiler.jakt
```diff
@@ -138,6 +139,14 @@ class Compiler {
...
+            let candidate_path = .std_include_path + "/" + module_name.substring(start: standard_module_name.length(), length: module_name.length() - standard_module_name.length()) + ".jakt"
+            if File::exists(candidate_path) {
```
alimpfard, line 146: Same note about path, except do it sooner here

selfhost/repl.jakt
```diff
@@ -229,6 +229,7 @@ struct REPL {
             dump_parser: false
             ignore_parser_errors: false
             debug_print: false
+            std_include_path: ""
```
alimpfard, line 227: :interrobang: y no correct standard lib path in repl? at least add a FIXME comment

meta/update_bootstrap.sh
```diff
@@ -11,7 +11,7 @@ set -e
 # Next, use that to build the selfhost
 rm -fr selfhost_build
 mkdir selfhost_build
-build/main --binary-dir selfhost_build -S -T unknown-unknown-unknown-unknown --runtime-library-path "$JAKT_RUNTIME_DIR" selfhost/main.jakt
+build/main --binary-dir selfhost_build -S -T unknown-unknown-unknown-unknown --runtime-library-path "$JAKT_RUNTIME_DIR" -I runtime selfhost/main.jakt
```
alimpfard, line 14: This should not be needed, why is it needed?

runtime/Standard/libc/io.jakt
```diff
@@ -0,0 +1,16 @@
...
+    extern function fopen(anon str: raw c_char, anon mode: raw c_char) -> raw FILE
+    extern function fclose(anon mut file: raw FILE) -> c_int
+    extern function feof(anon mut file: raw FILE) -> c_int
+
+    extern function fgetc(anon mut file: raw FILE) -> c_int
+    extern function fgets(s: raw c_char, n: usize, stream: raw FILE) -> raw c_char
+    extern function putchar(anon ch: c_int) -> c_int
+}
+
+extern function system(anon command: raw c_char) -> c_int
```
alimpfard, line 16: This should be in `stdlib.h`.

selfhost/compiler.jakt
```diff
@@ -148,9 +148,16 @@ class Compiler {
             }
         }
         for include_path in .include_paths.iterator() {
-            let candidate_path = include_path + "/" + module_name + ".jakt"
-            if File::exists(candidate_path) {
-                return Path::from_string(candidate_path)
+            let candidate_path = Path::from_parts([include_path, module_name + ".jakt"])
+            if candidate_path.exists() {
+                return Path::from_string(candidate_path.to_string())
```
alimpfard, line 153: wut?

selfhost/compiler.jakt
```diff
@@ -148,9 +148,16 @@ class Compiler {
...
+            }
+        }
+        let current_file_path = .current_file_path()
+        if current_file_path.has_value() {
+            let candidate_path = Path::from_parts([current_file_path!.parent().to_string(), module_name + ".jakt"])
+            if candidate_path.exists() {
+                return Path::from_string(candidate_path.to_string())
```
alimpfard, line 160: wut x2?

samples/modules/subfolder/submodule.jakt
```diff
@@ -0,0 +1,9 @@
+/// Expect: Skip
+
+import extern "Hello.h" {
+   extern function get_hello() -> String
+
+   extern class Printer {
+    public function print_something() -> void
```
alimpfard, line 7: spaces aren't particularly nutritious, I'd recommend not eating them at all.

alimpfard (commented):
+ my usual "but what if we call it 'library'" remark you may ignore.

samples/apps/cat.jakt
```diff
@@ -1,6 +1,8 @@
 /// Expect:
 /// - stderr: "usage: cat <path>\n"
 
+import Standard::libc::io { fopen, fclose, fgetc, feof, putchar }
```
jntrnr, line 4: can we call the standard library `jakt`?

demizer:
Hi folks, will start working on making the suggested changes this week. Thanks!

runtime/Standard/libc/io.jakt
demizer: @alimpfard I am not sure what you mean. Should there not be a an io.jakt?

ADKaster: We should import extern "stdlib.h" for the `system` function

selfhost/compiler.jakt
```diff
@@ -136,14 +136,17 @@ class Compiler {
         .get_file_id_or_register(file_name)
     }
 
-    public function search_for_path(this, anon module_name: String) throws -> Path? {
+    public function search_for_path(this, anon input_module_name: String) throws -> Path? {
+        mut builder = StringBuilder::create()
+        builder.append(get_path_separator())
+        let separator = builder.to_string()
```
alimpfard, line 143: ouch, but okay

alimpfard:
Forgot to type out the comment:

Looks good, just one nitpick.

selfhost/compiler.jakt
demizer: I felt the same way. Would it be possible to get a string version of the separator? Or is there a better way to convert u8 to a String?

demizer: Added FIXME.

jntrnr:
@demizer - out of curiosity - is there a way in a submodule to refer to something in a parent directory?

maddanio:
hmm, I guess you can make a symlink, but then you make your submodule depend on the parent, which seems wrong, because it is in effect an independent repo...

jntrnr:
@maddanio - it's pretty common to refer to parent modules. Say you want to refactor common code into a common module, there's no way to get to the common module without backtracking through the parent or starting from the root of the project to get to the common directory.

demizer:
@jntrnr I agree that there should be a way to reference a module in a parent directory from a submodule. I will work on a fix. Thanks!

maddanio:
Ah, now i get it, that was stupid, i was thinking git submodules, sorry:-D. Yah, that can make sense I guess in a transition phase, feels like code smell, but guess would be too much to forbid it

PR #1223, opened 2022-10-12
errors: Fix where error arrow points as number of lines grows

jntrnr:
This fixes where the arrow points in when an error occurs earlier in a larger file. Before, we were miscalculating the largest line number, which caused us to miscount how many spaces over to put the arrow and label.

Before:
```
Error: Type mismatch: expected ‘i64’, but got ‘String’
----- testme.jakt:3:15
 2 |   let x = "bob"
 3 |   let y = x / 3
                    ^- Type mismatch: expected ‘i64’, but got ‘String’
 4 |
-----
Error: Binary arithmetic operation between incompatible types (‘String’ and ‘i64’)
----- testme.jakt:3:11
 2 |   let x = "bob"
 3 |   let y = x / 3
                ^- Binary arithmetic operation between incompatible types (‘String’ and ‘i64’)
 4 |
-----
```

Now:
```
Error: Type mismatch: expected ‘i64’, but got ‘String’
----- testme.jakt:3:15
 2 |   let x = "bob"
 3 |   let y = x / 3
                   ^- Type mismatch: expected ‘i64’, but got ‘String’
 4 |
-----
Error: Binary arithmetic operation between incompatible types (‘String’ and ‘i64’)
----- testme.jakt:3:11
 2 |   let x = "bob"
 3 |   let y = x / 3
               ^- Binary arithmetic operation between incompatible types (‘String’ and ‘i64’)
 4 |
-----
```

This happened because in larger lines, we used the total number of lines as an indicator of what the largest line number in the error.

### DIFF
--- a/selfhost/error.jakt
+++ b/selfhost/error.jakt
@@ -71,10 +71,18 @@ function display_message_with_span(anon severity: MessageSeverity, file_name: St
     let line_spans = gather_line_spans(file_contents)
 
     mut line_index = 1uz
-    let largest_line_number = line_spans.size()
 
+    mut largest_line_number = 0uz
+    while line_index < line_spans.size() {
+        if span.start >= line_spans[line_index].0 and span.start <= line_spans[line_index].1 {
+            largest_line_number = line_index + 2 // 1 (row number) + 1 (extra source line)
+        }
+        ++line_index
+    }
+    
     let width = format("{}", largest_line_number).length()
 
+    line_index = 1uz
     while line_index < line_spans.size() {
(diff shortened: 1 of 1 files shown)


jntrnr:
Working on some improvements to this

jntrnr:
Should be good now. Second improvement: errors can now span across a different number of digits in line numbers and stay aligned:

```
Error: Binary arithmetic operation between incompatible types (‘String’ and ‘i64’)
----- testme.jakt:9:11
 8  |
 9  |   let y = x / 3
                ^- Binary arithmetic operation between incompatible types (‘String’ and ‘i64’)
 10 |
-----
```

Issue #1226, opened 2022-10-16
miscompile with functions in struct


wtholliday:
main.jakt:

```
import foo
```

foo.jakt:

```
struct Foo {

   function a() { }
   function b() { a() }

}
```

running `./build/bin/jakt main.jakt` I get:

```
build/foo.cpp:15:1: error: no member named 'a' in namespace 'Jakt::foo'; did you mean simply 'a'?
foo::a();
^~~~~~
a
build/foo.cpp:8:16: note: 'a' declared here
void foo::Foo::a() {
               ^
1 error generated.
Error: Compilation failed
```

Note that this works when `Foo` is simply in main.jakt.

wtholliday:
Seems like in `codegen_call` we need to detect when calling a static method via `function_.is_static()` and then append the struct name as a namespace. So `issue_1226_sub::Foo::a` instead of `issue_1226_sub::a`.

wtholliday:
Wondering why `struct_id` is `None` here: https://github.com/SerenityOS/jakt/blob/c0e8f6cb9dedcab13b39a82fde028941160fe4a1/selfhost/typechecker.jakt#L1432

awesomekling:
This issue no longer reproduces :)

PR #1232, opened 2022-10-18
fix Issue 1226

wtholliday:
Fully qualify the names of static functions in case generated code is inside a namespace (such as in an imported module).

See #1226.


tests/codegen/issue_1226.jakt
```diff
@@ -0,0 +1,8 @@
+/// Expect:
+/// - output: "PASS\n"
+
+import issue_1226_sub
+
+function main() {
+   println("PASS")
```
jntrnr, line 7: For the test, should this be calling the static functions, or do you need to not call the static function to cause the issue?

wtholliday: Don't need to call the static function. The test will fail without the change.

wtholliday:
Thanks @jntrnr !

PR #1236, opened 2022-10-20
typechecker: Hide type hints for explicitly typed enum fields

networkException:
This pull request disables dumping of type hints for enum fields that have an explicit type, removing hints that duplicate type information.

Before

image

After

image

Note that this depends on #1234


jntrnr:
@networkException #1234 is in now if you want to rebase

PR #1239, opened 2022-11-01
Typechecker: Error on return with no value in function returning a value

Itai-Nelken:
Previously this was reported by the c++ compiler.

I was wondering if the hint necessary because it feels a bit repetitive to me because it uses the same span as the error.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -4112,6 +4112,17 @@ struct Typechecker {
             .error("‘return’ is not allowed inside ‘defer’", span)
         }
         if not expr.has_value() {
+            if .current_function_id.has_value() {
+                let current_function = .get_function(.current_function_id!)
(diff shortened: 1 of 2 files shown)


selfhost/typechecker.jakt
```diff
@@ -4112,6 +4112,17 @@ struct Typechecker {
             .error("‘return’ is not allowed inside ‘defer’", span)
         }
         if not expr.has_value() {
+            if .current_function_id.has_value() {
```
jntrnr, line 4115: Nice fix.

Can you also add a test so that we can make sure this continues to work in the future?

Itai-Nelken: I just realized that this breaks the compiler. I'm trying to find what condition I missed.

Itai-Nelken: I added the test in the last force push.

PR #1241, opened 2022-11-05
Import path improvements: Add `relative` keyword

demizer:
This change adds the `relative` keyword to allow importing modules relative to the current compiler file.

- `import relative foo::bar` for sub paths
- `import relative parent::foo::bar` for parent paths
- `import relative parent::parent::parent::foo::boo`

I took some “artistic liberty” and added some syntactic sugar to reduce repetition

- `import relative parent(3)::foo:baz` will go up three parent paths from the source file where the import is included.
- `import relative parent()::foo::bar` will go up one parent path, `()` is optional in this instance.

This support also required adding `realpath_()` to the platform module, so it is now possible to get the absolute path of a file in the compiler.

**A note to reviewers**: Please provide extra scrutiny with respect to the unsafe behavior as well as the compiler error messages as these instances are well out of my domain of expertise.


README.md
```diff
@@ -68,6 +68,35 @@ Some of the features that encourage more readable programs:
 - [ ] Trailing closure parameters can be passed outside the call parentheses.
 - [ ] Error propagation with `ErrorOr<T>` return type and dedicated `try` / `must` keywords.
 
+## Code reuse
+
+Jakt is flexible in how a project can be structured with a built-in module system.
```
alimpfard, line 73: Should this also mention `import fn()`?

README.md
```diff
@@ -68,6 +68,35 @@ Some of the features that encourage more readable programs:
...
+
+```jakt
+import a                                (1)
```
alimpfard, line 76: Let's put these `(1)` in comments (`// (1)`)

README.md
```diff
@@ -68,6 +68,35 @@ Some of the features that encourage more readable programs:
...
+import relative foo::bar                (3)
+import relative parent::foo::baz        (4)
+import relative parent(3)::foo::baz     (5)
+```
+
+1. Import a module from the same directory as the file.
+1. Import only `use_cool_things()` from module `a`.
+1. Import a module using the relative keyword when the module is a sub path of the directory containing the file.
+1. Import a module in a parent path one directory up from the file.
+1. Syntactic sugar for importing a module three parent paths up from the file.
```
alimpfard, line 87: :thinking: does this actually show up as a numbered list? looks goofy lol

selfhost/windows_path.jakt
```diff
@@ -1,2 +1,25 @@
...
+    extern function _fullpath(anon absolute_path: raw c_char, anon relative_path: raw c_char, anon max_length: i32) -> raw c_char
+}
+
+function realpath_(anon path: String) throws -> String {
+    mut builder = StringBuilder::create()
+    unsafe {
+        mut absolute_path: raw c_char = allocate<c_char>(count: 256)
+        let buffer = _fullpath(absolute_path, path.c_string(), 256)
+        if buffer == null<c_char>() {
+            // FIXME: throw an error, errno is posix!
```
alimpfard, line 19: Errno here is converted from the windows category of errors into posix equivalents (courtesy of c++'s stdlib lol), I don't think the fixme is necessary.

alimpfard (commented):
Unsafe stuff LGTM.

`parent(N)` looks like it could conflict with `import fn()`, but it might be just fine if it looks like a keyword in the editor?

Alternatively, ban `import relative fn()`?

selfhost/windows_path.jakt
demizer: Oh sweet. I will remove the fixme. Thanks!

README.md
demizer: It does! Markdown is a little goofy. :D

README.md
demizer: Good idea.

README.md
demizer: I will have to play around with this, I am a little naive about `import fn()`.

demizer:
> `parent(N)` looks like it could conflict with `import fn()`, but it might be just fine if it looks like a keyword in the editor?

`parent(n)` only works if it is followed by `::`.

PR #1244, opened 2022-11-07
error: Move error label to end of span

jntrnr:
If we want it, here's a PR that will put the label at the end of the error span rather than the beginning.

Example:

image


jntrnr:
A few folks seem to like this one, so I'm going to land. It should be pretty easy to tweak if we find we want something slightly different.

PR #1250, opened 2022-11-10
editors/vscode: Add syntax highlighting for traits

Lenart12:
Now correctly highlights `trait`, `implements` and `requires`.


ADKaster:
This looks pretty good (I think, I haven't tested it yet). If you want to, a similar change in compiler explorer would probably be a good idea so that people using that tool can get pretty highlighting.

Similar to the PR I made a few months ago here https://github.com/compiler-explorer/compiler-explorer/pull/4021

PR #1251, opened 2022-11-11
editors/vscode: Add syntax highlighting for `extern c`

Lenart12:


editors/vscode/syntaxes/jakt.tmLanguage.json
```diff
@@ -784,8 +784,16 @@
           "match": "\\b(as)\\b\\s"
         },
         {
-          "name": "keyword.control.other.jakt",
-          "match": "\\b(import)\\b"
+          "name": "keyword.control.import.jakt",
+          "match": "\\b(import)(?:\\s*)(?:(?<=\\s+)(?:(extern\\s+c)|(extern)))?\\b",
+          "captures": {
+            "2": {
+              "name": "keyword.control.import.extern-c.jakt"
```
alimpfard, line 791: :thinking: that's a valid scope name? I thought they had to match `(\w+\.)+\w+`

editors/vscode/syntaxes/jakt.tmLanguage.json
```diff
@@ -784,8 +784,16 @@
           "match": "\\b(as)\\b\\s"
         },
         {
-          "name": "keyword.control.other.jakt",
-          "match": "\\b(import)\\b"
+          "name": "keyword.control.import.jakt",
+          "match": "\\b(import)(?:\\s*)(?:(?<=\\s+)(?:(extern\\s+c)|(extern)))?\\b",
```
alimpfard, line 788: Why the unbounded backwards assertion? that's technically not valid in ecma262
```
\b(import)(?:\s*)(?:(\bextern\s+c)|(\bextern))?\b
```
sounds just fine

editors/vscode/syntaxes/jakt.tmLanguage.json
Lenart12: There is already scopes like `fat-arrow` and `return-throws`. If that is true we should fix all of them

alimpfard: Oh, if it works then it's alright, it's just a name

editors/vscode/syntaxes/jakt.tmLanguage.json
Lenart12: Changed to your propsal :)

PR #1252, opened 2022-11-11
jakttest: Don't spin while polling child processes

### LINKED ISSUE #1228: jakttest: Spins at ~100% CPU on Linux while tests are running
A quick `strace` shows us that we're aggressively doing `wait` syscalls:

```
wait4(-1, 0x7ffcf9e0658c, WNOHANG, NULL) = 0
wait4(-1, 0x7ffcf9e0658c, WNOHANG, NULL) = 0
wait4(-1, 0x7ffcf9e0658c, WNOHANG, NULL) = 0
wait4(-1, 0x7ffcf9e0658c, WNOHANG, NULL) = 0
wait4(-1, 0x7ffcf9e0658c, WNOHANG, NULL) = 0
wait4(-1, 0x7ffcf9e0658c, WNOHANG, NULL) = 0
wait4(-1, 0x7ffcf9e0658c, WNOHANG, NULL) = 0
wait4(-1, 0x7ffcf9e0658c, WNOHANG, NULL) = 0
...
```

This should definitely not be necessary, so something here needs a rework.

weltschildkroete:
This patch adds a `sigwait` call before each `poll_running_tests` while the last tests run in the scheduler so that we don't spin calling `waitpid` in a loop.

Fixes https://github.com/SerenityOS/jakt/issues/1228.

### DIFF
--- a/jakttest/jakttest.jakt
+++ b/jakttest/jakttest.jakt
@@ -660,8 +660,24 @@ struct TestScheduler {
             unsafe { cpp { "fflush(stderr);" }}
         }
 
-        while not scheduler.running_tests.is_empty() {
-            scheduler.poll_running_tests()
+	// NOTE: wait for SIGCHLD on non win32, same as in `wait_for_free_dictionary`.
+        unsafe {
+            cpp { "#ifndef _WIN32" }
+            cpp { "sigset_t set;" }
+            cpp { "int signal;"   }
+            cpp { "sigemptyset(&set);" }
+            cpp { "sigaddset(&set, SIGCHLD);" }
+            cpp { "#endif" }
+
+            while not scheduler.running_tests.is_empty() {
+                unsafe {
+                    cpp { "#ifndef _WIN32" }
+                    cpp { "if (sigwait(&set, &signal) > 0)" }
+                    cpp { "    return Error::from_errno(errno);" }
+                    cpp { "#endif" }
+                }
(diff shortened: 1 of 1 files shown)


cybergsus:
The runtime error you're getting is an 'interrupted syscall' error, which probably happens when calling `waitpid`. Probably ignoring the `EINTR` that you're getting from [`waitpid`](https://linux.die.net/man/3/waitpid) would suffice to remove the hanging.

Problem aside, I think this might not be the main issue. Reason is that the main Jakt process only calls `waitpid` with `WNOHANG` repeatedly to pop one by one all the children that have exited in case someone else exits when it's checking another child's exit. When `waitpid` errors with `ECHILD`, it goes back to the "wait for free directory" part where it (at least under POSIX) tells the OS to only wake up on SIGCHLD (see [`man 3 sigwait`](https://man7.org/linux/man-pages/man3/sigwait.3.html)): https://github.com/SerenityOS/jakt/blob/520128b89e433f4a6cc81681ca4edc0856314007/jakttest/jakttest.jakt#L575-L587.

cybergsus:
yeah, since it's just waiting it would be better to let stuff run instead of staying there. Good catch, thanks!

alimpfard:
Can we just use the process stuff that the compiler itself is using? not sure how this is used, but there's a process pool abstraction in there that might be useful too.

PR #1259, opened 2022-11-13
parser+test: Fix precedence of ranges

jntrnr:
This changes the parsing rules for ranges, which lets them have a more correct precedence when used in more situations. The common one this will fix is when a range is used in a match separated by `|` to say "one of these ranges match".

Example:
```
match abc {
    100..200 | 300..400 => println("success")
    else => println("failure")
}
```

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -1967,8 +1967,8 @@ struct CodeGenerator {
 
             yield output
         }
-        else => {
-            todo(format("codegen_expression else: {}", expression))
+        Garbage(span) => {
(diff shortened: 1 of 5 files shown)


alimpfard (approved):
LGTM!

alimpfard:
...is what I would normally say, but looks like we have some non-"operand"s in the codebase...

jntrnr:
@alimpfard ahh looks like we'll need to stop treat or like an operator at least somewhat.

Will try to take another look in a bit but if other folks want to have a try please do.

jntrnr:
Okay, for ranges expressions as an endpoint will need parens. Otherwise we just assume it's an operand. Had to update a couple places in selfhost but hopefully that fixes it.

jntrnr:
...and we're green. Going ahead and landing. cc @awesomekling

PR #1266, opened 2022-11-20
codegen: Fix codegen for optional chaining with classes

mateusauler:
The operator for accessing members of classes is `->`, since they are of type NonnullRefPtr. But when mapping an optional for chaining, the operator generated was the default `.`.

My code seems a bit weird. Any pointers on how to make it look better are appreciated.

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -1535,7 +1535,8 @@ struct CodeGenerator {
             output += object
             output += ")"
 
-            match .program.get_type(expr.type()) {
+            let expression_type = .program.get_type(expr.type())
+            match expression_type {
                 RawPtr => {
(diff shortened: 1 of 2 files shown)


lanmonster (commented):
You can have a variable that starts with ‘“.”’ and only update it in the certain scenario and remove all of the else branches. Also, you could use if bindings and reduce some rightward drift.

mateusauler:
> You can have a variable that starts with ‘“.”’ and only update it in the certain scenario and remove all of the else branches. Also, you could use if bindings and reduce some rightward drift.

Done! But I couldn't use the if bindings with multiple `is` matches `if a is B(id) or a is C(id)`. In this case `id` has type `void`. Is that expected?

PR #1268, opened 2022-11-22
main+lib: Add -fi option to format inplace for files and directories

robryanx:


runtime/jaktlib/file_iterator.jakt
```diff
@@ -0,0 +1,66 @@
...
+                return .next()
+            }
+
+            let path = Path::from_string(builder.to_string())
+            if path.extension() == .extension {
+                return path.to_string()
+            } else {
+                return .next()
+            }
+        } else if .directory_list.size() != 0 {
```
alimpfard, line 58: nit:
```suggestion
        } else if not .directory_list.is_empty() {
```

runtime/jaktlib/file_iterator.jakt
```diff
@@ -0,0 +1,66 @@
...
+            .current_directory = is_directory(path: .directory_list.pop()!)
```
alimpfard, line 59: `is_directory` gives you a `DirectoryIterator`...?

selfhost/posix_fs.jakt
```diff
@@ -35,3 +47,49 @@ function current_directory() throws -> String {
...
+            return None
+        }
+
+        mut builder = StringBuilder::create()
+        mut is_directory = false
+
+        unsafe {
+            builder.append_c_string((*file).d_name)
+
+            if (*file).d_type == 4 {
```
alimpfard, line 70: Please leave a comment here about this constant; we should really think up a way to import #defines sooner rather than later

runtime/jaktlib/file_iterator.jakt
```diff
@@ -0,0 +1,66 @@
...
+
+import path { Path, get_path_separator }
+import os { platform_fs, platform_module, Target }
+
+import platform_fs() {
+    is_directory
+    DirectoryIterator
+}
+
+class RecursiveFileIterator implements(ThrowingIterable<String>) {
```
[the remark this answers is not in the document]
robryanx: Not sure what you mean here?

selfhost/posix_fs.jakt
robryanx: Do we want constants or similar in the language? This will probably come up a lot with serenity interop.

Might be clearer atm just to do something like:
```
cpp {
    "if (*(file).d_type == DT_DIR)"
    "is_directory = true"
}
```

runtime/jaktlib/file_iterator.jakt
robryanx: Maybe it could be better named, the idea being if it is a directory it will return a `DirectoryIterator` if not it will return `None`. So that the one `opendir` call can be used both to test if it is a directory and iterate over it if it is. I will fix the windows version as it would return `(filename, is_directory)` currently instead on `None` if it is a file and not a directory.

runtime/jaktlib/file_iterator.jakt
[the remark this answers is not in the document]
alimpfard: I meant `implements(ThrowingIterable<Path>)`

runtime/jaktlib/file_iterator.jakt
alimpfard: Could this just be `DirectoryIterator::from_path()`?

selfhost/posix_fs.jakt
alimpfard: I think I'm leaning towards "yes, we want constants", but extern-importing them is pretty much required :shrug:

runtime/jaktlib/file_iterator.jakt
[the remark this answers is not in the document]
robryanx: Sure, I changed it to pass around Paths rather than Strings.

runtime/jaktlib/file_iterator.jakt
robryanx: Sure, changed that.

selfhost/posix_fs.jakt
robryanx: Both of the constant cases used cpp and added a fixme for extern importing them.

PR #1270, opened 2022-11-23
all: Move platform specific functions to the standard library

robryanx:
This provides better organisation for the platform specific code. Also most non trivial Jakt programs will need this to support cross platform.

This depends on #1269, I have split them out so if we don't want to make this change can still merge the bug fix.


runtime/jaktlib/platform.jakt
```diff
@@ -38,8 +36,16 @@ function platform_import_names() throws -> [String] {
...
+    mut output: [String] = []
+    for str in strings {
+        output.push(prefix + str + suffix)
+    }
+    return output
+}
+
 comptime platform_module(anon name: String) throws -> [String] {
-    return append_to_each(platform_import_names(), suffix: format("_{}", name))
+    return add_to_each(platform_import_names(), prefix: "jakt::platform::", suffix: format("_{}", name))
```
alimpfard, line 48: This seems a bit odd to me, you'd expect
```
import platform_module("foo")
```
to read "import platform-dependent module foo" and not "import platform-dependent _standard_ module foo"
perhaps this should accept something like
```
import platform_module("jakt::platform::foo")
```
and transform it into `jakt::platform::something_foo`, which would allow the user to also say `import platform_module("foo")` and get `posix_foo`.

selfhost/repl.jakt
```diff
@@ -2,10 +2,11 @@ import typechecker { Typechecker, Interpreter, LoadedModule, ModuleId, ScopeId,
 import compiler { Compiler, FileId }
 import lexer { Lexer }
 import parser { Parser }
-import utility { Span, allocate, null }
+import utility { Span }
 import error { JaktError }
 import interpreter { value_to_checked_expression }
-import path { Path }
+import jakt::path { Path }
+import jakt::platform::utility { allocate, null }
```
alimpfard, line 9: this feels like an implementation detail, not a fan of leaking it like this

runtime/jaktlib/platform.jakt
robryanx: Yeah that makes sense, have updated it to work like that.

selfhost/repl.jakt
robryanx: Yeah, I have changed this back. Maybe in future we need exports or a way to make something private to a module.

Issue #1063, opened 2022-08-09
Iterable type is unknown when passing an iterable as a generic parameter


robryanx:
For example
```
function foo<T>(anon xs: T) {
    for x in xs.iterator() {
        println("{}", x)
    }
}

function main() {
    let arr = [0; 5]

    foo(arr)
}
```
This happens both with builtin array/arrayslice and a user defined iterable.

I think the intention is to handle it here
```
match iterable_type {
            TypeVariable => {
                // Since we're not sure, just make it mutable.
                iterable_should_be_mutable = true
            }
```

The issue being currently that typechecking the `iterable_expr` currently gives an unknown type as when typechecking the Method call
```
let parent_id = match .get_type(checked_expr_type_id) {
```
It doesn't match on `TypeVariable`. If it did I am unsure what `parent_id` to return.

PR #1274, opened 2022-11-28
typechecker: Set call return type to Unknown when unresolved

robryanx:
Previously the return type would be Void which would cause a cpp error when for example trying to assign to a void type. With Unknown auto will be used in the generated code.

This fixes: #1063 and makes it possible to iterate on a generic type.

### DIFF
--- a/samples/generics/iterator.jakt
+++ b/samples/generics/iterator.jakt
@@ -1,0 +1,18 @@
+/// Expect:
+/// - output: "1\n2\n3\nA\nB\nC\n"
+
+function foo<T>(anon xs: T) {
+    for x in xs.iterator() {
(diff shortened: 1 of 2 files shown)


samples/generics/iterator.jakt
```diff
@@ -0,0 +1,18 @@
+/// Expect:
+/// - output: "1\n2\n3\nA\nB\nC\n"
+
+function foo<T>(anon xs: T) {
```
alimpfard, line 4: :thinking: I wonder if we should require `foo<T requires(Iterable<U>)>` to allow iteration?
This is fine for now since that requirement doesn't actually fix the issue, and we still have template-y generics

Issue #752, opened 2022-06-25
Codegen: Implicit conversion to optionals within tuples


greytdepression:
```jakt
function test() -> (bool?) {
    return (false, )
}

function main() {
    println("{}", test())
}
```
In the above function, the Jakt compiler accepts `(false, )` as a valid `(bool?)` tuple, as it allows implicit conversion from `bool` to `bool?`. However, the generated C++ code does not compile and instead produces the following error:
```
error: no viable conversion from returned value of type 'Tuple<bool>' to function return type 'Tuple<Optional<bool>>'
```

PR #1275, opened 2022-11-28
typechecker: Wrap values for optional inner tuple types in `Some`

robryanx:
Previously there would be a cpp error for example with:
```
let foo: (bool?, bool?) = (false, false)
```
It required
```
let foo: (bool?, bool?) = (Some(false), Some(false))
```
This change automatically wraps the values when typechecking.

Fixes: #752


alimpfard:
Nice!

Issue #1148, opened 2022-09-03
Builtins::Array has no operator+, but jakt accepts it anyway


ADKaster:
This causes a compiler error in C++

jakt:
```
let a = ["hello", "world"]
let b = ["sup", "everyone"]
let c = a + b
```
Expected result: compile error, or creating a new array that is the concat of a and b
Result: C++ error like so
```
error: invalid operands to binary expression ('const JaktInternal::Array<String>' and 'const JaktInternal::Array<String>')
```

PR #1276, opened 2022-11-29
codegen+runtime: Add array concatenation via `+` and `+=` operators

robryanx:
Resolves: #1148

### DIFF
--- a/runtime/Builtins/DynamicArray.h
+++ b/runtime/Builtins/DynamicArray.h
@@ -236,6 +236,11 @@ public:
         return array;
     }
 
+    ErrorOr<void> operator+=(DynamicArray<T> const& b)
+    {
+        return push_values(b);
+    }
+
     bool is_empty() const { return m_storage->is_empty(); }
     size_t size() const { return m_storage->size(); }
     size_t capacity() const { return m_storage->capacity(); }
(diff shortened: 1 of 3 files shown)


runtime/Builtins/Array.h
```diff
@@ -224,6 +224,11 @@ class Array {
         return array;
     }
 
+    ErrorOr<void> operator+=(Array<T> const& b)
```
ADKaster, line 227: I'm not sure about this one. The canonical binary operators are supposed to return T or T&, not void.

https://en.cppreference.com/w/cpp/language/operators

This should probably return *this; via `ErrorOr<Array<T&>>`

I'm surprised it's valid to return ErrorOr from these, it looks quite strange. But if it compiles and works then 🤷‍♂️

robryanx: Can't seem to make that one work:
```cpp
ErrorOr<Array<T>&> operator+=(Array<T> const& b)
{
    TRY(push_values(b));
    return *this;
}
```
Gives me:
```
conversion from 'JaktInternal::Array<Jakt::String>' to 'ErrorOr<Array<Jakt::String> &>' is ambiguous
```

Having a look in serenity there are no cases where a += operator uses ErrorOr, there is ByteBuffer though which has
```cpp
void operator+=(ByteBuffer const& other)
{
    MUST(try_append(other.data(), other.size()));
}
```

AtkinsSJ:
This has conflicts.

robryanx:
> This has conflicts.

Yeah, will go through and rebase my prs this weekend.

runtime/Builtins/Array.h
trflynn89: Just happened to be perusing issues this morning - FYI we can't store a reference in an `ErrorOr`. The underlying storage is a `Variant`, which itself can't hold references. I'm surprised this didn't hit an error on this constraint:
https://github.com/SerenityOS/serenity/blob/5b4818df22769bc3353c33b1f1a8d299d96e6efc/AK/Variant.h#L225

robryanx: Any better way that you know of or is how the PR currently has it best?

trflynn89: I'm not familiar with the context, but I would suggest avoiding operator overloading entirely, and use a fallible named method instead.

PR #1277, opened 2022-12-01
lib: Add Sorter

robryanx:


runtime/jaktlib/sorter.jakt
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) 2022, Rob Ryan <rob@robryan.me>
+//
+// SPDX-License-Identifier: BSD-2-Clause
+
+enum SortType {
+    Asc
+    Desc
```
networkException, line 7: I don't think there's a reason to abbreviate those

```suggestion
    Ascending
    Descending
```

networkException (changes requested):

alimpfard:
Not going one way or another yet, but future impl could look nicer imo:
```
trait Sortable {
    function sort(mut this, direction: SortDirection) -> void
    function sorted(this, direction: SortDirection) -> Self { ... } // Maybe
}

type Array implements(Sortable) {
    function sort(mut this, direction: SortDirection) -> void { your_impl_here }
}

mut foo = blah.blah()
foo.sort()
foo.sorted()
```

though this would depend on
- extension trait impls on imported classes (Array isn't implemented in Jakt...) OR implementing the method in C++
- Self type in traits (can merge now if needed)
- figuring out the hierarchy of these traits (copy Rust probably lol)

robryanx:
> 



> Not going one way or another yet, but future impl could look nicer imo:
> 
> ```
> trait Sortable {
>     function sort(mut this, direction: SortDirection) -> void
>     function sorted(this, direction: SortDirection) -> Self { ... } // Maybe
> }
> 
> type Array implements(Sortable) {
>     function sort(mut this, direction: SortDirection) -> void { your_impl_here }
> }
> 
> mut foo = blah.blah()
> foo.sort()
> foo.sorted()
> ```
> 
> though this would depend on
> 
>     * extension trait impls on imported classes (Array isn't implemented in Jakt...) OR implementing the method in C++
> 
>     * Self type in traits (can merge now if needed)
> 
>     * figuring out the hierarchy of these traits (copy Rust probably lol)

Yeah, I was trying to figure this out recently for enumerate. I think certainly the trait impls on imported classes will be nice to have to avoid the original issue of an ever expanding prelude and runtime.

runtime/jaktlib/sorter.jakt
robryanx: Yeah that makes sense, this isn't sql. Updated it.

PR #1282, opened 2022-12-08
selfhost+tests: Fix a crash when trying to typecheck generic call

awesomekling:
...in the presence of previous type errors that made the call impossible to specialize.

When this happens, we will already have queued up errors about the type errors, so we just bail out of the generic typecheck and let said errors bubble to the user.


alimpfard (approved):
I had a fix similar to this in #1281, but let's get this in instead of waiting for that PR to get to a merge-able state.

PR #1293, opened 2022-12-23
parser: Fix infinite loop when parsing generic type

### LINKED ISSUE #1287: Cast array.size() in if comparison spins at 100% CPU
To reproduce:

```
function main() {
  let arr: [u8] = []
  if arr.size() as! i64 < 0 { }
}
```

SimonRask:
Fixes #1287

### DIFF
--- a/selfhost/parser.jakt
+++ b/selfhost/parser.jakt
@@ -3460,26 +3460,28 @@ struct Parser {
             }
         }
         Identifier(name) => {
-            let span = .current().span()
+            let name_span = .current().span()
             .index++
-            mut parsed_type =  ParsedType::Name(name, span)
+            mut parsed_type = ParsedType::Name(name, span: name_span)
             if .current() is LessThan {
+                let generic_start_span = .current().span()
+                .index++
                 mut params: [ParsedType] = []
-                if .current() is LessThan {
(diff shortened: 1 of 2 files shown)


alimpfard (commented):
Could you also add a test to make sure this doesn't break in the future?

SimonRask:
I have now added a test for the reduced version of the code provided in #1287.

Im not sure if the code provided should be valid, but the test ensures that the parser doesn't stall for that case.

selfhost/parser.jakt
```diff
@@ -3398,7 +3398,13 @@ struct Parser {
                 if .current() is LessThan {
                     .index++
                     while not .current() is GreaterThan and not .eof() {
-                        params.push(.parse_typename())
+                        let parsed_type_parameter = .parse_typename() 
+                        if parsed_type_parameter is Empty {
+                            .error("Expected type inside generic", .current().span())
```
alimpfard, line 3403: :thinking: Could we be a bit more helpful with this error? something like
> Error: Expected a type here (pointing after the `<`)
> Hint: This type forms a generic instantiation (pointing at the `i64 <` range)

alimpfard:
Also, please rebase on top of main, CI will be unhappy otherwise.

PR #1295, opened 2022-12-26
Simplify error display + some fixes (including a fixed panic)

0GreenClover0:
I've been hunting a weird panic that I've stumbled upon. Turned out it was a logic error in the display_message_with_span in error.jakt. While being there I've also fixed some other issues.

- 0081da911eec952ee0468c794d1ebaa9a3969294 we didn't properly print the span of the errors that occurred in first line of the source code

test.jakt:
```cpp
function main() { return y }
```
Previously:
```
greenclover@CloverLaptop:~/jakt$ ./build/bin/jakt ./own/test.jakt 
Error: Variable 'y' not found
───┴─
```
Now:
```
greenclover@CloverLaptop:~/jakt$ ./build/bin/jakt ./own/test.jakt 
Error: Variable 'y' not found
───┬─ ./own/test.jakt:1:26
 1 │ function main() { return y } 
   │                          ╰─ Variable 'y' not found
───┴─
```

- 865cb956ab2ad51bf2615d953eb0fb4e8f7bc16c we were not gathering a span of the last line of the file, if the line was empty. Seems like a bug and if not fixed, it would cause issues in the next commit

- f6e20b44929eb7dbd650449150a89be0510ae05f we were calculating the largest line number that we will display based on the ***start*** of the error span. The span could be several lines long tho, and we will display all of them, so the largest line number could be off. If it was off by enough line numbers, that the actual largest line number had more digits than we had expected, it later led on to the panic in the unsigned integers subtraction

- 229d5d16a784358c4aa478148b07d2c1716510d6 by calculating index of the first error line, we can later omit a `while` loop and an `if` inside it, which removes two levels of indentation. I've also added some comments

- b3f28c3a39f8529a679b795657c4de2202f766f9 previously, we gave return statements that return void (so they end with an `eof`, `eol` or `}`) a span that consists of the next token (`eof`, `eol`, `}`) instead of the span of the `return` keyword like we do if we return a non-void value (in this case we also include a span of the returned expression, but when returning void, there is no expression). This causes some oddities when displaying an error message of the error that happened in this statement

test.jakt:
```cpp
function foo(anon x: i64) -> i64 {
    return 
}
function main() { }
```

Previously:
```
greenclover@CloverLaptop:~/jakt$ ./build/bin/jakt ./own/test.jakt 
Error: ’return’ with no value in function ’foo’ returning ’i64’
───┬─ ./own/test.jakt:2:12
 1 │ function foo(anon x: i64) -> i64 { 
 2 │     return  
   │╰─ ’return’ with no value in function ’foo’ returning ’i64’
 4 │ function main() { } 
───┴─
```

Now:
```
greenclover@CloverLaptop:~/jakt$ ./build/bin/jakt ./own/test.jakt 
Error: ’return’ with no value in function ’foo’ returning ’i64’
───┬─ ./own/test.jakt:2:5
 1 │ function foo(anon x: i64) -> i64 { 
 2 │     return  
   │          ╰─ ’return’ with no value in function ’foo’ returning ’i64’
 3 │ } 
───┴─
```

### DIFF
--- a/selfhost/error.jakt
+++ b/selfhost/error.jakt
@@ -73,7 +73,7 @@ function display_message_with_span(anon severity: MessageSeverity, file_name: St
     mut line_index = 0uz
     mut largest_line_number = 0uz
     while line_index < line_spans.size() {
-        if span.start >= line_spans[line_index].0 and span.start <= line_spans[line_index].1 {
+        if span.end >= line_spans[line_index].0 and span.end <= line_spans[line_index].1 {
             largest_line_number = line_index + 2 // 1 (row number) + 1 (extra source line)
         }
         ++line_index
--- a/selfhost/error.jakt
+++ b/selfhost/error.jakt
@@ -70,8 +70,7 @@ function display_message_with_span(anon severity: MessageSeverity, file_name: St
     let file_contents = contents!
     let line_spans = gather_line_spans(file_contents)
 
-    mut line_index = 1uz
-
+    mut line_index = 0uz
     mut largest_line_number = 0uz
     while line_index < line_spans.size() {
         if span.start >= line_spans[line_index].0 and span.start <= line_spans[line_index].1 {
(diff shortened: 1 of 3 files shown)


ADKaster:
These changes look reasonable overall. Could you add some tests to make sure that the specific cases you're fixing don't regress in the future?

0GreenClover0:
> These changes look reasonable overall. Could you add some tests to make sure that the specific cases you're fixing don't regress in the future?

I've added a test for f6e20b44929eb7dbd650449150a89be0510ae05f.
Can't really add tests for other issues as they don't affect _what_ error is displayed, but _how_ it is displayed in the console

PR #1302, opened 2022-12-30
parser: Fix possible out-of-bounds array access in ParsedExpr::equals() for ParsedExpr:: Array, Dictionary, Set and Tuple

ge0mk:
A typo / missing parenthesis when comparing the size of both arrays caused an out-of-bounds array access during element-wise comparison.

### DIFF
--- a/selfhost/parser.jakt
+++ b/selfhost/parser.jakt
@@ -1108,7 +1108,8 @@ boxed enum ParsedExpression {
                         return false
                     }
                 }
-                if not lhs_values.size() == rhs_values.size() {
+
+                guard lhs_values.size() == rhs_values.size() else {
                     return false
                 }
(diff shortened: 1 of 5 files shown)


selfhost/parser.jakt
```diff
@@ -1108,7 +1108,8 @@ boxed enum ParsedExpression {
                         return false
                     }
                 }
-                if not lhs_values.size() == rhs_values.size() {
```
alimpfard, line 1111: :thinking: interesting that this is allowed here, the typechecker definitely disallows this:
```
> function a() => 42uz
> function b() => 42uz
> not a() == b()
Error: Invalid type for unary operator
```

yet the same expression is allowed in "normal" code...

PR #1306, opened 2022-12-31
typechecker: Properly track control flow when dealing with an "if... else" block

### LINKED ISSUE #1152: Guard followed by if statement allows non-void function to return nothing
This causes Jakt to emit a non-void C++ function that does not return a value in all code paths, which leads to the C++ compiler emitting only a warning.

Jakt code:
```
function test(argument: u32) -> bool {
	guard argument > 30 else {
		return false
	}

	if argument < 10 {
		return true
	}
}
```

Expected result: Jakt compiler error
Actual result: Compiles with only the C++ warning:
```
warning: non-void function does not return a value in all control paths [-Wreturn-type]
```

Note that if the `if` statement is removed, the compiler behaves correctly.

0GreenClover0:
Previously, when control flow of an `if` block was determined to be one of:
`PartiallyNeverReturns`, `PartialAlwaysReturns`, `PartialAlwaysTransfersControl`, `AlwaysTransfersControl`,
and an `else` block was present, we were basically completely ignoring the control flow of an `if` block and going with what the `else` block control flow was telling us.

Every test with the name which doesn't contain "continue_case" was not working before this PR. "continue_case" tests were working, but I still added them so they won't break in the future.
There is a test added for every case added to the match in typechecker, other than when `if` block has control_flow of either `PartialAlwaysReturns` or `PartialNeverReturns`, and the `else` block has any of the `Partial...` control_flows. I couldn't really find tests for these cases, because when there is a `break` or `continue` inside of them, `unity_with()` will promote them to `PartialAlwaysTransfersControl`. And if there is no `break` or `continue` inside of them, then they don't really matter all that much (if the `else` block also is one of the Partials) as far as I'm aware 🤷.

Fixes #1152. Not sure if a specific test with a `guard` is needed, since `guard`s are expanded to `if`s anyway.

Plan for the future:
- Move the whole `if... else` control flow resolving to its own function in the `types.jakt` and potentially use it for the `match` (the control flow for `match`es is currently also not properly tracked)

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -3638,7 +3638,15 @@ struct Typechecker {
                     PartialAlwaysReturns(might_break) => BlockControlFlow::PartialAlwaysReturns(might_break: might_break or then_might_break)
                     PartialAlwaysTransfersControl(might_break) => BlockControlFlow::PartialAlwaysTransfersControl(might_break: might_break or then_might_break)
                 }
-                PartialAlwaysReturns | PartialAlwaysTransfersControl | AlwaysTransfersControl => .maybe_statement_control_flow(else_statement, then_block.control_flow)
+                PartialAlwaysReturns(might_break: then_might_break) => match .maybe_statement_control_flow(else_statement, then_block.control_flow) {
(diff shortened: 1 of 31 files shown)


alimpfard:
Some of the test cases were a real headscratcher, but they _seem_ alright :P

PR #1310, opened 2023-01-01
typechecker: Make `while` block control flow resolving less definitive + use more info

0GreenClover0:
Previously, control flow of a `while` block was a little bit too definitive. Ex. in:
```cpp
while condition {
    return
}
println("Hello, condition might not be true!")
```
it would be resolved to `AlwaysReturns` and the `println()` would be marked as an "Unreachable code".
But `while` blocks act somewhat similarly to the `if` blocks that have no `else` statement. The **condition** might not be `true` so the result cannot be definitive, only partial.

`While` blocks, however, also consume all the `break`s and `continue`s, and without them, partial results can be downgraded to `MayReturn`s (like we already did before this patch).

This means that if the condition is not known to be always true, no matter what the inner control flow of the `while` block is, we will resolve it to `MayReturn`.

Tests 1 and 2 didn't work before, test 3 worked, but it prevents from leaking the `might_break` value from inner `while` loops to outer loops (which happened to me during some experiments for this PR, and nothing complained). Tests 4 and 5 did work before this PR, but it was only because we were definitive about the control flow result of the `while` blocks, not caring if the condition is always true or not.

### DIFF
--- a/tests/typechecker/control_flow_while_1.jakt
+++ b/tests/typechecker/control_flow_while_1.jakt
@@ -1,0 +1,14 @@
+/// Expect:
+/// - output: "PASS\n"
+
+function test(anon mut argument: i64) -> String {
+    while argument > 0 {
+        return "NOT PASSED" // This should be reachable (in case argument > 0)
+    }
+
+    return "PASS" // This should be reachable as well (in case argument <= 0)
+}
+
+function main() {
+    println("{}", test(-1))
+}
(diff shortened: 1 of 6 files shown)


tests/typechecker/control_flow_while_1.jakt
```diff
@@ -0,0 +1,14 @@
+/// Expect:
+/// - output: "PASS\n"
+
+function test(anon mut argument: i64) -> String {
+	while argument > 0 {
```
alimpfard: lgtm, but could you make these use spaces instead of hard tabs?

0GreenClover0: vscode had a "from now on, i will use tabs instead of spaces" incident, should be fixed now

PR #1314, opened 2023-01-02
parser: Fix comparison of match cases to be order independent & also compare case bodies

ge0mk:
- patterns, that are enum variants now get treated as equal when one is a subset of the other
  -> `Foo::A` is the same as `A`, but not the same as `Bar::A`
- match cases no longer get detected as equal, when they have different bodies
- match cases now get detected as equal, even when they are in different order
- duplicate patterns in a single match expression are now an error

### DIFF
--- a/selfhost/parser.jakt
+++ b/selfhost/parser.jakt
@@ -875,28 +875,28 @@ enum ParsedMatchPattern {
         return true
     }
 
-    function equals(this, anon rhs_parsed_match_pattern: ParsedMatchPattern) -> bool => .defaults_equal(rhs_parsed_match_pattern.defaults) and match this {
-        EnumVariant(variant_names: lhs_variant_names, variant_arguments: lhs_variant_arguments) => match rhs_parsed_match_pattern {
-            EnumVariant(variant_names: rhs_variant_names, variant_arguments: rhs_variant_arguments) => {
-                guard lhs_variant_names.size() == rhs_variant_names.size() and lhs_variant_arguments.size() == rhs_variant_arguments.size() else {
+    function is_equal_pattern(this, anon rhs_parsed_match_pattern: ParsedMatchPattern) -> bool => match this {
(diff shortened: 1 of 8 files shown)


selfhost/parser.jakt
```diff
@@ -905,13 +905,55 @@ enum ParsedMatchPattern {
...
+            guard rhs_parsed_match_pattern is EnumVariant(variant_arguments: rhs_variant_arguments) else {
+                return false
+            }
+
+            guard lhs_variant_arguments.size() == rhs_variant_arguments.size() else {
+                return false
+            }
+
+            for i in ..lhs_variant_arguments.size() {
+                guard lhs_variant_arguments[i].equals(rhs_variant_arguments[i]) else {
```
alimpfard, line 928: Let's just use `if not ...` for this.

selfhost/parser.jakt
```diff
@@ -4980,6 +4980,17 @@ struct Parser {
         let expr = .parse_expression(allow_assignments: false, allow_newlines: true)
         let cases = .parse_match_cases()
 
+
+        if cases.size() > 1 {
+            for i in ..(cases.size() - 1) {
+                for k in (i + 1)..cases.size() {
+                    if cases[i].has_equal_pattern(cases[k]) {
+                        .error("Duplicated match pattern", cases[k].marker_span)
```
alimpfard, line 4988: Please also point to the original case here, something like
```
.error_with_hint(
    "Duplicated match pattern", cases[k].marker_span
    "Original pattern here", cases[i].marker_span
)
```

alimpfard (commented):
Please cleanup the commit messages a bit:
> `parser: ParsedMatchCase::equals() now also compares bodies`

`parser: Make ParsedMatchCase::equals() also compare case bodies`

typo here:
> `parser: Disallow duplicated match patters`

`parser: Disallow duplicated match patterns`

PR #1315, opened 2023-01-02
typechecker: Control flow, don't lose information about `break`s

0GreenClover0:
Very similar to #1306.

When resolving an `if... else` statement, whenever an `if` block was resolved to `MayReturn` we were losing information about the potential `break`s in the `else` block. And the other way around as well, whenever an `else` block was resolved to `MayReturn`, we were losing information about `break`s in the `if` block.

Tests not added for `PartialAlwaysReturns` and `PartialNeverReturns` because of the same reason as in the mentioned PR. I've still changed the behavior for these to the more correct one, in case `unify_with()` changes its behavior.


0GreenClover0:
Changed some formatting in the tests, not sure if this is better (but I hope so)

PR #1316, opened 2023-01-06
README: fix typo

RGBCube:


lanmonster (commented):
I don't think this is a typo. It means that the weak pointer becomes empty when the thing it points to is destroyed.

RGBCube:
oh, i see

PR #1323, opened 2023-01-15
typechecker: Unwrap return in throwing fatarrow functions

### LINKED ISSUE #1305: void fat arrow functions marked as throws fails C++ compilation
I get a compilation error if I have the following source:
```
/// Expect:
/// - output: "OK\n"

function f() throws {
    println("OK")
}
function g() throws => f()

function main() {
    f()
}
```

The error is:
```
Building: 1/3
Building: 2/3
build/Root Module.cpp:5:8: error: no viable conversion from returned value of type 'void' to function return type 'ErrorOr<void>'
return (TRY((f())));
       ^~~~~~~~~~~~
/home/hugh/src/osdev/jakt//runtime/Jakt/Error.h:86:5: note: candidate constructor not viable: cannot convert argument of incomplete type 'void' to 'Jakt::Error' for 1st argument
    ErrorOr(ErrorType error)
    ^
/home/hugh/src/osdev/jakt//runtime/Jakt/Error.h:99:5: note: candidate constructor not viable: cannot convert argument of incomplete type 'void' to 'Jakt::ErrorOr<void> &&' for 1st argument
    ErrorOr(ErrorOr&& other) = default;
    ^
/home/hugh/src/osdev/jakt//runtime/Jakt/Error.h:100:5: note: candidate constructor not viable: cannot convert argument of incomplete type 'void' to 'const Jakt::ErrorOr<void> &' for 1st argument
    ErrorOr(ErrorOr const& other) = default;
    ^
1 error generated.
Building: 3/3
Error: Compilation failed
```

Changing to `function g() throws { f() }` instead of using fat arrow fixes this.

Cheers,

Hugh

mattisboeckle:
Expressions in fatarrow functions are always appended by a Return, which causes Problems in the specific case where a throwing fatarrow function calls a non-throwing, void-returning function.
In this case we can unwrap the return as it is already handeled in codegen.

Fixes #1305

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -3626,7 +3626,7 @@ struct Typechecker {
             scope.can_throw = true
         }
 
-        let block = .typecheck_block(
+        mut block = .typecheck_block(
             parsed_function.block
             parent_scope_id: function_scope_id
             safety_mode: SafetyMode::Safe
@@ -3642,6 +3642,21 @@ struct Typechecker {
         let return_type_id = match function_return_type_id.equals(unknown_type_id()) {
             true => .infer_function_return_type(block)
             else => .resolve_type_var(type_var_type_id: function_return_type_id, scope_id: function_scope_id)
+        }
+
+        // Unwrap return in throwing fat arrow function that returns void
+        // void can't automatically be cast to ErrorOr<void> which crashes c++ compilation
+        if parsed_function.is_fat_arrow and parsed_function.can_throw and return_type_id.equals(void_type_id()) {
+            let statement = block.statements[0]
+            match statement {
+                Return(val, span) => {
+                    let stmt = CheckedStatement::Expression(expr: val!, span: span!)
+                    block.statements = [stmt]
+                }
+                else => {
+                    panic("No return in fat arrow function")
+                }
+            }
         }
 
         if not function_linkage is External and not return_type_id.equals(void_type_id()) and not block.control_flow.always_transfers_control() {
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -3111,7 +3111,7 @@ struct Typechecker {
             scope.can_throw = true
         }
 
-        let block = .typecheck_block(parsed_block: func.block, parent_scope_id: function_scope_id, safety_mode: SafetyMode::Safe)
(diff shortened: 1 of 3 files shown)


selfhost/typechecker.jakt
```diff
@@ -3644,6 +3644,17 @@ struct Typechecker {
...
         }
 
+        if parsed_function.is_fat_arrow and parsed_function.can_throw and return_type_id.equals(void_type_id()) {
+            let statement = block.statements[0]
+            match statement {
+                Return(val, span) => {
+                    let stmt = CheckedStatement::Expression(expr: val!, span: span!)
+                    block.statements = [stmt]
+                }
+                else => println("No return in fat arrow function")
```
0GreenClover0, line 3654: I think we should `panic` here instead

```suggestion
                else => {
                    panic("No return statement in a fat arrow function")
                }
```

selfhost/typechecker.jakt
```diff
@@ -3644,6 +3644,17 @@ struct Typechecker {
             else => .resolve_type_var(type_var_type_id: function_return_type_id, scope_id: function_scope_id)
         }
 
+        if parsed_function.is_fat_arrow and parsed_function.can_throw and return_type_id.equals(void_type_id()) {
```
0GreenClover0, line 3647: This whole block is comment-worthy imo

0GreenClover0:
Also: the whole change also needs to be added to the `typecheck_method()` to fix a sample like this:
```cpp
class Test {
    function f() {
        println("OK")
    }

    public function g() throws => f()
}

function main() {
    Test::g()
}
```

Maintaining both `typecheck_method` and `typecheck_function` is pretty dirty, but it is how it is for now :P

mattisboeckle:
Sounds reasonable, ill add the changes

hughdavenport:
Can confirm this fixes my use case in #1305

tests/typechecker/return_void_in_throwing_fatarrow_function.jakt
```diff
@@ -0,0 +1,11 @@
+/// Expect:
+/// - output: "OK\n"
+
+function f() {
+    println("OK")
+}
+function g() throws => f()
+
+function main() {
+    f()
```
madebr, line 10: Shouldn't this call `g` instead, to better reflect the unit test intention?

PR #1327, opened 2023-01-19
parser: Allow trailing closure arguments to occur after the call's ()

AtkinsSJ:
See `samples/closures/trailing_closure_parameter.jakt` for an example.

As shown by the tests, these are not parsed in the expressions of if/for/guard/match, since that leads to ambiguous syntax and also just looks bad.

This is a very limited first pass. A single block is parsed, and treated as a closure. This means that closure parameters, captures, and the return type (including the throwingness) cannot be defined. Ideally the return type would be inferred based on the closure's contents and the called function's parameter type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Some of the features that encourage more readable programs:
 - [x] None coalescing for optionals (`foo ?? bar` yields `foo` if `foo` has a value, otherwise `bar`)
 - [x] `defer` statements.
 - [x] Pointers are always dereferenced with `.` (never `->`)
-- [ ] Trailing closure parameters can be passed outside the call parentheses.
+- [x] Trailing closure parameters can be passed outside the call parentheses.
 - [ ] Error propagation with `ErrorOr<T>` return type and dedicated `try` / `must` keywords.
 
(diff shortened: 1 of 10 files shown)


robryanx:
It is a bit ambiguous the distinction between the tests and samples directory but I'd put the expected error tests under the tests directory and just leave the success case under samples.

AtkinsSJ:
> It is a bit ambiguous the distinction between the tests and samples directory but I'd put the expected error tests under the tests directory and just leave the success case under samples.

Oops, I really thought that's where I'd put them, but apparently not!

AtkinsSJ:
New changes:
- Put the failing tests in `tests`.
- Implement "reference everything" for captures.
- The closure implicitly captures everything by reference.
- The closure parameter is now passed as a `&function`.

AtkinsSJ:
Another change: Ticked the box on the README page. :^)

Also maybe CI won't flake this time.

alimpfard (approved):
CI should be good now (windows flake should be fixed too)

alimpfard:
Let's get this in, we can improve on it in-tree.

PR #1330, opened 2023-01-24
jakttest: Add `target-triple` option to Jakttest

cybergsus:
On a musl platform Jakttest won't be able to build anything due to
target being set by default to 'x86_64-pc-linux-gnu'. I don't know yet
if the system detection from #defines can do more, but since run_one.py
already supported passing `-T/--target-triple` to the C++ compiler,
Jakttest is now able to forward the argument.

Thanks to this I could get to run the tests correctly, using the clang++
binary provided in the nix derivation.

EDIT: here's a screenshot of it working under void linux, musl:
image

### DIFF
--- a/jakttest/jakttest.jakt
+++ b/jakttest/jakttest.jakt
@@ -147,6 +147,7 @@ function print_usage()  {
     eprintln("\t--temp-dir <DIR>\t\tUse DIR as the temporary directory.")
     eprintln("\t\t\tWill try to use $TMP_DIR or /tmp in UNIX-based systems, or %TEMP% in Windows")
     eprintln("\t-C,--cpp-compiler <PATH>\t\tUse the C++ compiler at PATH to compile objects. Defaults to clang++.")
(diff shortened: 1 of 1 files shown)


jakttest/jakttest.jakt
```diff
@@ -633,13 +647,15 @@ struct TestScheduler {
         total_test_count: usize
         build_dir: String
         cpp_compiler_path: String? = None
+        opt_target_triple: String? = None
```
alimpfard, line 650: :thinking: nothing is passing a value to this though?

cybergsus: looks like I reverted changes when trying to move the changes to a separate branch. Lemme fix this

cybergsus: I fixed it and it turns out to work inside the nix shell without any target triple overloading. I'll close this since it's not needed after all.

cybergsus:
It turns out that using the nix derivation already removes the need to overload the target triple for the tests. Will close this since it isn't needed.

PR #1331, opened 2023-01-25
ci: Use clang-15 to build on CI

alimpfard:
Following the serenity CI update, switch to using clang-15.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,10 +27,10 @@ jobs:
       fail-fast: true
       matrix:
         platform: [ubuntu-22.04]
-        cxx_compiler: [clang++-14]
+        cxx_compiler: [clang++-15]
         include:
           - platform: macos-latest
-            cxx_compiler: "$(brew --prefix llvm@14)/bin/clang++"
+            cxx_compiler: "$(brew --prefix llvm@15)/bin/clang++"
 
     runs-on: ${{ matrix.platform }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -43,9 +43,17 @@ jobs:
           version: 1.10.0
           dest: ${{ github.workspace }}/ninja_bin
 
+      - name: Install Linux dependencies
+        if: ${{ matrix.platform == 'ubuntu-22.04' }}
+        run: |
+          wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
```
ADKaster, line 49: Using apt-key add like this is deprecated, this should be something like

```
          . /etc/lsb-release
          sudo wget -O /usr/share/keyrings/llvm-snapshot.gpg.key https://apt.llvm.org/llvm-snapshot.gpg.key
          sudo echo "deb [signed-by=/usr/share/keyrings/llvm-snapshot.gpg.key] http://apt.llvm.org/${DISTRIB_CODENAME}/ llvm-toolchain-${DISTRIB_CODENAME} main" | sudo tee -a /etc/apt/sources.list.d/llvm.list
          sudo echo "deb [signed-by=/usr/share/keyrings/llvm-snapshot.gpg.key] http://apt.llvm.org/${DISTRIB_CODENAME}/ llvm-toolchain-${DISTRIB_CODENAME}-15 main" | sudo tee -a /etc/apt/sources.list.d/llvm.list
```

alimpfard: :shrug: that's a copy of what we have in serenity
switch over there and I'll copy it here :eyes:

ADKaster: I mean... We don't *have* to propagate a "bad" pattern, but if that's how you wanna play it sure, I can update them all at the same time

alimpfard: I'll update this one in this PR then, awaiting your fix for serenity~

alimpfard:
Update: Plus one _deflake_ for windows CI, yay!

PR #1344, opened 2023-02-05
error: Underline error span for better readability

jntrnr:
This adds an underline for a bit better readability of an error span.

Before:
```
Error: Call to unknown function: ‘fooo’
────┬─ andreas.jakt:9:5
 8  │ fn main() {
 9  │     fooo()
    │        ╰─ Call to unknown function: ‘fooo’
 10 │ }
────┴─
```
Now:
```
Error: Call to unknown function: ‘fooo’
────┬─ andreas.jakt:9:5
 8  │ fn main() {
 9  │     fooo()
    │     ───┬
    │        ╰─ Call to unknown function: ‘fooo’
 10 │ }
────┴─
```

A few more examples:
```
Error: Binary arithmetic operation between incompatible types (‘String’ and ‘i64’)
────┬─ testme.jakt:9:11
 8  │
 9  │   let y = x / 3
    │           ────┬
    │               ╰─ Binary arithmetic operation between incompatible types (‘String’ and ‘i64’)
 10 │
────┴─
Error: Type mismatch: expected ‘i64’, but got ‘String’
────┬─ testme.jakt:15:15
 14 │
 15 │   let z = x / 3
    │               ┬
    │               ╰─ Type mismatch: expected ‘i64’, but got ‘String’
 16 │
────┴─
```

### DIFF
--- a/selfhost/error.jakt
+++ b/selfhost/error.jakt
@@ -108,6 +108,7 @@ fn display_message_with_span(anon severity: MessageSeverity, file_name: String, 
     while line_index < line_spans.size() and span.end > line_spans[line_index].0 {
         print_source_line(severity, file_contents, file_span: line_spans[line_index], error_span: span, line_number: line_index + 1, largest_line_number)
         if span.end <= line_spans[line_index].1 {
+            print_underline(severity, width, file_span: line_spans[line_index], error_span: span, line_number: line_index + 1, largest_line_number)
             break
         }
         ++line_index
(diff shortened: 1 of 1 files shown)


selfhost/error.jakt
```diff
@@ -172,6 +173,35 @@ fn print_source_line(severity: MessageSeverity, file_contents: [u8], file_span:
...
+fn print_underline(severity: MessageSeverity, width: usize, file_span: (usize, usize), error_span: Span, line_number: usize, largest_line_number: usize) throws {
+    mut index = file_span.0
+
+    // Print the error label
+    for x in 0..(width + 2) {
+        eprint(" ")
+    }
+    eprint("│ ")
+
+    while index <= file_span.1 {
```
konradekk, line 185: somewhat strange not using a for loop for this seeing one above…! 😝

(itʼd be even nice to have it working with `for index in file_span`)

konradekk (approved):

selfhost/error.jakt
jntrnr: Are you saying you want to to add a feature to the language to iterate over a 2-element tuple as a (start,end) pair?

jntrnr: Oh I think I get what you're saying. You have two different ideas, and the first one is using a range? Yeah, could probably rewrite it to do something like that

jntrnr: Fixed - the tuple idea could be worth a brainstorm, too, for some future updates

konradekk: oh! mixed things too much—happy the message got through! 😝

PR #1346, opened 2023-02-06
typechecker: Argument label validation fixes

### LINKED ISSUE #1326: classes: Can't assign private field in implicit constructor by value
I would expect the following to compile:

```
class A {
    priv: i64
}

function main() {
    let priv = 1
    let a1 = A(priv)
    let a2 = A(priv: 2)
    println("OK")
}
```

I get the following error:
```
Error: No function with matching signature found.
───┬─ test.jakt:8:14
 7 │     let a1 = A(priv)
 8 │     let a2 = A(priv: 2)
   │              ╰─ No function with matching signature found.
 9 │     println("OK")
───┴─
Error: Can't access field 'priv' when calling implicit constructor of 'A' because it is marked private
───┬─ test.jakt:8:16
 7 │     let a1 = A(priv)
 8 │     let a2 = A(priv: 2)
   │                   ╰─ Can't access field 'priv' when calling implicit constructor of 'A' because it is marked private
 9 │     println("OK")
───┴─

```

It doesn't complain for the first one, only when you try to set it inline in the constructor. Workaround is to just let a variable of the same name to pass to the constructor.

Cheers,

Hugh

0GreenClover0:
1. Don't allow overriding the argument label when the label is **required**, but argument name matches the parameter name (as opposed to the fix in 5c0b593 which only fixed situations where the label was explicitly **not required** (it somehow went under my radar)). This is done by using the `get_argument_name()` in `validate_argument_label()`, which returns the label first, if it exists, so if it exists, we later only compare the label with the parameter name during validation

2. When the argument label doesn't match the parameter name and argument name is retrievable, but it also doesn't match the parameter name - don't highlight the **argument name** when printing an error, highlight the **wrong label**. This is a side effect of using `get_argument_name()` in `validate_argument_label()` as described in the commit message

Previously:
```
Error: Wrong parameter name in argument label (got 'variable', expected 'good')
───┬─ ./tests/typechecker/wrong_parameter_name_in_argument_label.jakt:8:15
 7 │     let variable = 2 
 8 │     test(bad: variable) 
   │                      ╰─ Wrong parameter name in argument label (got 'variable', expected 'good')
 9 │ } 
───┴─
```

Now:
```
Error: Wrong parameter name in argument label (got 'bad', expected 'good')
───┬─ ./tests/typechecker/wrong_parameter_name_in_argument_label.jakt:8:10
 7 │     let variable = 2 
 8 │     test(bad: variable) 
   │            ╰─ Wrong parameter name in argument label (got 'bad', expected 'good')
 9 │ } 
───┴─
```
Note the `got 'bad'` instead of `got 'variable'`

---
Previously:
```
Error: Wrong parameter name in argument label (got '', expected 'good')
───┬─ ./tests/typechecker/wrong_parameter_name_in_arg_label.jakt:8:10
 7 │     let variable = 2 
 8 │     test(variable) 
   │                 ╰─ Wrong parameter name in argument label (got 'variable', expected 'good')
 9 │ } 
```
Now:
```
Error: Wrong parameter name in argument label (got '', expected 'good')
───┬─ ./tests/typechecker/wrong_parameter_name_in_arg_label.jakt:8:10
 7 │     let variable = 2 
 8 │     test(variable) 
   │                 ╰─ Wrong parameter name in argument label (got '', expected 'good')
 9 │ } 
```
Note the `got ''` instead of `got 'variable'`

3. While validating the argument visibility in class/struct implicit constructor, take into account that the argument might not have a label if the argument name matches the parameter name (by using the `get_argument_name()` added in the first commit). Resolves #1326

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -547,7 +547,7 @@ struct CodeGenerator {
                 .postorder_traversal(
                     encoded_type_id: entry.0
                     visited: seen_types
-                    dependency_graph: encoded_dependency_graph
+                    encoded_dependency_graph
                     output: traversal
                 )
                 for type_id in traversal {
--- a/selfhost/compiler.jakt
+++ b/selfhost/compiler.jakt
@@ -49,7 +49,7 @@ class Compiler {
                                 file_contents = file.read_all()
                             } catch error {}
                         }
-                        print_error(file_name, contents: file_contents, error)
+                        print_error(file_name, file_contents, error)
                     }
                 }
             }
--- a/selfhost/parser.jakt
+++ b/selfhost/parser.jakt
@@ -2058,7 +2058,7 @@ struct Parser {
                 Identifier(name, span) => {
                     alias.alias_name = ParsedName(
                         name
-                        name_span: span
+                        span
                     )
                     .index++
                 }
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -2634,7 +2634,7 @@ struct Typechecker {
         }
     }
 
-    fn typecheck_enum_constructor(mut this, record: ParsedRecord, enum_id: EnumId, parent_scope_id: ScopeId) throws {
+    fn typecheck_enum_constructor(mut this, parsed_record: ParsedRecord, enum_id: EnumId, parent_scope_id: ScopeId) throws {
         mut next_constant_value = 0u64
         mut seen_names: {String} = {}
 
@@ -2658,7 +2658,7 @@ struct Typechecker {
             common_fields.push(field.variable_id)
         }
 
-        match record.record_type {
+        match parsed_record.record_type {
             ValueEnum(underlying_type, variants) => {
                 let underlying_type_id = .typecheck_typename(parsed_type: underlying_type, scope_id: parent_scope_id, name: None)
                 mut module = .current_module()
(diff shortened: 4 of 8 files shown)


jntrnr:
I wonder if we should update the wording just a tiny bit to make it easier to visually understand what's happening

`Wrong parameter name in argument label (got '', expected 'good')` is better. What do you think of:

`Missing parameter name in argument label (expected 'good:')`

0GreenClover0:
> I wonder if we should update the wording just a tiny bit to make it easier to visually understand what's happening
> 
> `Wrong parameter name in argument label (got '', expected 'good')` is better. What do you think of:
> 
> `Missing parameter name in argument label (expected 'good:')`

Yeah, I was thinking about that too. I've changed this to `Missing argument label (expected 'good:')`, because there is no label in that case, but we can go with your version if you want.

I've also moved the [samples/functions/argument_labels_bad2](https://github.com/SerenityOS/jakt/pull/1346/files#diff-cc6f610c53f30330f774bfdbaa25771e62b3da6ac90631e684ed6336482d72ca) to the tests/typechecker/no_argument_label_when_required.jakt and adjusted it accordingly, as it is checking for that exact situation.

There is an unfortunate bug (well, it might not be a bug but it's still a little bit weird) when displaying an error that causes us to always point at the first argument that the label was wrong, but this behavior was already present so I won't be touching this in this PR

 ```7 │ fn main() { 
 8 │     foo(true, true) 
   │            ╰─ Missing argument label (expected 'b:')
 9 │ } 
───┴─
Error: Missing argument label (expected 'a:')
───┬─ ./tests/typechecker/no_argument_label_when_required.jakt:8:9
 7 │ fn main() { 
 8 │     foo(true, true) 
   │            ╰─ Missing argument label (expected 'a:')
 9 │ } 
───┴─```

jntrnr:
Looks good

PR #1348, opened 2023-02-06
parser: Don't infinite loop when parsing bogus type longhand

awesomekling:


selfhost/parser.jakt
```diff
@@ -3564,13 +3564,17 @@ struct Parser {
                             if .previous() is ColonColon {
                                 namespaces.push(namespace_name)
                                 .index++
+                            } else {
+                                .error("Expected ‘::’ before", span)
```
alimpfard, line 3568: before...?

awesomekling: Before whatever is highlighted by the span 😅

alimpfard: :thonk:
```
.error("Expected `::' here", span: Span(file_id: span.file_id, start: span.start, end: span.start))
```
?

konradekk: (format/interpolation string could have a clearer intent here… 🤔 yet it is understandable after moment of though—thatʼs least than two moments, so i acceptable for me! 😁)

PR #1349, opened 2023-02-06
interpreter: Fully implement comptime format

alimpfard:
This makes it possible to print out all values in comptime, given any format string.


jntrnr:
You might want to throw a test or two in there, too

alimpfard:
Sure, I'll do that in a bit :^)

PR #1350, opened 2023-02-07
typechecker: Allow subclass to be assigned to type of it's base class

### LINKED ISSUE #1341: typechecker: Allow assignment from derived class to base class
This program currently fails typecheck, but the assignment should be allowed.

```
class Base {}

class Derived: Base {}

fn main() {
    let derived = Derived()
    mut base = Base()
    base = derived
}
```

Furthermore, we should also allow:
```
    let base2: Base = derived
```

lanmonster:
fixes #1341


awesomekling:
Looks like this didn't pass CI testing

PR #1360, opened 2023-02-13
"public" => "pub"

awesomekling:
People keep asking for this, and it seems like a fine idea.


awesomekling:
> Personally not a fan, but can we at least wait until my static reflection PR is done? I _really_ don't feel like rebasing all that on top of this monster patch...

Sure thing :)

konradekk:
i feel non-symmetrised now with `private`! 🤔 `override` is also lengthy and now it optically pops out.

i think i was comfortable with lengthier `public` here (iʼm probably just pretty accustomed to that word not being rusty enough…).

PR #1361, opened 2023-02-13
The Great Yakshave (+static reflection)

alimpfard:
This PR fixes...many things, with the goal of having an initial implementation of static reflection.

new things:
- comptime indexing: `foo.[comptime_expr]` will access the field that `comptime_expr` resolves to (an index, or a string)
- comptime-dependent lambdas: `fn[comptime foo]() { ... }` will capture the value of `foo` as a constant, and depend on it
- static reflection: `reflect T` will resolve to an object tree (described in the prelude for now) describing the type `T`
- specialised generics are now compiled separately, into a separate TU. this actually makes incremental compilation much faster as a million files are no longer recompiled on change (only one is).
- reference-type locals: `let there = &be.references` - you can now stop copying things, plus lifetimes!
- `stores_arguments()` annotation: tell the compiler about the behaviour of external functions
- return types are now mandatory on all functions, except the ones that can be locally (and independently) resolved.

README changes TODO.


jntrnr:
Andreas and I have looked through it. Let's land and improve as we go.

PR #1364, opened 2023-02-15
selfhost: Add the `sizeof` operator

awesomekling:
This first version only works on types, i.e `sizeof T`. It codegens to a simple C++ `sizeof(T)` expression.


konradekk:
i feel that `sizeof` is not really explicit—what would you tell about `bitsize` for the name instead? 🤔
(or something equally descriptive, e.g. `bitsin`?!)

awesomekling:
It's called `sizeof` in multiple major languages. It's an established term of art. 

`bitsize` and `bitsin` are not even accurate, as we're yielding the size in bytes, not in bits.

konradekk:
> `bitsize` and `bitsin` are not even accurate, as we're yielding the size in bytes, not in bits.

oh, oh! i was reading the other commit where bits were relevant—thanks for clarifying that!

PR #1366, opened 2023-02-15
Bring back --emit-cpp-source-only and --prettify-cpp-source compiler cli flags.

fahrradflucht:
While still advertised, in the help, some compiler flags seam to have been lost. Bring back `--emit-cpp-source-only`, which still had its short-form `-S` working, and re-implement `--prettify-cpp-source`.

Sidenote: While cleaning up the help output format I tried to understand the difference between flags that are classified as "flags" and those that are classified as "options" - I couldn't figure it out. Is there any logic behind this grouping? 🤔


robryanx:
I would expect the flags to have no argument and the options to have an argument. But yeah, there are some under flags with arguments currently, could just move them.

fahrradflucht:
> I would expect the flags to have no argument and the options to have an argument. But yeah, there are some under flags with arguments currently, could just move them.

Why is there no 💡 reaction in Github? That makes a lot of sense, thanks.

PR #1367, opened 2023-02-16
types: Make try-catch use the same control flow resolution as if-else

### LINKED ISSUE #1365: Incorrect "reaches end of non-void function" error when `catch` clause calls a never-returning function
This currently fails to compile, despite being valid. `abort()` has return type `never` in the prelude, so the compiler should see that all branches return here.

```
fn foo() -> [i64] {
    try {
        return []
    } catch {
        abort()
    }
} 
```

0GreenClover0:
In theory fixes #1365 (and probably many more situations), unfortunately we can't currently `return` from `try` blocks.

### DIFF
--- a/selfhost/types.jakt
+++ b/selfhost/types.jakt
@@ -1579,6 +1579,13 @@ boxed enum CheckedExpression {
             true => BlockControlFlow::NeverReturns
             else => BlockControlFlow::MayReturn
         }
+        TryBlock(stmt, catch_block) => {
+            guard stmt is Block(block) else {
+                panic("Try block doesn't have a block")
+            }
+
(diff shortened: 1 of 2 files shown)


selfhost/types.jakt
```diff
@@ -1579,6 +1579,13 @@ boxed enum CheckedExpression {
             true => BlockControlFlow::NeverReturns
             else => BlockControlFlow::MayReturn
         }
+        TryBlock(stmt, catch_block) => {
+            guard stmt is Block(block) else {
+                panic("Try block doesn't have a block")
```
konradekk, line 1584: suggestion: “Try clause is not a block”?

konradekk (changes requested):

PR #1374, opened 2023-02-17
parser: Don't incorrectly error when value enum has eof after ending `}`

0GreenClover0:
By applying the same logic as in the normal enums.

Spotted in https://youtu.be/IRT5wGTH4iY?t=282


awesomekling:
Hah, thank you for tracking that down! :^)

PR #1383, opened 2023-02-20
typechecker: Typecheck function overrides

### LINKED ISSUE #1335: typechecker: Override functions must have same `throws` attribute as the virtual they are overriding
The following program currently passes typecheck but fails in the C++ compiler.
We should reject the override function because it doesn't have the same throwsiness as the base class function it's overriding.
```
class Base {
    public virtual function foo(this) throws { }
}

class Derived: Base {
    public override function foo(this) { }
}
```

### LINKED ISSUE #1336: typechecker: Override functions need to be matched based on parameters, not just name
The following currently passes typecheck, but chokes in the C++ compiler.
We should reject the override function because it doesn't have the same signature as the one in the base class.
```
class Base {
    public virtual function foo(this, x: i64) { }
}

class Derived: Base {
    public override function foo(this) { }
}
```

fahrradflucht:
Actually typecheck that the relevant parts of an override function match the virtual they target.

Closes #1335
Closes #1336

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -2878,16 +2878,13 @@ struct Typechecker {
         }
 
         for method in record.methods {
-            if method.is_override {
-                if not all_virtuals.contains(method.parsed_function.name) {
-                    .error("Missing virtual for override", method.parsed_function.name_span)
-                }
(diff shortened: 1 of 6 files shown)


awesomekling:
Awesome! Thank you so much for fixing these issues :)

Issue #1385, opened 2023-02-21
Build fails after AK const-laundering changes


donelli:
I'm trying to build stage0 on macos, but i'm getting an error.

Here are the steps that I have made:

- Cloned the repo
- Runned: `cmake -B build -GNinja -DCMAKE_CXX_COMPILER=clang++`
- Runned: `ninja -C build`

Then this error happens:

```sh
➜  jakt git:(main) ✗ ninja -C build                                     
ninja: Entering directory `build'
[0/2] Re-checking globbed directories...
[1/71] Building CXX object runtime/CMakeFiles/jakt_main.dir/Main.cpp.o
FAILED: runtime/CMakeFiles/jakt_main.dir/Main.cpp.o 
/usr/bin/clang++  -I/Users/user/Desktop/jakt/runtime -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.0.sdk -fPIC -Wno-unused-local-typedefs -Wno-unused-function -Wno-unused-variable -Wno-unused-parameter -Wno-unused-but-set-variable -Wno-unused-result -Wno-implicit-fallthrough -Wno-trigraphs -Wno-parentheses-equality -Wno-unqualified-std-cast-call -Wno-user-defined-literals -Wno-literal-suffix -Wno-return-type -Wno-deprecated-declarations -Wno-unknown-warning-option -Wno-unused-command-line-argument -fdiagnostics-color=always -Wno-unknown-attributes -fno-exceptions -std=gnu++20 -MD -MT runtime/CMakeFiles/jakt_main.dir/Main.cpp.o -MF runtime/CMakeFiles/jakt_main.dir/Main.cpp.o.d -o runtime/CMakeFiles/jakt_main.dir/Main.cpp.o -c /Users/user/Desktop/jakt/runtime/Main.cpp
In file included from /Users/user/Desktop/jakt/runtime/Main.cpp:1:
In file included from /Users/user/Desktop/jakt/runtime/lib.h:65:
/Users/user/Desktop/jakt/runtime/Jakt/DeprecatedString.h:22:40: error: no viable overloaded '='
    *bit_cast<RefPtr<StringImpl>*>(&a) = result.impl(); // I'm sorry.
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~
[...]

PR #1386, opened 2023-02-21
runtime: Pin version of serenity before RefPtr const-correctness changes

ADKaster:
https://github.com/SerenityOS/serenity/pull/17557 has conflicts with how we do codegen of class types. For one, const-ness is not a property of the type, but the binding. We need to figure out a way to pass this information to codegen_struct_type so that it can generate const-correct code.

Ref #1385. The true fix will need some elbow grease.


awesomekling:
Let's revert the RP/NNRP change in serenity instead, otherwise we risk accumulating more Jakt-incompatible changes while swimming in the dark like this.

PR #1395, opened 2023-02-24
typechecker: Teach match about true & false being exhaustive

0GreenClover0:
Also:
- change, well, most but probably not all occurrences of the `true => else =>` pattern (or much less frequently -
`false => else =>`) to the `true => false =>` pattern,
- fix one test that was affected by this (which had a `true => false => else =>` pattern, `else` obviously not reachable),
- fix one error not starting with an uppercase.

Funny business like this is still possible though:

```rust
fn main() {
    let x = false
    let y = false
    match x {
        true => { }
        false => { }
        (y == y) => { }
    }
}
```

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -916,7 +916,7 @@ struct Typechecker {
                 let maybe_file_name = .compiler.search_for_path(name_and_span.0)
                 let file_name = match maybe_file_name.has_value() {
                     true => maybe_file_name!
-                    else => .get_root_path().parent().join(name_and_span.0).replace_extension("jakt")
+                    false => .get_root_path().parent().join(name_and_span.0).replace_extension("jakt")
                 }
                 if File::exists(file_name.to_string()) {
                     module_name_and_span = name_and_span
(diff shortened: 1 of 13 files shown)


selfhost/typechecker.jakt
```diff
@@ -7751,6 +7751,11 @@ struct Typechecker {
                 .error("Can't match on 'void' type", checked_expr.span())
             }
             else => {
+                let is_boolean_match = type_to_match_on is Bool
+                mut seen_true = false
+                mut seen_false = false
+                mut catch_all_span: Span? = None
```
alimpfard, line 7757: This is perfectly fine for now, but going further...we should probably move the value exhaustiveness check somewhere else to open space for other types.

selfhost/typechecker.jakt
```diff
@@ -7883,6 +7890,14 @@ struct Typechecker {
                                 )
                                 let checked_expression = .typecheck_expression_and_dereference_if_needed(new_condition, scope_id, safety_mode, type_hint: Some(subject_type_id), span)
 
+                                if is_boolean_match and checked_expression is Boolean(val) {
+                                    if val {
+                                        seen_true = true
+                                    } else {
+                                        seen_false = true
+                                    }
```
alimpfard, line 7898:
```suggestion
                                    seen_true = val
```

alimpfard (commented):
Big fan of this change, <3

selfhost/typechecker.jakt
alimpfard: oh nevermind, I'm silly :P

0GreenClover0: when I was first writing this I thought about this too! haha

PR #1396, opened 2023-02-24
typechecker: Auto-dereference range expression in the for-loop

### LINKED ISSUE #1391: typechecker: `for` should auto-dereference the iterable expression
This currently does not compile:
```
fn whee(values: &[i64]) {
    for v in values {
    }
}

fn main() {
    let values = [1, 2, 3]
    whee(&values)
}
```
I would expect the compiler to auto-dereference `values` in the `for` loop.

Basically, `whee()` should behave as if you had typed this:
```
fn whee(values: &[i64]) {
    for v in *values {
    }
}
```

0GreenClover0:
Resolves #1391


awesomekling:
Oh wow, that was easier than I expected! Thank you so much for fixing it :^)

PR #1401, opened 2023-02-26
typechecker: Allow subclass to be assigned to type of it's base class (v2)

### LINKED ISSUE #1341: typechecker: Allow assignment from derived class to base class

0GreenClover0:
Partially based on #1350. Resolves #1341 + assigning Derived to Optional\<Base\> and assigning Derived to Base weak ptr.

### DIFF
--- a/samples/classes/assign_derived_to_base.jakt
+++ b/samples/classes/assign_derived_to_base.jakt
@@ -1,0 +1,17 @@
+/// Expect:
+/// - output: "PASS\n"
+
+class Base {
+    public fn test(this) {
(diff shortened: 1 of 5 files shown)


lanmonster:
Thank you! I ended up not getting as much time to work on this as I hoped.

PR #1413, opened 2023-03-13
repl: Avoid showing an error when entering an empty line

aloussase:
Currently the repl shows a parser error "Unsupported expression" when the user enters an empty line. This commit adds a check for the empty line and skips the loop to avoid the error.


AtkinsSJ (approved):
Welcome to the project! Looks good to me. :^)

PR #1414, opened 2023-03-14
repl: Add multiline input support

aloussase:
Now multiple lines can be entered at the repl using a pair of delimitters. For now they are "{-" to start and "-}" to end but they can be changed in the future.

Example:

```
> {-
.. struct Person {
..     name: String
.. }
.. -}
> let p = Person(name: "John")
```

### DIFF
--- a/selfhost/repl.jakt
+++ b/selfhost/repl.jakt
@@ -31,12 +31,54 @@ struct Editor {
             throw Error::from_errno(42)
         }
 
-        let editor = Editor(
+        return Editor(
             standard_input_file: std_in
             line_pointer: allocate<c_char>(count: 4096)
             prompt
         )
-        return editor
+    }
+
+    fn check_parens(anon input: String) -> bool {
+        mut unmatched_parens   = 0
+        mut unmatched_brackets = 0
+        mut unmatched_curlies  = 0
+
+        mut pos: usize = 0
+
+        while pos < input.length() {
+            match input.byte_at(pos) {
+                b'(' => { unmatched_parens += 1   }
+                b')' => { unmatched_parens -= 1   }
+                b'[' => { unmatched_brackets += 1 }
+                b']' => { unmatched_brackets -= 1 }
+                b'{' => { unmatched_curlies += 1  }
+                b'}' => { unmatched_curlies -= 1  }
+                else => {}
+            }
+
+            pos += 1
+        }
+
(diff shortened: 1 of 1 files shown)


alimpfard (commented):
:thinking: This seems alright as a first cut, but alternative options:
- Run the lexer on the input and keep track of open enclosures (`[{(` sounds like the most sane ones to track), accept the input only if open==0
- Have the parser tell you whether a parse error could be resolved if there was more input (e.g. <https://github.com/SerenityOS/serenity/blob/master/Userland/Shell/AST.h#L1453>); though this option doesn't make the result intuitive since syntax errors might come from nested syntactic structures that we mistakenly parsed - jakt is _somewhat_ context-sensitive.

aloussase:
Hi! Thanks for the feedback, but I don't understand why would I want to run the lexer on the input just to allow the user to enter multiple lines. You mean so that we don't need the delimiters?
My proposed solution is inspired in what the GHCi repl does (GHCi uses `:{` and `:}` as delimiters).

alimpfard:
Yes, my suggestions are indeed trying to get rid of the delimiters.
I'm aware of ghci's `:{/}` repl...commands, but I'm not a fan of that style in REPLs.

We could still keep this style if it's popular enough (we're gonna need some help string to actually tell the user about it too), but I personally would've used `^v enter` if I wanted to be explicit.

aloussase:
Yes I can see what you are saying. Maybe I can add some code to check for unbalanced parens as you said `[({`, but I think ultimately we would need something else to make sure it works 100% of times. It would be easier if Jakt had statement terminators haha.

What do you say if I add code for checking for unbalanced parens and keep the delimiters for now, just for cases where checking for parens does not work?

alimpfard:
I don't see how counting enclosures wouldn't work, but sure, I don't mind keeping the delimiters

selfhost/repl.jakt
```diff
@@ -31,12 +31,54 @@ struct Editor {
...
+
+    fn check_parens(anon input: String) -> bool {
+        mut unmatched_parens   = 0
+        mut unmatched_brackets = 0
+        mut unmatched_curlies  = 0
+
+        mut pos: usize = 0
+
+        while pos < input.length() {
+            match input.byte_at(pos) {
```
alimpfard, line 49: This will break with parens/brackets/curlies inside strings, that's why I suggested using the lexer

PR #1424, opened 2023-04-12
readme: fix a typo (pointee -> pointer)

Wertzui123:


AtkinsSJ:
Hi there!

Thanks, but this isn't actually a typo. It's saying that when the pointee (the thing pointed to) is destructed, the WeakPtr becomes empty automatically. To be fair, my spell checker is telling me that "pointee" is not a real word, so I can understand thinking it's a typo.

Wertzui123:
Ah ok, thanks for the clarification

PR #1426, opened 2023-04-25
`unsafe` expressions

### LINKED ISSUE #1407: `unsafe` expressions
We are currently limited by `unsafe {}` only working in a statement context.

This makes it awkward to initialize new variables with the result of an unsafe function call, i.e:

```
mut c = 0
unsafe {
	c = some_unsafe_function(a, b)
}
```

It would be great if we could also do things like:

```
let c = unsafe {
	yield some_unsafe_function(a, b)
}
```

Perhaps this could also work, although I haven't thought very hard about the implications for ambiguity etc:

```
let c = unsafe some_unsafe_function(a, b)
```

ArthurPV:
Resolves #1407.

### DIFF
--- a/bootstrap/stage0/__unified_forward.h
+++ b/bootstrap/stage0/__unified_forward.h
@@ -418,6 +418,7 @@ struct Try;
 struct TryBlock;
 struct Reflect;
 struct Garbage;
+struct Unsafe;
 }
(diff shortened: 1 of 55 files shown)


awesomekling:
Oh awesome, thank you so much for implementing this :)

PR #1427, opened 2023-04-26
Allow default arguments for sum enum variant constructors

### LINKED ISSUE #1399: Allow default arguments for sum enum variant constructors
It would be handy to allow something like this:
```
enum Animal {
    Dog(name: String, cute: bool = true)
}
```

ArthurPV:
Resolves #1399.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -2756,31 +2756,45 @@ struct Typechecker {
                         for field in common_fields {
                             fields.push(field)
                         }
+                        
+                        if (variant.params is Some(variant_params) and variant.default_values is Some(variant_default_values)) {
+                            for i in 0..variant_params.size() {
+                                let param = variant_params[i]
+                                let default_value = variant_default_values[i]
 
-                        for param in variant.params! {
-                            if seen_fields.contains(param.name) {
-                                .error(format("Enum variant '{}' has a member named '{}' more than once", variant.name, param.name), param.span)
-                                continue
+                                if seen_fields.contains(param.name) {
+                                    .error(format("Enum variant '{}' has a member named '{}' more than once", variant.name, param.name), param.span)
+                                    continue
+                                }
+                                seen_fields.add(param.name)
+                                let type_id = .typecheck_typename(parsed_type: param.parsed_type, scope_id: enum_.scope_id, name: param.name)
+                                let checked_var = CheckedVariable(
+                                    name: param.name
+                                    type_id
+                                    is_mutable: param.is_mutable
+                                    definition_span: param.span
+                                    type_span: None
+                                    visibility: CheckedVisibility::Public
+                                )
+
+                                mut checked_default_value: CheckedExpression? = None
+
+                                if default_value is Some(expression) {
+                                    // FIXME: The typechecker can't find expressions that are outside the scope of the enum.
+                                    // FIXME: Get the real safety mode from somewhere, right now it assumes it is safe.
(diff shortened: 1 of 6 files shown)


ArthurPV:
I don't know the language very well, but I was asking if it was normal that these two codes don't compile:

```rust
fn h() -> i64 => 42

enum S {
    X
    fn my_func(this, anon def: i64 = h()) {
        println("{}", def)
    }
}

fn main() {
    S::X.my_func()
}
```

```rust
fn h() -> bool => true

enum Animal {
    Dog(name: String, cute: bool = h())
}

fn main() {
    let animal = Animal::Dog(name: "John")

    match animal {
        Dog(cute) => {
            if cute {
                println("I am cute")
            }
        }
    }
}
```

Here are the compiler outputs I get for these two codes:

```
Error: Call to unknown function: ‘h’
───┬─ ./test.jakt:5:38
 4 │     X 
 5 │     fn my_func(this, anon def: i64 = h()) { 
   │                                      ┬     
   │                                      ╰─ Call to unknown function: ‘h’
 6 │         println("{}", def) 
───┴─
```

```
Error: Call to unknown function: ‘h’
───┬─ ./test.jakt:4:36
 3 │ enum Animal { 
 4 │     Dog(name: String, cute: bool = h()) 
   │                                    ┬   
   │                                    ╰─ Call to unknown function: ‘h’
 5 │ } 
───┴─
```

alimpfard:
There are definitely a lot of rough edges around default arguments (this, also whether they can throw, where they're actually supposed to be evaluated, etc), so I'd say what you're seeing is a bug with either:
- order of function declarations
- where the default argument is evaluated
- where the default argument is typechecked
- ?

selfhost/typechecker.jakt
```diff
@@ -2780,7 +2780,8 @@ struct Typechecker {
                                 mut checked_default_value: CheckedExpression? = None
 
                                 if default_value is Some(expression) {
-                                    // FIXME: The typechecker can't find expressions that are outside the scope of the enum.
+                                    //FIXME: The typechecker can't find expressions that are outside the scope of the enum.
+                                    //FIXME: Get the real safety mode from somewhere, right now it assumes it is safe.
```
alimpfard, line 2784: Please squash this commit into the second commit, and please format the comment correctly (`// FIXME`, not `//FIXME`)

tests/typechecker/default_arguments_sum_enum.jakt
```diff
@@ -0,0 +1,17 @@
+/// Expect:
+/// - output: "I am cute\n"
+enum Animal {
```
alimpfard, line 3: Please leave one or two empty lines here for a nicer viewing experience :)

alimpfard:
(Build failure is expected, caused by AK changes)

ArthurPV:
Yes, I know the problem. It's the `runtime/AK/Format.h` file at line 718 that makes the compilation fail, just comment it out: 

```c++
template<typename T>
struct Formatter<Optional<T>> : Formatter<FormatString> {
    static constexpr bool is_debug_only() { return true; }

    ErrorOr<void> format(FormatBuilder& builder, Optional<T> const& optional)
    {
        if (optional.has_value())
            return Formatter<FormatString>::format(builder, "Optional({})"sv, *optional);
        return builder.put_literal("OptionalNone"sv);
    }
};
```

alimpfard:
Could you rebase on top of main? somehow GH seems okay with merging this as-is, but the last commit has a different hash and is a duplicate, so I'm not convinced GH knows what it's doing :sweat_smile:

ArthurPV:
Ok it's done.

PR #1443, opened 2023-05-11
typechecker: Infer correct type for negative ints

### LINKED ISSUE #1437: No number type inference for negative numbers
```
fn number_inference() -> i32 {
    return -1
}
```
`Type mismatch: expected ‘i32’, but got ‘i64’`

Would expect this to work as it does for zero or positive numbers and infer i32 type based on the return type.

abooishaaq:
Fixes #1437.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -7643,6 +7643,7 @@ struct Typechecker {
         UnaryOp(expr, op, span) => {
             let checked_expr = match op {
                 Dereference => .typecheck_expression(expr, scope_id, safety_mode, type_hint: None)
+                Negate => .typecheck_expression_and_dereference_if_needed(expr, scope_id, safety_mode, type_hint, span)
                 else => .typecheck_expression_and_dereference_if_needed(expr, scope_id, safety_mode, type_hint: None, span)
             }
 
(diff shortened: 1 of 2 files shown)


cybergsus:
Looks great! If you could add a regression test to the test suite, that'd be great as well. Something in the lines of:
```
/// Expected:
/// - output: ""

// brought up by #1437 

fn number_inference() -> i32 {
    return -1
}

fn main() {}
```

0GreenClover0:
The test fails because it's actually `/// Expect:` instead of `/// Expected:`

selfhost/typechecker.jakt
```diff
@@ -7643,6 +7643,7 @@ struct Typechecker {
         UnaryOp(expr, op, span) => {
             let checked_expr = match op {
                 Dereference => .typecheck_expression(expr, scope_id, safety_mode, type_hint: None)
+                Negate => .typecheck_expression_and_dereference_if_needed(expr, scope_id, safety_mode, type_hint, span)
```
alimpfard, line 7646: Please make the commit message explain what's being done, not its result ("typechecker: Pass the correct type hint for unary negate" or something like that)

alimpfard (commented):
changes LGTM, just one nitpick:

robryanx:
Looks good! Fixes the issue.

PR #1447, opened 2023-05-27
lexer: use a set for cpp keyword check

robryanx:
On an optimised build:
Using match
```
rob@rob-linux:~/jakt$ time jakt -e 2000 selfhost/main.jakt
{"hover": "mut output: struct String"}

real	0m9.275s
user	0m9.182s
sys	0m0.080s
```
Using a set
```
rob@rob-linux:~/jakt$ time jakt -e 2000 selfhost/main.jakt
{"hover": "mut output: struct String"}

real	0m4.774s
user	0m4.681s
sys	0m0.093s
```

Stylistically the match is nicer, maybe it could be made just as fast. Match seems (at least to me, maybe someone else can confirm those numbers) to be costing a lot of performance on a hot code patch like this, for something similar like this that can map to an exists lookup.

### DIFF
--- a/selfhost/lexer.jakt
+++ b/selfhost/lexer.jakt
@@ -187,72 +187,6 @@ enum Token {
     }
 }
 
-fn is_illegal_cpp_keyword(string: String) throws -> bool => match string {
-    "alignas" => true
-    "alignof" => true
-    "and_eq" => true
-    "asm" => true
-    "auto" => true
-    "bitand" => true
-    "bitor" => true
-    "case" => true
-    "char" => true
-    "char8_t" => true
-    "char16_t" => true
-    "char32_t" => true
-    "compl" => true
-    "concept" => true
-    "const" => true
-    "consteval" => true
-    "constexpr" => true
-    "constinit" => true
-    "const_cast" => true
-    "co_await" => true
-    "co_return" => true
-    "co_yield" => true
-    "decltype" => true
-    "delete" => true
-    "do" => true
-    "double" => true
-    "dynamic_cast" => true
-    "explicit" => true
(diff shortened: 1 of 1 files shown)


selfhost/lexer.jakt
```diff
@@ -302,9 +236,75 @@ struct Lexer implements(ThrowingIterable<Token>) {
     input: [u8]
     compiler: Compiler
     comment_contents: [u8]?
+    illegal_cpp_keywords: [String:bool]
 
     fn lex(compiler: Compiler) throws -> [Token] {
-        mut lexer = Lexer(index: 0, input: compiler.current_file_contents, compiler, comment_contents: None)
+        let illegal_cpp_keywords: [String:bool] = [
```
alimpfard, line 242: How about a set instead?

selfhost/lexer.jakt
```diff
@@ -436,7 +436,7 @@ struct Lexer implements(ThrowingIterable<Token>) {
                 .error("reserved identifier name", span)
             }
 
-            if is_illegal_cpp_keyword(string) {
+            if .illegal_cpp_keywords.get(string).has_value() {
```
alimpfard, line 439: ...which would turn this into
```suggestion
            if .illegal_cpp_keywords.contains(string) {
```

selfhost/lexer.jakt
robryanx: Sure, performance seems around the same for both, updated it.

alimpfard: Yeah, didn't expect a performance difference, just less noise and a more proper data structure.

PR #1448, opened 2023-06-03
meta: Syntax highlight jakt files viewed in github

IrishBruse:
I added syntax highlighting to jakt files viewed in github.
Should make looking through PRs and example code for new comers a bit nicer

If you look at a jakt file in my fork you will see that it has colors rust has the most similar syntax so i went with that language.

https://github.com/IrishBruse/jakt/blob/main/samples/apps/cat.jakt

image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 * eof=lf
 
+*.jakt linguist-language=rust


alimpfard:
I'd rather we get our own linguist-compatible syntax definitions and teach GH about Jakt instead of using Rust's syntax highlighting as even a temporary stopgap.

robryanx:
Yeah I was looking into this a while ago, the requirements for inclusion into linguist are a bit fuzzy with the changes to github search. If we PRed to linguist they might accept.

PR #1449, opened 2023-06-16
Automatically detect the imported extern declarations 

alimpfard:
This uses either Clang or LibCpp to parse the imported C/C++ file and
produce bindings on the fly.
Currently supported decl kinds:
- namespaces (LibCpp and Clang)
- non-generic functions (LibCpp and Clang)
- non-generic structs (LibCpp and Clang)
- enums (Clang)
- generic structs (Clang)

## Build

CMake will handle it, but pass `--config jakt.cpp_import.processor=clang` to build with clang manually.

## Samples

SDL: 
image

AK:
image

PR #1451, opened 2023-06-22
CMakeLists: Fix Windows symlink issues

Optimoos:
There are permission issues on Windows when creating symlinks and hard links. I've tested all permutations, and the only consistent method was to use junctions. CMake has the flags for creating junctions in its core, but they are not exposed anywhere. I will try to submit a patch there to get that corrected, but for now this change should allow CMake to function reliably on Windows.


Optimoos:
Do I need to do anything about the CI failures? These look erroneous, I'm not sure how the change I made could be causing the issue, especially given there should be zero change to the Ubuntu build.

alimpfard:
Nope, those should be fixed by https://github.com/SerenityOS/jakt/pull/1449/commits/29074e82edc9801afc1ef5c9e0ebde03ad04ffe9
I'll split that commit up and push it tomorrow.

alimpfard:
(Fix pushed as 13e73b5daed777695cc03eb7eeca3a68dd6c3b2f)

CMakeLists.txt
```diff
@@ -185,11 +185,23 @@ endif()
 if (CMAKE_CONFIGURATION_TYPES)
   foreach (build_type IN LISTS CMAKE_CONFIGURATION_TYPES)
     execute_process(COMMAND "${CMAKE_COMMAND}" -E make_directory "${CMAKE_CURRENT_BINARY_DIR}/${build_type}/include")
-    execute_process(COMMAND "${CMAKE_COMMAND}" -E ${LINK_COMMAND} "${CMAKE_CURRENT_SOURCE_DIR}/runtime" "${CMAKE_CURRENT_BINARY_DIR}/${build_type}/include/runtime")
+    if (CMAKE_HOST_WIN32)
+      file(TO_NATIVE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/runtime" _srcDir)
+      file(TO_NATIVE_PATH "${CMAKE_CURRENT_BINARY_DIR}/${build_type}/include/runtime" _dstDir)
+      execute_process(COMMAND cmd.exe /c mklink /J "${_dstDir}" "${_srcDir}")
+    else()
+      execute_process(COMMAND "${CMAKE_COMMAND}" -E ${LINK_COMMAND} "${CMAKE_CURRENT_SOURCE_DIR}/runtime" "${CMAKE_CURRENT_BINARY_DIR}/${build_type}/include/runtime")
```
ADKaster, line 193: If we're going to use a junction instead of a hard/softlink, then we should get rid of all the logic above that sets LINK_COMMAND, including the cache variable.

CMakeLists.txt
```diff
@@ -185,11 +185,23 @@ endif()
...
+    endif()
   endforeach()
 else()
   execute_process(COMMAND "${CMAKE_COMMAND}" -E make_directory "${CMAKE_CURRENT_BINARY_DIR}/include")
-  execute_process(COMMAND "${CMAKE_COMMAND}" -E ${LINK_COMMAND}  "${CMAKE_CURRENT_SOURCE_DIR}/runtime" "${CMAKE_CURRENT_BINARY_DIR}/include/runtime")
+  if (CMAKE_HOST_WIN32)
+    file(TO_NATIVE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/runtime" _srcDir)
+    file(TO_NATIVE_PATH "${CMAKE_CURRENT_BINARY_DIR}/include/runtime" _dstDir)
+    execute_process(COMMAND cmd.exe /c mklink /J "${_dstDir}" "${_srcDir}")
```
ADKaster, line 201: What about the cmake/install-rules.cmake parts that use the link command?

Do those work on your machine? 

```
cmake -B build -T ClangCL -DCMAKE_INSTALL_PREFIX=jakt-install
cmake --build build
cmake --install build
```

Or do we need to do the same check there?

ADKaster (changes requested):

CMakeLists.txt
Optimoos: `LINK_COMMAND` is still effective on *nix if someone wanted a hard link over symlink there.

CMakeLists.txt
Optimoos: That part doesn't complain for some reason, I don't really understand why it would be any different.

ADKaster (approved):
This is fine, but let's wait for CxByte's cross-compiling patch to land first.

ADKaster (approved):
This is fine, but let's wait for CxByte's cross-compiling patch to land first.

PR #1452, opened 2023-06-22
CMake: More Windows build fixes

Optimoos:
There was some inconsistent behaviour when building on Windows depending on the environment and options provided. These changes aim to capture those and provide direction.

If a generator isn't specified with -GNinja, cmake may try to use Visual Studio which won't work. The CMake documentation states that projects shouldn't set CMAKE_GENERATOR, so an error is thrown. SEND_ERROR doesn't halt the build, so it may make sense to change this to FATAL_ERROR.

MSVC only gets set by clang-cl, not clang++. The MSVC test blocks have been expanded to better catch clang++.

Using -Wall with clang-cl enables a bunch of C++ version errors that aren't suppressed by /std:c++20. Using -W[123] with clang-cl specifically solves this. See
https://github.com/llvm/llvm-project/issues/42913

The bootstrap documentation has had a few lines added to assist with understanding these quirks and getting things going on Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,7 @@ target_include_directories(jakt_stage0
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/runtime>
 )
 apply_output_rules(jakt_stage0)
-if (MSVC)
+if (MSVC OR (DEFINED MSVC_VERSION AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
     target_link_options(jakt_stage0 PRIVATE LINKER:/STACK:0x800000)
 endif()
 
@@ -155,7 +155,7 @@ target_include_directories(jakt_stage1
 )
 add_executable(Jakt::jakt_stage1 ALIAS jakt_stage1)
 apply_output_rules(jakt_stage1)
-if (MSVC)
+if (MSVC OR (DEFINED MSVC_VERSION AND CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "GNU"))
     target_link_options(jakt_stage1 PRIVATE LINKER:/STACK:0x800000)
 endif()
 
@@ -175,7 +175,7 @@ if (FINAL_STAGE GREATER_EQUAL 2)
   )
   add_executable(Jakt::jakt_stage2 ALIAS jakt_stage2)
   apply_output_rules(jakt_stage2)
-  if (MSVC)
+  if (MSVC OR (DEFINED MSVC_VERSION AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
     target_link_options(jakt_stage2 PRIVATE LINKER:/STACK:0x800000)
   endif()
 endif()
--- a/cmake/jakt-executable.cmake
+++ b/cmake/jakt-executable.cmake
@@ -9,7 +9,6 @@ STRING(TOLOWER ${JAKT_TARGET_IN} JAKT_TARGET)
 
 function(add_jakt_compiler_flags target)
   target_compile_options("${target}" PRIVATE
-    -Wall
     -Wextra
     -Werror
     -Wno-unused-local-typedefs
@@ -36,9 +35,9 @@ function(add_jakt_compiler_flags target)
   )
   if (MSVC)
     # For clang-cl, which shows up to CMake as MSVC and accepts both kinds of arguments
-    target_compile_options("${target}" PRIVATE /permissive- /utf-8 /EHsc-)
+    target_compile_options("${target}" PRIVATE /permissive- /utf-8 /EHsc- -W4)
   else()
-    target_compile_options("${target}" PRIVATE -fno-exceptions -fdiagnostics-color=always)
+    target_compile_options("${target}" PRIVATE -fno-exceptions -fdiagnostics-color=always -Wall)
   endif()
   if (CYGWIN OR MSYS)
     target_compile_options("${target}" PRIVATE -Wa,-mbig-obj)
--- a/cmake/jakt-executable.cmake
+++ b/cmake/jakt-executable.cmake
@@ -35,7 +35,7 @@ function(add_jakt_compiler_flags target)
   )
   if (MSVC)
     # For clang-cl, which shows up to CMake as MSVC and accepts both kinds of arguments
-    target_compile_options("${target}" PRIVATE /permissive- /utf-8 /EHsc- -W4)
+    target_compile_options("${target}" PRIVATE /permissive- /utf-8 /EHsc- -W[123])
   else()
     target_compile_options("${target}" PRIVATE -fno-exceptions -fdiagnostics-color=always -Wall)
   endif()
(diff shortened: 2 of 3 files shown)


ADKaster:
Last I tried on windows it worked fine to use clang.exe from a Visual Studio Developer Powershell/Command Prompt, and worked to use MsBuild. What Windows and Visual Studio version did you use?

cmake/jakt-executable.cmake
```diff
@@ -8,8 +8,12 @@ set(JAKT_TARGET_IN "${CMAKE_SYSTEM_PROCESSOR}-unknown-${CMAKE_SYSTEM_NAME}-unkno
 STRING(TOLOWER ${JAKT_TARGET_IN} JAKT_TARGET)
 
 function(add_jakt_compiler_flags target)
+  if (MSVC)
+    target_compile_options("${target}" PRIVATE -W[123])
```
ADKaster, line 12: this should just be /W4.

Extra space below in the -Wall branch.

instead of doing this, can you add /W4 and -Wall to the branches on lines 43 and 45 below with the other MSVC check?

CMakeLists.txt
```diff
@@ -22,6 +22,10 @@ if (CMAKE_HOST_WIN32)
   set(NO_SYMLINKS_DEFAULT ON)
 endif()
 
+if (CMAKE_GENERATOR MATCHES "Visual Studio")
+  message(SEND_ERROR "Visual Studio generators are not supported. Please use Ninja (-GNinja) instead.")
```
ADKaster, line 26: This is not correct. you can use Visual Studio generators with clang-cl like so:

```
cmake -B build -T ClangCL
```

CMakeLists.txt
```diff
@@ -155,7 +159,7 @@ target_include_directories(jakt_stage1
 )
 add_executable(Jakt::jakt_stage1 ALIAS jakt_stage1)
 apply_output_rules(jakt_stage1)
-if (MSVC)
+if (MSVC OR (DEFINED MSVC_VERSION AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
```
ADKaster, line 162: you're looking for `CMAKE_CXX_COMPILER_FRONTEND_VARIANT`

https://gitlab.kitware.com/cmake/cmake/-/issues/22666

https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_FRONTEND_VARIANT.html

documentation/cmake-bootstrap.md
```diff
@@ -45,7 +45,17 @@ for validation, set the CMake cache variable `FINAL_STAGE` to `2`.
 
 Jakt is known to compile with clang >=13 on Linux, macOS and Windows. g++ also works, provided the version is >=10.2.
 
-MSVC is not supported, however clang-cl.exe and clang.exe do work and clang-cl is used in CI.
+MSVC is not supported, however clang-cl.exe and clang.exe do work and clang-cl is used in CI. You must specify `-GNinja`, as CMake may otherwise default to MSVC generators which will fail. Additionally, clang-cl seems dependent on environment variables to find MSVC, so ensure you're running it from the Developer Command Prompt. Clang++ doesn't seem to have the same requirement.
```
ADKaster, line 48: Developer command prompt or developer powershell work equally well. I have only tried with VS 2022 Community Preview, I'm not sure what version of VS 2022 is the newest that will work.,  (on 17.7.0 preview rn).

ADKaster (changes requested):

CMakeLists.txt
ADKaster: That problem is fixed by my open PR to serenity in Assertions.h

Optimoos:
Latest push should be closer, let me know if there's anything else.

cmake/jakt-executable.cmake
```diff
@@ -35,7 +35,7 @@ function(add_jakt_compiler_flags target)
   )
   if (MSVC)
     # For clang-cl, which shows up to CMake as MSVC and accepts both kinds of arguments
-    target_compile_options("${target}" PRIVATE /permissive- /utf-8 /EHsc- -W4)
+    target_compile_options("${target}" PRIVATE /permissive- /utf-8 /EHsc- -W[123])
```
ADKaster, line 38: If this truly is the only way, then it should be fixup squashed back into the first commit.

However, when I tested it, using `/W4` rather than `-W4` compiled just fine.

CMakeLists.txt
```diff
@@ -155,7 +155,7 @@ target_include_directories(jakt_stage1
 )
 add_executable(Jakt::jakt_stage1 ALIAS jakt_stage1)
 apply_output_rules(jakt_stage1)
-if (MSVC)
+if (MSVC OR (DEFINED MSVC_VERSION AND CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "GNU"))
```
ADKaster, line 158: you only used the new option in one place. 

What is this checking for?

clang.exe? clang-cl?

If it's checking for clang with GNU variant, then ... is MSVC even the right check here? Should this just check that we're targeting windows? `if (WIN32)` ?

ADKaster (changes requested):

CMakeLists.txt
Optimoos: MSVC matches clang-cl. MSVC_VERSION and GNU are meant to match clang++ on Windows. The issue is that in both scenarios I had linker memory errors and this resolved them. We could `if (WIN32)` for now, but if anyone tries to build with mingw that might get in their way, which is why I was trying to be targeted with what this will catch.

ADKaster: `if (WIN32 and NOT (MINGW OR MSYS2))`?

ADKaster: I suspect that this will also fail if you use clang.exe and tell it to use `lld.link` as the linker though, right? this is a link.exe thing only?

cmake/jakt-executable.cmake
Optimoos: Just double checked as I don't think I'd tried `/W4`, but I get the same stack of errors no matter if it's `-W4` or `/W4`. Tons of `error-unused`, `parenthesis-inequality`, things that have explicitly been shut off by other options. As I linked up top, apparently this is expected behavior for certain logging levels in MSVC.

CMakeLists.txt
Optimoos: I _thought_ that clang++ and lld-link were what got me started down this rabbit hole, but it's entirely possible I was just doing something wrong. Learned a few things in the past couple days!

I'll get mingw going and test all combinations.

Optimoos: So, AK isn't going to work with mingw in its current form. MinGW doesn't always use win32 threads, it depends which package you're using, and they're mostly posix unless you want to build your own. AK hard expects win32 threads to be available if it detects the OS as Windows. I don't know how much effort/attention should be put into this given MSVC and Clang work.

ADKaster: > I don't know how much effort/attention should be put into this given MSVC and Clang work.

None :). From cmake-bootstrap.md

> On MSYS2, g++ may error out with a "string table overflow" error. In that case, re-configure the build directory with -DCMAKE_BUILD_TYPE=MinSizeRel to get the -Os flag. Do note that using WSL2 or clang directly on windows is a more supported build platform. Maintainers will be reluctant to merge runtime or jakttest patches for MSYS2 quirks to keep the number of supported platfoms under control.

If it turns out that MSYS2 doesn't work anymore, so be it. we can replace that information with a "welp lmao" comment instead.

PR #1456, opened 2023-06-28
codegen: Nested branch `release_return` handles `void` explicitly

cybergsus:
Since C++ can't implicitly construct `ExplicitValue<void>` from `void`,
Jakt outputs explicit constructors of `ExplicitValue<void>()`. Now it
also outputs them when calling `release_return` under nested matches.

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -78,21 +78,41 @@ struct ControlFlowState {
         )
     }
     fn is_match_nested(this) -> bool => .match_nest_level != 0
-    fn apply_control_flow_macro(this, anon x: String) throws -> String {
+
+    // Instead of trying to use `release_return()` which returns `void`, we'll
+    // directly construct a `ExplicitValueOrControlFlow<MatchResult, void>` so
+    // that the compiler doesn't try to construct one from a `void` value,
+    // which it doesn't know how to do.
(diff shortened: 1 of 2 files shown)


selfhost/codegen.jakt
```diff
@@ -78,21 +78,39 @@ struct ControlFlowState {
...
     fn is_match_nested(this) -> bool => .match_nest_level != 0
-    fn apply_control_flow_macro(this, anon x: String) throws -> String {
+
+    // Instead of trying to use `release_return()` which returns `void`,
+    // we'll directly construct a `ExplicitValue<void>` so that the compiler
+    // doesn't try to construct one from a `void` value, which it doesn't
+    // know how to do.
+    fn nested_release_return_expr(match_result_type: TypeId) throws -> String {
+
+        let returns_void = match_result_type.equals(void_type_id()) or match_result_type.equals(unknown_type_id())
```
alimpfard, line 88: :thinking: I'd say leave `unknown` be, if we don't know what it is, let's not assume it's `void`.
unless that breaks something?

cybergsus: I... copied that from other section of the codegen that handled void. There were instances where stuff would return void and it would end up with unknown. Not sure if that's the case though.

alimpfard: eh, let's keep it this way then, we'll fix it if something breaks because of it.

PR #1457, opened 2023-06-29
Cpp import perf + Build for Serenity

alimpfard:
Continuation of #1449.


jakttest/run_one.py
```diff
@@ -10,8 +10,8 @@
 from pathlib import Path
 
 # .lib on windows, .a otherwise
-MAIN_LIBRARY_NAME = "libjakt_main.a" if os.name != "nt" else "jakt_main.lib"
-RUNTIME_LIBRARY_NAME = "libjakt_runtime.a" if os.name != "nt" else "jakt_runtime.lib"
+MAIN_LIBRARY_NAME = "libjakt_main_%s.a" if os.name != "nt" else "jakt_main_%s.lib"
```
ADKaster, line 13: wew, this is python3, we use f-strings or .format() instead of % strings in this household

CMakeLists.txt
```diff
@@ -73,45 +79,64 @@ if (is_multi_config)
...
-    ${JAKT_STAGE0_SOURCES}
-    ${JAKT_STAGE0_RUNTIME_SOURCES}
-    ${JAKT_STAGE0_RUNTIME_LIB_SOURCES}
-)
+set(JAKT_TARGET_TRIPLE "${JAKT_TARGET_TARGET}" CACHE STRING "Target triple to use for cross-compilation (defaults to cmake's target triple)")
 
-# Note: This currently does not build under windows, but it is not needed by the runtime.
-list(FILTER JAKT_STAGE0_ALL_SOURCES EXCLUDE REGEX ".*AK/Time\.cpp$")
-list(FILTER JAKT_STAGE0_ALL_SOURCES EXCLUDE REGEX ".*AK/DOSPackedTime\.cpp$")
+if (NOT ${JAKT_TARGET_TRIPLE} STREQUAL ${JAKT_TARGET_TARGET})
```
ADKaster, line 95: TARGET_TARGET and TARGET_TRIPLE? I know what this patch is supposed to do and I still don't understand the difference between these variables.

Is what's going on here that one of them is JAKT_HOST_TRIPLE and the other is JAKT_TARGET_TRIPLE?

We have access to CMAKE_HOST_SYSTEM_PROCESSOR and CMAKE_HOST_SYSTEM, are those what you're really looking for?

CMakeLists.txt
```diff
@@ -254,12 +284,13 @@ add_executable(Jakt::jakt ALIAS jakt_stage${FINAL_STAGE})
...
 if (NOT CMAKE_SKIP_INSTALL_RULES)
   include(cmake/install-rules.cmake)
+  jakt_install_runtime(TARGET ${JAKT_TARGET_TRIPLE})
 endif()
 
 # FIXME: Remove if we decide to use CTest
 option(JAKT_BUILD_TESTING "Whether to build tests or not, default on" ON)
 
-if (JAKT_BUILD_TESTING)
+if (JAKT_BUILD_TESTING AND NOT CMAKE_CROSSCOMPILING)
```
ADKaster, line 293: We could use a cmake_dependent_option to say "default to off if cross-compiling" but meh.

cmake/install-rules.cmake
```diff
@@ -44,45 +48,49 @@ foreach (stage IN LISTS stages)
...
-        DESTINATION "${CMAKE_INSTALL_LIBDIR}"
-        COMPONENT Jakt_Runtime
-        NAMELINK_COMPONENT Jakt_Development
-    ARCHIVE #
-        DESTINATION "${CMAKE_INSTALL_LIBDIR}"
-        COMPONENT Jakt_Development
-)
+# FIXME: foreach(target IN LISTS JAKT_RUNTIME_TARGETS) for this
+function(jakt_install_runtime)
+  cmake_parse_arguments(PARSE_ARGV 0 JAKT_INSTALL "" "TARGET" "")
```
ADKaster, line 53: It looks like the TARGET var is required? If so, we could just make it a normal argument to the function. If you want to keep it as a single-value-argument, we should at least complain (FATAL_ERROR) if it's not set.

runtime/CMakeLists.txt
```diff
@@ -71,29 +71,52 @@ set(RUNTIME_SOURCES
     Jakt/DeprecatedStringBuilder.cpp
 )
 
-add_library(jakt_runtime STATIC ${RUNTIME_SOURCES})
-add_jakt_compiler_flags(jakt_runtime)
-target_include_directories(jakt_runtime
+# FIXME: create a new external_project foreach(target IN LISTS JAKT_RUNTIME_TARGETS)
+function(make_jakt_runtime_available)
+  cmake_parse_arguments(PARSE_ARGV 0 JAKT_RUNTIME "" "TARGET" "VAR_NAME")
```
ADKaster, line 76: same comment here re:required target arg

jakttest/run_one.py
ADKaster: (if this is really the cleanest way then sure, but it strikes me as a python2-ism)

alimpfard: okay, .format() is fine too

CMakeLists.txt
alimpfard: I've probably named these in the worst way possible, here's what's going on:
- `JAKT_TARGET_TRIPLE` is the user-facing config thingy, defaults to `JAKT_TARGET_TARGET`
- `JAKT_TARGET_TARGET` is what CMake thinks the target triple is
- there's no `JAKT_HOST_TRIPLE` anymore.

CMakeLists.txt
alimpfard: I'll use your "meh" as an excuse to run away from learning about that thing :P

CMakeLists.txt
alimpfard: err...they both refer to the same thing, `TARGET_TRIPLE` is just there to let the user override it in case cmake is wrong
I can remove TARGET_TARGET and use TARGET_TRIPLE exclusively if that's what you prefer?

ADKaster: Yeah that would be preferred lol. Just let the variable from the jakt-executable.cmake file be user-settable as a cache variable and it'll be great.

PR #1458, opened 2023-06-29
JaktTest: Build directory validation

Optimoos:
This change does a test for the existence of the jakt binary before proceeding to the test run. Without this, if there was an error in the build path or if the build was not at the default location, jakttest would start anyway and begin dumping masses of python errors.

### DIFF
--- a/jakttest/jakttest.jakt
+++ b/jakttest/jakttest.jakt
@@ -354,6 +354,14 @@ struct Options {
             options.job_count = os::get_num_cpus()
         }
 
+        let jakt_path_test = fs::stat_silencing_enoent(path: String::format("{}/bin/jakt", options.build_dir))
+        let jakt_path_test_win = fs::stat_silencing_enoent(path: String::format("{}/bin/jakt.exe", options.build_dir))
+
+        // The test could be more comprehensive than this, but this works for now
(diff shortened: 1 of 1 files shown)


jakttest/jakttest.jakt
```diff
@@ -354,6 +354,14 @@ struct Options {
             options.job_count = os::get_num_cpus()
         }
 
+        let jakt_path_test = fs::stat_silencing_enoent(path: String::format("{}/bin/jakt", options.build_dir))
+        let jakt_path_test_win = fs::stat_silencing_enoent(path: String::format("{}/bin/jakt.exe", options.build_dir))
+
+        // The test could be more comprehensive than this, but this works for now
+        if (not jakt_path_test.has_value()) and (not jakt_path_test_win.has_value()) {
+            options.errors.push(String::format("Could not find jakt binary at {}/bin/jakt", options.build_dir))
+        }
```
alimpfard, line 363:
```suggestion
        let binary_path = Path::from_string(options.build_dir).join("bin/jakt")
        if not binary_path.exists() and not binary_path.replace_extension("exe").exists() {
            options.errors.push(String::format("Could not find jakt binary at {}", binary_path.to_string()))
        }
```

alimpfard: (Sidenote: `String::format` is invalid, that's a compiler bug)

PR #1459, opened 2023-06-29
CI: Workflow to build Clang on Windows

Optimoos:
This workflow will fetch, build, and cache LLVM Clang on an x64 action runner. The build needs to be done in release mode or the standard runners will not have enough space to complete the build.

This job will take approximately 3 hours and 45 minutes to complete.

### DIFF
--- a/.github/workflows/win64_llvm_clang_build_and_cache.yml
+++ b/.github/workflows/win64_llvm_clang_build_and_cache.yml
@@ -1,0 +1,50 @@
+name: Build CMake Libraries
+
+on:
+  workflow_dispatch:
+
+env:
+    LLVM_VER: 16.0.6
+    BRANCH: llvmorg-16.0.6
+
+jobs:
+
+   build:
+    runs-on: windows-latest
+ 
+    steps:
+    - uses: actions/checkout@v3
+    - name: Setup Ninja
+      uses: ashutoshvarma/setup-ninja@master
+      with:
+        # ninja version to download. Default: 1.10.0
+        version: 1.10.0
+        dest: ${{ github.workspace }}/ninja_bin
+               
+    - name: Clone LLVM
+      uses: actions/checkout@v3
+      with:
+        repository: 'llvm/llvm-project'
+        ref: ${{ env.BRANCH }}
+        path: 'llvm-project-${{ env.LLVM_VER }}'
+        fetch-depth: 1
+
+
+    - name: Configure CMake
+      shell: cmd
+      run: |
+        "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64 &^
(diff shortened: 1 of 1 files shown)


.github/workflows/win64_llvm_clang_build_and_cache.yml
```diff
@@ -0,0 +1,54 @@
+name: Build CMake Libraries
+
+on:
+  workflow_dispatch:
+
+env:
+    LLVM_VER: 16.0.4
+    BRANCH: release/16.x
```
ADKaster, line 8: we should use version 16.0.6 and the tag `llvmorg-16.0.6` as that is the newest patch release

.github/workflows/win64_llvm_clang_build_and_cache.yml
```diff
@@ -0,0 +1,54 @@
...
+
+
+    - name: Configure CMake
+      shell: cmd
+      run: |
+        "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64 &^
+        cd llvm-project-%LLVM_VER% &^
+        mkdir build &^
+        cd build &^
+        cmake -GNinja -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_C_COMPILER=clang-cl -DLLVM_ENABLE_PROJECTS=clang -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXE_LINKER_FLAGS="/machine:x64" -DCMAKE_INSTALL_PREFIX="%GITHUB_WORKSPACE%/llvm-project-%LLVM_VER%/install" ..\llvm
```
ADKaster, line 40: Instead of doing all that magic sauce with the directories

```
cmake -GNinja -S llvm-project-%LLVM_VER%\llvm -B llvm-build  <....>
```

and leave off the `../llvm` from the end.

.github/workflows/win64_llvm_clang_build_and_cache.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: Build
+      shell: cmd
+      run: |
+        "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64 &^
+        cd llvm-project-%LLVM_VER%/build &^
+        ninja clang install
```
ADKaster, line 47: again, instead of cd, use `ninja -C llvm-build clang install`

.github/workflows/win64_llvm_clang_build_and_cache.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: Cache LibClang
+      id: cache-clang
+      uses: actions/cache@v3
+      with:
+        path: 'llvm-project-${{ env.LLVM_VER }}/install'
+        key: 'libclang-${{ runner.os }}-${{ github.run_id }}'
```
ADKaster, line 54: Don't cache based on `github.run-id`, cache based on the llvm version. That way we can pull it down (restore) from other workflows in a reasonably memorable way. The run-id is apparently just some random number assigned to the workflow that's unique for the repository

>  A unique number for each workflow run within a repository. This number does not change if you re-run the workflow run

ADKaster (changes requested):

ADKaster (approved):
Thanks!

@alimpfard this will let you import a cache from id cache-clang and a restore key associated with the llvm version in windows CI to use libclang (I'm pretty sure anyway)

```
      id: cache-clang
      with:
        path: 'llvm-install-${{ env.LLVM_VER }}/'
        key: 'libclang-${{ runner.os }}-${{ env.LLVM_VER }}'
```

Issue #1334, opened 2023-02-03
Make `jakt --version` print the git revision


awesomekling:
Instead of saying `unreleased`, let's make `jakt --version` print the git revision it was built from.

lanmonster:
fixed by #1480

PR #1460, opened 2023-06-30
Jakt: Configure version number from git hash

Optimoos:
This change is my attempt at configuring the version number from the git hash. I've tried to keep the change as modular as possible and out of the way of the primary files (selfhost/main.jakt and CMakeLists.txt).

This addresses Andreas' request in issue #1334.


ADKaster:
There's a general comment about reproducible builds that should be made on this PR, but I don't have the right words for it. @nico is this the "right way" or the "Wrong way" to include a git hash in the build?

PR #1463, opened 2023-07-02
typechecker: Detect `*print*` and `format` calls exclusively on empty namespace

### LINKED ISSUE #1461: Nonexistent object method called 'format' is erroneously accepted
`String::format` is invalid, but is assumed to refer to the implicit top-level hacky `format()` function.

```
let foo = String::format("test {}", 123)
```

_See https://github.com/SerenityOS/jakt/pull/1458#discussion_r1248781476 for context_.

cybergsus:
This removes the detection of `format` from arbitrary namespaces where
the function may not exist, or be an entirely different function.

Fixes #1461.

### DIFF
--- a/jakttest/jakttest.jakt
+++ b/jakttest/jakttest.jakt
@@ -634,9 +634,9 @@ struct TestScheduler {
             run_once_script
             "jakttest/run_one.py"
             "--jakt-binary"
-            String::format("{}/bin/jakt", build_dir)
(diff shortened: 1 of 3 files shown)


cybergsus:
The diff looks funky because of some de-indentations I did to the inner code. GitHub's "Hide whitespace" functionality should help in reviewing.

alimpfard:
I've resolved the conflicts, will merge after CI passes.

PR #1464, opened 2023-07-03
typechecker: Evaluate bound expressions only once

### LINKED ISSUE #1450: `if foo() is Some(x)` runs foo() twice
The codegen for an if like:

```
if foo() is Some(x) { ... }
```

Will run `foo()` twice, causing the side effects it causes twice. We should only run the expression once in the codegen.

cc @fahrradflucht

cybergsus:
Now Jakt inserts declarations to save the bound expressions before
checking them, and uses the variable names to access the bound parts.
This ensures that the original expression isn't run more than once.

Fixes #1450.

To do this I had to remove support for bindings inside `match` cases,  part of
a feature made by @robryanx in commit
https://github.com/SerenityOS/jakt/commit/abab87133e9a1894b30fcb60cb8f76a030ef15c8, since it both allowed for weird match use cases and made the solution to this
issue more complicated.


selfhost/typechecker.jakt
```diff
@@ -6437,7 +6504,28 @@ struct Typechecker {
...
+        let if_result =  CheckedStatement::If(condition: checked_condition, then_block: checked_block, else_statement: checked_else, span)
+	
+        if saved_expr_decls.is_empty() {
+            return if_result
+        } else {
+            saved_expr_decls.push(if_result)
+
+            mut control_flow = checked_block.control_flow
+            if checked_else is Some(stmt) and stmt is Block(block: checked_else_block) {
+            control_flow = control_flow.unify_with(checked_else_block.control_flow)
```
alimpfard, line 6516: This line needs an indent

alimpfard (commented):
Seems like this breaks stage 2 :thinking:

cybergsus:
> Seems like this breaks stage 2

Yeah, I'm working on it. I've messed up some control flow based checks with producing `CheckedBlock`s manually.

cybergsus:
Sorry for the delay. The idea here was to wrap the resulting values from a guard or an `if` in a block, converting: 

```jakt
if some_computation() is Some(x) {
    do_something_with(x)
}
```

into: 

```jakt
{
   let _cached_result = some_computation()
   if _cached_result is Some(x) {
       do_something_with(x)
   }
}
```

This looks redundant in Jakt code, but internally the typechecker transforms the first snippet to something like:
```jakt
if some_computation().has_value() {
   do_something_with(some_computation())
}
```

This was hard to fix at the codegen level because it was already receiving the duplicate expressions from the typechecking stage. I opted for doing this in the typechecker, then. But by doing this I didn't get the metadata for the wrapping blocks correct, and the branch analysis of the typechecker for reachable/unreachable stuff then bit me in stage 2. 

I haven't been able to get this metadata correct and I don't think it would be good anyway, I think it would be better to somehow generate parsed expressions again and let the typechecker expand on it, or generate those parsed expressions earlier in the parsing stages.

I will be leaving this as closed, I've become burnt out after trying seemingly incoherent modifications to solve the stage 2 problem.

PR #1474, opened 2023-08-25
More work towards a serenity application in jakt

alimpfard:
Continuation of #1472.


alimpfard:
This is good enough to get the toolchain in https://github.com/SerenityOS/serenity/pull/20765 working, so let's call it ready for review.

alimpfard:
YOLO, let's just merge it.

PR #1476, opened 2023-08-31
codegen: Generate explicit tagged unions for enums

cybergsus:
Jakt now generates tagged unions to avoid instancing `AK::Variant` with as many
types as number variants held by the enum.

After these changes, a gap of ~500s can be observed in the C++ parsing times for stage 2:
image

## Reproducing

- Enable `-ftime-trace` on `stage2` build.
- Use [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer) to collect the JSON data and reveal textual summaries for both `main` and this branch.

- Use your favorite diff tool to check the difference between those summaries.

In `results.tar.gz` I include my analysis summaries results, but you'll also find a `CMakeLists.txt.patch` patch to
have those enabled by CMake.

With clang++ 15, and the above mentioned patch, you can use & tweak this build command:
```sh
cmake -B build-profile -G Ninja -DCMAKE_CXX_COMPILER=clang++-15 -DCMAKE_BUILD_TYPE=Release -DCMAKE_BUILD_WITH_INSTALL_RPATH=on -DFINAL_STAGE=2 -DTRACE_STAGE2=on
```
Make sure to leave `FINAL_STAGE=2` and `TRACE_STAGE=on` intact. `CMAKE_BUILD_TYPE=Release` is also important.

For each of the branches, do:
```sh
ClangBuildAnalyzer --start build-profile
ninja -C build-profile jakt_stage2
ClangBuildAnalyzer --stop build-profile profile.bin
ClangBuildAnalyzer --analyze profile.bin > profiled-<main|no-variants>.txt
```

Since `-ftime-trace` was active, you can check the results with speedscope or
other visualization tools.
The JSON files will be under `<build dir>/CMakeFiles/jakt_stage2.dir/jakt_stage2_tmp/`.

### Saving space

I used different directories to avoid going back and forth between `main` and
my branch, but if you want to do this as a one-shot, reusing the build
directory will save space.
From my test runs, JSON trace files for `main` occupied around `4.2G` and the ones for
`no-variant-enums` around `33M`.

### Appendix

[results.tar.gz](https://github.com/SerenityOS/jakt/files/12486348/results.tar.gz)


selfhost/codegen.jakt
```diff
@@ -2562,7 +2377,11 @@ struct CodeGenerator {
         }
         output += "debug_description() const {\n"
         output += "auto builder = DeprecatedStringBuilder::create();\n"
-        output += "switch (this->index()) {"
+        output += "switch (this->__jakt_init_index()) {"
+
+        // NOTE: This should be replaced by a set of strings and checking each
+        // if `StructLike`'s fields aren't guaranteed to contain all of the common fields first.
```
alimpfard, line 2383: wdym by this?
common fields are only guaranteed to be _present_ in all variants, the placement can be arbitrary if it helps you out.

cybergsus: By "common" fields I mean the fields that are declared in `enum_.fields`, which are duplicated only in `StructLike`'s fields. Those are found always at the start of `StructLike`'s fields, although I can switch to using a `{String}` to avoid putting a constraint there.

cybergsus: That is: ```jakt
enum SomeEnum {
    common_value: String
    VariantA(span: Span)
    VariantB(i32)
}
``` has `common_value` in both `enum_.fields` and `VariantA`'s `CheckedVariant::StructLike`

alimpfard: ah, _that_'s what you mean.

alimpfard: should be fine to assume they always come first, the codegen implementation is the only component that cares.

cybergsus: OK, so I can keep this then. Should I add a similar note (or a redirection to this one) to the rest of the places where I use this assumption? It's not the only place that has `let common_field_count = enum_.fields.size()`

alimpfard: Up to you really.
one should be enough for a "the order matters" warning imo.

cybergsus: I'll move the note to the type definition then.

alimpfard:
The windows problem should be fixed, but you're gonna have to regenerate the boostrap (sorry...)

PR #1480, opened 2023-09-14
selfhost: Report git hash for version

lanmonster:


selfhost/main.jakt
```diff
@@ -36,6 +36,18 @@ import platform_compiler() {
 
 import jakt::platform { Target }
 
+comptime get_current_git_hash() throws -> String {
+    mut head_file = File::open_for_reading("../.git/HEAD")
```
alimpfard, line 40: let's allow building outside the tree:
```suggestion
    mut head_file = try File::open_for_reading("../.git/HEAD") catch { return "unreleased" }
```

PR #1488, opened 2023-10-02
typechecker: Ensure that no name is redefined as a different symbol

### LINKED ISSUE #1485: We should fail when member field and member function have same name
The following program should fail to compile since we can't have both a member variable `foo` and a member function `foo`.

It currently only fails once we get to the C++ compiler.

```jakt
struct S {
    foo: i32
    fn foo() => true
}
```

lanmonster:
Fixes #1485

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -28,7 +28,7 @@ import types {
     builtin, comptime_format_impl, never_type_id, unknown_type_id, void_type_id
 }
 import types
-import utility { FileId, Span, add_arrays, escape_for_quotes, join, panic, todo }
+import utility { FileId, Span, add_arrays, escape_for_quotes, join, panic, todo, map }
 import jakt::path { Path }
 import compiler { Compiler }
 import interpreter { Interpreter, InterpreterScope, ExecutionResult, value_to_checked_expression }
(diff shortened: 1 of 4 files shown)


selfhost/utility.jakt
```diff
@@ -74,6 +74,14 @@ fn join(anon strings: [String], separator: String) throws -> String {
     return output
 }
 
+fn map<T, U>(input: [T], mapper: fn(item: T) throws -> U) throws -> [U] {
```
alimpfard, line 77: Please take this function as a reference so it'll play nice with trailing closures
with the trailing closure parameter support in #1474, the following will be possible if `map` takes a `&fn` instead of `fn`.
```
map(stuff) |item: i32| {
    return item + 1
}
```

selfhost/typechecker.jakt
```diff
@@ -3144,6 +3144,32 @@ struct Typechecker {
             super_struct_id: None
             external_name: parsed_record.external_name
         ))
+
+        match parsed_record.record_type {
+            Struct(fields) | Class(fields) => {
+                let field_names_and_spans = map(input: fields, mapper: fn(item: ParsedField) throws -> (String, Span) => (item.var_decl.name, item.var_decl.span))
+                let method_names_and_spans = map(input: parsed_record.methods, mapper: fn(item: ParsedMethod) throws -> (String, Span) => (item.parsed_function.name,item.parsed_function.name_span))
```
alimpfard, line 3151: a few spaces and newlines wouldn't hurt :P 
```suggestion
                let method_names_and_spans = map(
                    input: parsed_record.methods
                    mapper: fn(item: ParsedMethod) throws -> (String, Span) =>
                        (item.parsed_function.name, item.parsed_function.name_span)
                )
```

PR #1499, opened 2023-11-06
cmake: Fix build error when `clang` is not installed in path

shannonbooth:
I was getting:

```
CMake Error at CMakeLists.txt:32 (find_program):
  Could not find CLANG_PATH using the following names: clang
```

Fix this by using the clang target from the earlier call to:
```
find_package(Clang CONFIG)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,11 @@ find_package(Clang CONFIG)
 set(JAKT_CPP_AUTO_IMPORT_PROCESSOR_DEFAULT none)
 if (Clang_FOUND)
   set(JAKT_CPP_AUTO_IMPORT_PROCESSOR_DEFAULT clang)
-  find_program(CLANG_PATH clang REQUIRED)
+
+  string(REPLACE "." ";" CLANG_VERSION_LIST ${Clang_VERSION})
+  list(GET CLANG_VERSION_LIST 0 CLANG_MAJOR)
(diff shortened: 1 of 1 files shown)


shannonbooth:
hmm scratch that, I've messed it up

CMakeLists.txt
```diff
@@ -29,7 +29,11 @@ find_package(Clang CONFIG)
 set(JAKT_CPP_AUTO_IMPORT_PROCESSOR_DEFAULT none)
 if (Clang_FOUND)
   set(JAKT_CPP_AUTO_IMPORT_PROCESSOR_DEFAULT clang)
-  find_program(CLANG_PATH clang REQUIRED)
```
ADKaster, line 32: Have a look at the ClangConfig cmake file in your clang install. are there any variables there that point to the clang binary dir, or clang executable?

shannonbooth: Yeah, thanks!

I had actually been doing that when trying to make this MR. :thonk:

I can't immediately see anything super nice to use for this. There is `CLANG_INSTALL_PREFIX` which points to: `/usr/lib/llvm-17`, but then I need to know where clang is installed past that. The best option I was thinking of to try out would be to use a clang binary target from `/usr/lib/llvm-17/lib/cmake/clang/ClangTargets.cmake`, but I hadn't / haven't quite worked out how to do that yet.

PR #1523, opened 2023-12-21
  codegen: Simplify control flow forwarding rules for match IIFEs

### LINKED ISSUE #1520: C++ compile error with nested match/loop/match in function that returns void
The following code generates invalid C++:

```
fn boog() {
    match 1 {
        else => {
            loop {
                match 1 {
                    else => {
                        println("boog!")
                    }
                }
            }
        }
    }
}
```

You get an error like this:
```
build/test.cpp:22:16: error: no viable conversion from returned value of type 'void' to function return type 'JaktInternal::ExplicitValueOrControlFlow<void, void>'
        return _jakt_value.release_return();
```

cg-jl:
ControlFlowState was handling control flow forwarding in an error-prone and unintuitive manner. It also was being used to apply control flow forwarding from the scope of the match IIFE that is being called, instead of the state of the outer scope, which is where the calling and forwarding takes place.

The only tracking that is needed to apply control flow forwarding is the following two things about the parent scope:
    1. Is the current scope the outer loop where continue/break is handled or should it forward the special LoopContinue/LoopBreak signaling types?
    2. Is the current scope the function scope to return from or should it forward the special signaling return from `ExplicitValueOrControlFlow`?

Fixes #1520.

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -36,49 +36,44 @@ fn is_return_allowed(anon allowed_control_exits: AllowedControlExits) -> bool =>
 
 struct ControlFlowState {
     allowed_exits: AllowedControlExits
-    /// Whether `break` and `continue` should use `return ExplicitValueOrControlFlow` instead of
-    /// C++'s `break`/`continue`.
-    passes_through_match: bool
     passes_through_try: bool
-    match_nest_level: usize
+    // whether the current scope is a match IIFE
+    directly_inside_match: bool = false
+    // whether any outer scope is a match IIFE
+    indirectly_inside_match: bool = false
 
     fn no_control_flow() -> ControlFlowState {
         return ControlFlowState(
             allowed_exits: AllowedControlExits::Nothing
-            passes_through_match: false
             passes_through_try: false
-            match_nest_level: 0
+            directly_inside_match: false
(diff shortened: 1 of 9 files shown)


selfhost/codegen.jakt
```diff
@@ -36,49 +36,44 @@ fn is_return_allowed(anon allowed_control_exits: AllowedControlExits) -> bool =>
...
         )
     }
     fn enter_function(this) -> ControlFlowState {
         return ControlFlowState(
             allowed_exits: AllowedControlExits::JustReturn
-            passes_through_match: false
             passes_through_try: false
-            match_nest_level: .match_nest_level
+            indirectly_inside_match: .indirectly_inside_match
+            directly_inside_match: false
```
cg-jl, line 58: That looks like a bug waiting to happen. Entering a function means the scope for return changes, which means that it should set "indirectly inside match" to false as well.

cg-jl (commented):
Please till I resolve this bug too, I'll send a message in the discord when it's ready.

AtkinsSJ:
> If you're reviewing, please wait till I resolve this bug too, I'll send a message in the discord when it's ready. Thanks!

You can mark a PR as a draft and then it isn't mergeable. It's handy for situations like this. :^)

PR #1532, opened 2023-12-22
types: Don't assume all string literals will throw

### LINKED ISSUE #1530: Bad C++ generated when matching on result of NoneCoalescing expression
Simplified repro:

```cpp
fn boog() {
    let x: String? = None
    match (x ?? "") {
        else => {
        }
    }
}
```

alimpfard:
The information is already available, use that instead of assuming the worst.
Fixes #1530.


awesomekling:
Thank you!

PR #1537, opened 2023-12-22
codegen: Avoid creating temporary ByteString to match against literal

awesomekling:
When matching a vanilla string literal, we now codegen a comparison
against `"foo"sv` instead of wrapping it in a temporary `ByteString`.
This drastically cuts down on temporary string allocations. :^)

This gives ~10% speedup on `jakt -S selfhost/main.jakt`

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -13,7 +13,7 @@ import typechecker {
     never_type_id, unknown_type_id, void_type_id
 }
 import types { CheckedGenericParameter, CheckedParameter, FunctionGenericParameter, SafetyMode, StructLikeId }
-import utility { Queue, Span, interpret_escapes, join, panic, prepend_to_each, todo }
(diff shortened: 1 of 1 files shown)


selfhost/codegen.jakt
```diff
@@ -3357,7 +3361,16 @@ struct CodeGenerator {
                         }
                     } else {
                         output.append("if (__jakt_enum_value == ")
-                        .codegen_expression(expression, &mut output)
+                        if expression is QuotedString(val)
+                            and (val.type_id == byte_string_type_id or val.type_id == builtin(BuiltinType::JaktString)) {
+                            let original_string = val.to_string()
+                            let escaped_value = original_string.replace(replace: "\n", with: "\\n")
```
alimpfard, line 3367:
```diff
                            let escaped_value = escape_for_quotes(original_string)
```

`escape_for_quotes` is available in selfhost/utility.

PR #1538, opened 2023-12-22
typechecker: Emit an error when a passed argument is ignored 

alimpfard:


0GreenClover0:
Test case

```rs
/// Expect:
/// - error: "Invalid argument not used in call"

fn test(defaulted_arg: i64 = 0) {
    println("{}", defaulted_arg);
}

fn main() {
    test(12)
}
```

PR #1539, opened 2023-12-23
typechecker: Disallow usage of 'continue' and 'break' outside of loops

### LINKED ISSUE #1531: `continue` and `break` accepted in random places
This should probably not compile :sweat_smile: 

```jakt
fn boog() {
    continue
}
```

Same with
```jakt
fn boog() {
	break
}
```

0GreenClover0:
Fixes #1531

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -74,6 +74,12 @@ enum BreakContinueLegalityTracker {
     Defer
 }
 
+enum ReturnLegalityTracker {
+    None,
+    Lambda,
+    Defer
+}
+
 enum NumericOrStringValue {
     StringValue(String)
     SignedNumericValue(i64)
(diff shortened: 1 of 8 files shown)


selfhost/typechecker.jakt
```diff
@@ -6405,6 +6405,13 @@ struct Typechecker {
         Guard(expr, else_block, remaining_code, span) => .typecheck_guard(expr, else_block, remaining_code, scope_id, safety_mode, span)
     }
 
+    fn typecheck_continue(mut this, span: Span) -> CheckedStatement {
+        if not .inside_any_loop {
```
alimpfard, line 6409: This is invalid:
```rs
for foo in bar {
    defer { break }
    defer { continue }
}
```

Will have to also reject this.

alimpfard: ...but not this:
```rs
for foo in bar { defer { for baz in quux { break } } }
```

0GreenClover0:
Maybe we should do all of this in parser instead, not sure 🤔

StunxFS:
The typechecker is perfect for detecting invalid usages

PR #1550, opened 2024-01-01
Disallow assigning Optional<void>, allow yielding void functions and nothing

### LINKED ISSUE #1544: typechecker/codegen: `void?` is accepted as `void`
Given these two samples:
```
/// Expect:
/// - output: "PASS\n"

enum Foo {
    A
    B
}

fn may_throw() throws {}

fn thing() {
    let a = match Foo::A {
        A => try may_throw() catch {},
        B => {}
    }
}

fn main() {
    thing()
    println("PASS")
}
```

```
/// Expect:
/// - output: "PASS\n"

enum Foo {
    A
    B
}

fn may_throw() throws {}

fn is_void() {}

fn thing() {
    match Foo::A {
        A => try may_throw() catch { yield is_void() },
        B => {}
    }
}

fn main() {
    thing()
    println("PASS")
}
```

Both fail to compile the C++ for a couple of reasons:
    1. `try` codegen does not use statement-exprs when compiling void, so `return <try>` and `return ExplicitValue(<try>)` have statements inside them without having a block to support them.
    2. `Optional<void>` is used in both C++ outputs, the first one for the `a` variable, and the second one for the catch result value.
    
Depending on how we want to handle this, it's either  typechecker's fault or codegen's fault:
    - If we allow `void?` and `void` to be unified, then the codegen should handle these cases, maybe it needs to do so explicitly. This would be useful if we want the second sample to typecheck, although that would collide with the current behavior of requiring void match cases to be in non-yielding blocks.
    - If not, then it's obviously a type mismatch and the typechecker is not handling it correctly. Also, nothing should be assigned `void?` since that doesn't really make sense in the first place (`void?` is isomorphic to `bool`, so why use the former).

0GreenClover0:
Fixes #1544. First sample now shows an error "Cannot assign `Optional<void>` to a variable". Second sample prints "PASS".

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -4580,12 +4580,15 @@ struct CodeGenerator {
 
                 let (var_name, end_label) = .entered_yieldable_blocks.last()!
 
-                if expr.type() != void_type_id() {
-                    output.append(var_name)
-                    output.append(" = ")
+                if expr.has_value() {
+                    if expr!.type() != void_type_id() {
+                        output.append(var_name)
+                        output.append(" = ")
+                    }
+
+                    .codegen_expression(expr!, &mut output)
                 }
 
-                .codegen_expression(expr, &mut output)
                 output.append("; goto ")
                 output.append(end_label)
                 output.append(";\n")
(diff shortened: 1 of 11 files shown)


0GreenClover0:
Not stale

alimpfard:
I'm not 100% on how we want to treat void, but we can fix (or break) that in the future if needed. LGTM.

PR #1551, opened 2024-01-02
interpreter: Reserve throwing for soft errors

cg-jl:
"not implemented" and "invalid type" should not be throwing and instead
panicking, since getting to those means the typechecker failed at its
job. When wanting a checked expression from a void value, it can be
handled manually if needed, before calling in the function.

Also, there are parts of the interpreter that throw when encountering a
user-facing error like unwrapping a None. These should forward the error
using the `*Result` values from the interpreter, since that's the
mechanism for handling errors that are raised when interpreting user
code.


cg-jl:
:(

PR #1557, opened 2024-02-28
nix: fix build and add flake

sreehax:
This change makes it insanely simple for anyone with nix to download and run jakt. The existing jakt.nix meant for nix-shell was a great starting point, but it only let you get a dev shell with all the dependencies rather than letting you build and install the project too. The new workflow allows commands such as `nix build`, `nix develop`, and `nix run`, which do exactly what they sound like they do. Additionally, this git repo could easily be added as a flake input for home-manager or NixOS to allow system-wide installation with the included nixpkgs overlay. Legacy nix command (`nix-shell`, `nix-build`, etc) compatibility is retained for those who wish to keep using `nix-shell` or `nix-build`. I have elected to use clang 16, but that can be changed easily if preferred.


jakt.nix
```diff
@@ -1,13 +1,26 @@
...
-    pkgconfig
-    cmake
-    ninja
-    python3
-    clang_15
+stdenv.mkDerivation (finalAttrs: let
+  serenity = fetchFromGitHub {
+    owner = "serenityos";
+    repo = "serenity";
+    rev = "05e78dabdbceea46bae7dca52b63dc0a115e7b52"; # latest at the time
```
ADKaster: Do we really need to pin the Serenity version? generally the only expectation is that jakt main builds with serenity master. This means that the env will break without notice at any time, and anyone using this will have to update this rev and hash any time they want to spin up an env again.

flake.nix
```diff
@@ -0,0 +1,57 @@
...
+        inherit default jakt;
+      }
+    );
+    devShells = forAllSystems (system:
+      let
+        pkgs = nixpkgs.legacyPackages.${system};
+      in {
+        default = pkgs.mkShell {
+          inherit (self.packages.${system}.default) buildInputs nativeBuildInputs;
+        };
```
ADKaster, line 54: This is a lot of boilerplate. is it all necessary? Serenity's Toolchain/flake.nix is a lot simpler.

flake.nix
```diff
@@ -0,0 +1,57 @@
+{
+  description = "The Jakt Programming Language";
+  inputs = {
+    nixpkgs.url = github:NixOS/nixpkgs/nixos-unstable;
+    flake-utils.url = github:numtide/flake-utils;
+    gitignore = {
+      url = github:hercules-ci/gitignore.nix;
```
ADKaster, line 7: Why do we need a gitignore inside the flake? is the one for the top level directory not sufficient?

default.nix
```diff
@@ -0,0 +1,10 @@
+(import
+  (
+    let lock = builtins.fromJSON (builtins.readFile ./flake.lock); in
+    fetchTarball {
+      url = lock.nodes.flake-compat.locked.url or "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+      sha256 = lock.nodes.flake-compat.locked.narHash;
+    }
+  )
+  { src = ./.; }
+).defaultNix
```
ADKaster, line 10: I don't understand. why do legacy users need these files? We don't have them for the Serenity/Toolchain implementation.

ADKaster (commented):
@networkException do you have any time to review this PR?

RGBCube:
I created a PR to make this PR much simpler & use better practices here: https://github.com/sreehax/jakt/pull/1

jakt.nix
sreehax: To ensure reproducible builds, yes we do need to pin the Serenity version. I understand that this is a new change, however consider the possibility that Serenity could have a breaking change in master that would cause jakt to fail to build without any warning. By removing a pinned serenity version, we would have to remove support all support for nix builds.

sreehax: I realize that hardcoding a git revision here isn't exactly the prettiest solution to this problem. Perhaps git submodules could be used?

ADKaster: I still don't like pinning this, but that's the nature of nix I suppose. If something in serenity breaks jakt, then we should fix it rather than ignoring it until the next time someone bothers to update this hash. Git submodules are significantly worse than any other solution, especially for working across repos.

ADKaster (approved):
Thanks for improving the nix experience on jakt! I've force pushed some commit cleanup. The 'please don't iterate over multiple commits' rule should apply here in jakt as well :)

PR #1558, opened 2024-03-16
nix: Fix building and create flake

RGBCube:
nix: Fix permission error when the builder copies from nix store

Files in the nix store have weird permissions.

nix: Replace jakt.nix with flake

This allows for reproducible builds and insane ease of use. Now, anyone with a Linux machine with nix installed can just run `nix build` and easily build or install jakt.

Further, jakt contributors can easily get a shell with all the dependencies just by running `nix develop`.

nix: Add legacy shell.nix and default.nix

For anyone using nix-build or nix-shell instead of the new nix commands.

nix: Add wrapper

This should fix running jakt without clang++ in PATH, which could happen when using nix.

nix: Fix PR that fixes Nix


RGBCube:
Closing this as my PR to https://github.com/SerenityOS/jakt/pull/1557 got merged.

PR #1559, opened 2024-03-18
CMake: Always grab the resource-dir from the Clang we find_package'd

ADKaster:
If the developer has another `clang` in their path that is different
from the one that ends up getting found via find_package, this could
lead to some really unexpected behavior.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,10 @@ find_package(Clang CONFIG)
 set(JAKT_CPP_AUTO_IMPORT_PROCESSOR_DEFAULT none)
 if (Clang_FOUND)
   set(JAKT_CPP_AUTO_IMPORT_PROCESSOR_DEFAULT clang)
-  find_program(CLANG_PATH clang REQUIRED)
+  find_program(CLANG_PATH clang REQUIRED
+    PATHS "${CLANG_INSTALL_PREFIX}/bin"
+    NO_DEFAULT_PATH
+  )
   execute_process(
     COMMAND
       ${CLANG_PATH} -print-resource-dir
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,9 @@ find_package(Threads REQUIRED)
 
 SET(CMAKE_FIND_PACKAGE_SORT_ORDER NATURAL)
 SET(CMAKE_FIND_PACKAGE_SORT_DIRECTION DEC)
-find_package(Clang CONFIG)
+find_package(Clang CONFIG
+  PATHS /opt/homebrew/opt/llvm
+)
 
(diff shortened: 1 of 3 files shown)


ADKaster:
@alimpfard this patch also means that if you *do* end up passing something like `-DClang_DIR=/path/to/dir/with/ClangConfig.cmake`, then we will use the correct resource path for that clang.

CMakeLists.txt
```diff
@@ -29,7 +29,10 @@ find_package(Clang CONFIG)
 set(JAKT_CPP_AUTO_IMPORT_PROCESSOR_DEFAULT none)
 if (Clang_FOUND)
   set(JAKT_CPP_AUTO_IMPORT_PROCESSOR_DEFAULT clang)
-  find_program(CLANG_PATH clang REQUIRED)
+  find_program(CLANG_PATH clang REQUIRED
+    PATHS "${CLANG_INSTALL_PREFIX}/bin"
```
ADKaster, line 35: CLANG_INSTALL_PREFIX is set (and exported, gee thanks) by ClangConfig.cmake.

I also really hope this is auto-generated because... nice.

```cmake
# Compute the installation prefix from this LLVMConfig.cmake file location.
get_filename_component(CLANG_INSTALL_PREFIX "${CMAKE_CURRENT_LIST_FILE}" PATH)
get_filename_component(CLANG_INSTALL_PREFIX "${CLANG_INSTALL_PREFIX}" PATH)
get_filename_component(CLANG_INSTALL_PREFIX "${CLANG_INSTALL_PREFIX}" PATH)
get_filename_component(CLANG_INSTALL_PREFIX "${CLANG_INSTALL_PREFIX}" PATH)
```

alimpfard: What on earth am I looking at

ADKaster: that's ClangConfig.cmake.

```
serenity$ cat /opt/homebrew/Cellar/llvm/17.0.6_1/lib/cmake/clang/ClangConfig.cmake
# This file allows users to call find_package(Clang) and pick up our targets.

# Compute the installation prefix from this LLVMConfig.cmake file location.
get_filename_component(CLANG_INSTALL_PREFIX "${CMAKE_CURRENT_LIST_FILE}" PATH)
get_filename_component(CLANG_INSTALL_PREFIX "${CLANG_INSTALL_PREFIX}" PATH)
get_filename_component(CLANG_INSTALL_PREFIX "${CLANG_INSTALL_PREFIX}" PATH)
get_filename_component(CLANG_INSTALL_PREFIX "${CLANG_INSTALL_PREFIX}" PATH)

set(LLVM_VERSION 17.0.6)
find_package(LLVM ${LLVM_VERSION} EXACT REQUIRED CONFIG
             HINTS "${CLANG_INSTALL_PREFIX}/lib/cmake/llvm")
```

PR #1560, opened 2024-04-26
Collection of codegen improvements

cg-jl:
This is a collection of small changes that have made no performance impact, but
that I have made to achieve better readability on the generated code. I
originally intended to make these into a proper PR, but uni started hitting me
and it won't stop for at least one month more.

I don't intend to merge these changes as-is, but there are some things that we
might want to pull from, or may serve as a list of low-hanging fruits for other
devs to get into the weeds. I'll be separating the couple of changes that fix
bugs I found in the codegen so they can be put into a different PR and properly
merged.

- codegen: Disable use of TRY() macro when returning from a throwing function
- types: Keep the match bodies grouped by patterns in the typechecker
- codegen: Codegen branch body once
- codegen: Avoid generating forward namespace if nothing is inside
- codegen: Use temporary variable only for yielding from try
- codegen: Change accessor when binding a common enum member
- typecheck: Search for common fields in non-StructLike pattern arguments **FIXES BUG**
- codegen: Use the correct name to search for common enum members **FIXES BUG**
- bootstrap: Update stage0 snapshot
- lexer: Use common enum member for Token
    **NOTE**: This is not something I want to commit, like its description says
    it is here to magnify the effects of needless inclusion, by repeating potential pathological cases 50 times, and thus make them easier to spot.
    Makes no difference in compilation time or final binary performance.
- codegen: Skip generating variant cases for variants without data
- codegen: Remove uninit check in enums
- codegen: Make common enum fields be struct fields directly


alimpfard:
shoo bot, is draft.

cg-jl:
I started pulling the first thing that fixes binding common enum members and realized that we don't support it, so that one might have been me adding more code than necessary.

PR #1571, opened 2024-07-06
selfhost: Add support for `export`s

cg-jl:
Supports exporting types to handy header files, following the design in the SerenityOS discord: https://discord.com/channels/830522505605283862/983664833096466462/1258917037988777985

This PR supports only exporting types. Exporting functions and namespaces is left out for now.

Things to  fix (sorted by relevance):
- ~~support exporting generic types~~ inherently supported by just `using name;`
- ~~specify an export directory other than `<builddir>/exports`~~ done by alimpfard
- ~~find out how to get to the build directory from the export directory, so that includes work without setting the build directory as include dir~~ done by alimpfard
- ~~group exports by namespace tree, not by type.~~ Discarded, it's pretty but not needed

### DIFF
--- a/editors/emacs/jakt-mode.el
+++ b/editors/emacs/jakt-mode.el
@@ -43,7 +43,7 @@
              (jakt-repeat '("while" "for" "loop"))
              (jakt-execution '("return" "break" "continue" "throw" "yield"))
              (jakt-boolean '("true" "false"))
-             (jakt-keyword '("fn" "extern" "comptime"))
+             (jakt-keyword '("fn" "extern" "comptime" "import" "export"))
              (jakt-exception '("throws"))
              (jakt-macro '("defer" "unsafe" "throw" "try" "must" "catch" "cpp"))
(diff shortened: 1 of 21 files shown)


alimpfard:
You didn't add the vscode syntax rules (which I assume you made, based on the commit message) :P

windows: might just be a problem with creating the export directory?

cg-jl:
> You didn't add the vscode syntax rules (which I assume you made, based on the commit message) :P

I must've missed writing the buffer before staging & committing, at least it's recent enough that I remember what I wrote, thanks :P

> windows: might just be a problem with creating the export directory?

It should be fixed now. It was reusing the `results` dictionary, and any file in the `results` dictionary is a relative path from `build_dir`. I don't know how it worked under Linux, but after some println I got this funny message:

```
trying to open 'C:\Users\gsus\jakt\Testie\jakttest-tmp-15\C:\Users\gsus\jakt\Testie\jakttest-tmp-15\export_helper\Exported.h' as write binary
```

PR #1578, opened 2024-07-12
typechecker: Run `IsEnumVariant` LHS only once

### LINKED ISSUE #1450: `if foo() is Some(x)` runs foo() twice

cg-jl:
Generates temporary variables for `IsEnumVariant` LHS expressions and
adds them to the previous block.  This way pattern bindings use the
cached temporary and avoid re-evaluating any potential side effects.

Fixes #1450.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -160,6 +160,12 @@ struct Typechecker {
     had_an_error: bool = false
     cpp_import_cache: [String:ScopeId] = [:]
     cpp_import_processor: CppImportProcessor? = None
+    temp_var_count: u64 = 0
+    // The current statement block, available to typecheck_expression functions so that they can
+    // push temporary variables for `IsEnumVariant` checks.
+    // Note that this mechanism relies on Array being refcounted, so that the pushes that
+    // are done to `.current_block` are visible after the expression is checked.
+    current_block: [CheckedStatement] = []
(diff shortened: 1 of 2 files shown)


selfhost/typechecker.jakt
```diff
@@ -153,6 +153,10 @@ struct Typechecker {
     had_an_error: bool = false
     cpp_import_cache: [String:ScopeId] = [:]
     cpp_import_processor: CppImportProcessor? = None
+    temp_var_count: u64 = 0
+    // NOTE: This relies on Array being refcounted (shared).
```
alimpfard, line 157: where?
I see the copy below (which won't actually copy the array), but that's not so much "relies" as it is a bug?

cg-jl: oh, I'll be more explicit about that. I meant the mechanism of setting the current  block so statements can be added to it, I thought I explained it. Maybe putting the note on the paragraph

cg-jl: Anyway, looks like I forgot to run the tests locally, so I'll move this to draft and mark when it's resolved again, thanks :)

PR #1580, opened 2024-07-13
Make returning Optional, WeakPtr and None from match less of a pain. Fix for subclasses and classes

### LINKED ISSUE #1139: Arrow function type inference does not handle match yielding None
the following program compiles and works as expected.
```
function foo() -> bool? => match 0 {
    1 => false
    else => None
}

function main() {
    let a = foo()
}
```
however, if you take off the return type, it fails in the c++ compiler
```
runtime/lib.h:225:17: error: no matching constructor for initialization of 'ExplicitValue<bool>'
        : value(ExplicitValue<Value> { move(v.value) })
                ^                    ~~~~~~~~~~~~~~~~~
build/test.cpp:13:8: note: in instantiation of function template specialization 'JaktInternal::ExplicitValueOrControlFlow<bool, bool>::ExplicitValueOrControlFlow<Jakt::NullOptional>' requested here
return JaktInternal::ExplicitValue(JaktInternal::OptionalNone());
       ^
runtime/lib.h:198:8: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'Jakt::NullOptional' to 'const JaktInternal::ExplicitValue<bool>' for 1st argument
struct ExplicitValue {
       ^
runtime/lib.h:198:8: note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'Jakt::NullOptional' to 'JaktInternal::ExplicitValue<bool>' for 1st argument
struct ExplicitValue {
       ^
runtime/lib.h:199:5: note: candidate constructor not viable: no known conversion from 'Jakt::NullOptional' to 'bool' for 1st argument
    ExplicitValue(Value&& v)
    ^
runtime/lib.h:203:5: note: candidate constructor not viable: no known conversion from 'Jakt::NullOptional' to 'const bool' for 1st argument
    ExplicitValue(Value const& v)
    ^
1 error generated.
```

0GreenClover0:
2b7e74b
6e2ba49
Previously we would wrongly assume that given:
```rs
class A { }
class B : A { }
```
assignment of `A` to a variable of type `B` would be fine which led to a C++ compiler error.

acf7229
6f43208
When yielding ex. an `Optional<T>` or a `WeakPtr<T>` out of the match, if we already inferred the result type to be `T`, we would just print an error that types are incompatible. On the other hand, if we first inferred the result type to be ex. `T?`, everything would be fine. Inferring the "most compatible" type seems more reasonable. This not only works for Optionals, WeakPtrs, but also for classes and subclasses since the code is based on the `check_types_for_compat` method (which works correctly after the fixes in the first two commits).

5d80411
Yielding `None` if the result type is not directly declared is ambiguous (it can either mean an `Optional` or a `WeakPtr`.
This previously led to C++ compiler errors.

Fixes #1139

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -5469,7 +5469,7 @@ struct Typechecker {
                 }
             }
             Struct(lhs_struct_id) => {
-                if lhs_type_id.equals(rhs_type_id) or .is_subclass_of(ancestor_type_id: rhs_type_id, child_type_id: lhs_type_id) {
+                if lhs_type_id.equals(rhs_type_id) {
                     return true
                 }
 
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -9530,7 +9530,7 @@ struct Typechecker {
         arguments_span: Span
         scope_id: ScopeId
         safety_mode: SafetyMode
-    ) throws -> (String?, CheckedMatchCase, TypeId?) {
+    ) throws -> (String?, CheckedMatchCase, TypeId?, bool) {
         mut covered_name: String? = None
 
         let new_scope_id = .create_scope(
@@ -9657,7 +9657,7 @@ struct Typechecker {
             defaults.push(checked_var_decl)
         }
 
-        let (checked_body, result_type) = .typecheck_match_body(
+        let (checked_body, result_type, seen_none) = .typecheck_match_body(
             body: case_.body
             scope_id: new_scope_id
             safety_mode
@@ -9677,7 +9677,7 @@ struct Typechecker {
             marker_span: case_.marker_span
         )
 
-        return (covered_name, checked_match_case, result_type)
+        return (covered_name, checked_match_case, result_type, seen_none)
     }
 
     fn typecheck_match(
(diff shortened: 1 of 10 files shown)


selfhost/typechecker.jakt
```diff
@@ -5469,7 +5469,7 @@ struct Typechecker {
                 }
             }
             Struct(lhs_struct_id) => {
-                if lhs_type_id.equals(rhs_type_id) or .is_subclass_of(ancestor_type_id: rhs_type_id, child_type_id: lhs_type_id) {
+                if lhs_type_id.equals(rhs_type_id) {
```
alimpfard, line 5472: don't we still need this?
it should just be `is_subclass_of(lhs, rhs)` instead?

0GreenClover0:
"And I would have gotten away with it too, if it weren't for meddling stage1!"

selfhost/typechecker.jakt
0GreenClover0: The version with (lhs, rhs) is already there, just little below

alimpfard: Ah I see

PR #1589, opened 2024-07-20
typechecker: Choose broader type in arrays, sets and dicts

### LINKED ISSUE #1353: typechecker: Array type inference should infer base class instead of derived class when callee expects base class
This currently chokes in the C++ compiler:
```
class Base {}
class Derived: Base {}

fn func(anon a: [Base]) {}

fn main() {
    func([Derived()])
}
```
The problem is that `[Derived()]` creates a `[Derived]` when we really want `[Base]` since that is what the callee expects.

0GreenClover0:
Fixes #1353

### DIFF
--- a/tests/typechecker/function_base_class_set_argument_derived_given.jakt
+++ b/tests/typechecker/function_base_class_set_argument_derived_given.jakt
@@ -1,0 +1,20 @@
+/// Expect:
+/// - output: "PASS\n"
+
+class Base implements(Hashable, Equal<Base>) {
+    public fn hash(this) => 42u32
+    public fn equals(this, anon other: Base) -> bool => true
+}
+
+class Derived implements(Hashable, Equal<Derived>) : Base {
+    public fn hash(this) => 42u32
+    public fn equals(this, anon other: Derived) -> bool => true
+}
+
+fn func(anon a: {Base}) {
+    println("PASS")
+}
+
+fn main() {
+    func({Derived()})
+}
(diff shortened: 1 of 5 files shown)


0GreenClover0:
Alternatively we can just always use the type hint id if it exists, I think

tests/typechecker/function_base_class_set_argument_derived_given.jakt
```diff
@@ -0,0 +1,24 @@
+/// Expect:
+/// - output: "PASS\n"
+
+trait Hashable<Output> {
+    fn hash(self) -> Output
+}
+
```
alimpfard, line 7: uhh...this definitely shouldn't be accepted, this is not the "correct" definition of Hashable as assumed by the compiler

alimpfard: I mean, defining that trait is fine, but `{T}` where `T implements(Hashable<i64>)` should not be accepted

0GreenClover0: Hm, so how this should look like?

alimpfard: `Hashable` is already in scope, imported from the prelude, it should just `implements(Hashable)`

0GreenClover0: It took me a sec to figure out I'm missing `this` in `public fn hash(this) => 42u32`
Now I know why it didn't work for me before :p

PR #1592, opened 2024-08-01
Add index [] trait, disallow usage of indexing on types that don't implement it

0GreenClover0:


selfhost/typechecker.jakt
```diff
@@ -8972,6 +8972,10 @@ struct Typechecker {
                     let array_slice_struct_id = .find_struct_in_prelude("ArraySlice")
                     let dictionary_struct_id = .find_struct_in_prelude("Dictionary")
 
+                    if not id.equals(array_struct_id) and not id.equals(array_slice_struct_id) and not id.equals(dictionary_struct_id) {
+                        .error(format("Type ‘{}’ does not implement the indexing operation", .type_name(checked_base.type())), span)
```
alimpfard, line 8976: it's impossible to "implement" the indexing operation (at least atm), so maybe:
```suggestion
                        .error(format("Type ‘{}’ cannot be indexed by []", .type_name(checked_base.type())), span)
```

0GreenClover0: Now it's possible to implement it

alimpfard: well...I suppose it is :)

tests/typechecker/array_index_good_throwing_trait.jakt
```diff
@@ -0,0 +1,25 @@
...
+
+struct Vector3 {
+    private values: [f32]
+
+    public fn create(anon x: f32, anon y: f32, anon z: f32) -> Vector3 {
+        return Vector3(values: [x, y, z])
+    }
+
+    public fn test() throws {
+    }
```
alimpfard, line 12: Is this necessary?

runtime/jaktlib/prelude/operators.jakt
```diff
@@ -39,6 +39,14 @@ trait ThrowingDivide<Rhs, Output> {
     fn divide(this, anon rhs: Rhs) throws -> Output
 }
 
+trait Index<Output> {
+    fn index(this, anon index: usize) -> Output
+}
+
+trait ThrowingIndex<Output> {
+    fn index(this, anon index: usize) throws -> Output
+}
```
alimpfard, line 48: We have indexing with non-usize keys (for dictionaries), how about:
```suggestion
trait Index<Key, Output> {
    fn index(this, anon index: Key) -> Output
}

trait ThrowingIndex<Key, Output> {
    fn index(this, anon index: Key) throws -> Output
}
```

and following that,
```
type Array implements(Index<usize, T>) {
    [[name="operator[]"]]
    extern fn index(this, anon index: usize) -> T
}

type Dictionary implements(Index<K, V>) {
    [[name="operator[]"]]
    extern fn index(this, anon index: K) -> V
}
```

selfhost/codegen.jakt
```diff
@@ -2767,12 +2767,22 @@ struct CodeGenerator {
...
+                        &mut output
+                        forward_error_with_try
+                    )
+                } else {
+                    output.append("((")
+                    .codegen_expression(expr, &mut output)
+                    output.append(")[")
+                    .codegen_expression(index, &mut output)
+                    output.append("])")
+                }
```
alimpfard, line 2785: ...which would allow dropping this branch entirely.

runtime/jaktlib/prelude/operators.jakt
0GreenClover0: Eh, codegen puts parameter in the wrong place
 ```cpp
 ((((test).values))[](static_cast<i64>(1LL))));
 ```
 let's see how to fix that...

alimpfard: with `operator[]`? it should be calling `.operator[](...)` with that :thinking:

0GreenClover0: ah, I did `operator([])` by mistake

PR #1593, opened 2024-08-04
typechecker: Typecheck `for` loop directly into CheckedAST

cg-jl:
NOTE: This PR updates bootstrap to be able to fearlessly use `if <x> is Variant(y)`, although on request I can convert the `guard x is Some` and such to use a temporary variable in between.

Now Jakt generates CheckedAST directly when typechecking a for loop,
instead of generating ParsedAST to be re-typechecked by our generic
typechecker. This allows us to state explicitly what we want the code to
be shaped like, not just in terms of the parsing, but also in terms of
the typing, since the typecheck involves traits. It also allows for more
thorough assertions on how the typed side of the code is expected to
look like.
    
Like intuition would indicate, the runtime aspect is expected to at
least go a bit better. To ensure that I wasn't making the runtime worse
with this change, I benchmarked the time and general perf counters for
typechecking selfhost since it contains ~1030 for loops. It was
surprising to see a ~10MB decrease in peak memory usage, over a general
7-8% (1.10s new vs 1.18s current) improvement in other performance
characteristics, like cache/branch misses, instructions processed and
wall time. Benchmarks were done using https://github.com/andrewrk/poop.


cg-jl:
yeah, I will not submit this, at least not in this state. Besides the memory usage due to the amount of data processed, I didn't get what I was looking for. In particular I made the code more complex and I removed a very little portion of work from the current code. If I want a bigger change, it's more relevant that I suppress all the trait searching for prelude types like arrays, dictionaries and ranges.

PR #1596, opened 2024-08-22
codegen+runtime: Make C++ output easier to parse visually

cg-jl:
- codegen: Hint when parentheses are needed in codegen_expression
- codegen: Avoid generating extra parentheses with error handlers
- codegen: Use `this->` on accessors when appropiate

Codegen added a lot of extra opening parentheses that made expressions inside `while`/`if` hard to parse visually due to a large number of opening parentheses. Now codegen does not emit parentheses on most cases that it doesn't need to, it's only missing cases that are based on binary operator precedence.
These commits removed ~2000 parentheses and ~100kB worth of selfhost C++ code.

- runtime: Make integral comparisons use C++ operators

Due to the inlined comparison functions that used `.compare` with a cast, any Jakt expression that had a `<` (e.g the condition for `while bytes_written < writes_read`) would emit two nested IIFEs because it had two infallible conversions to and from `Ordering` plus a comparison to the `Ordering` enum. Now the integral comparisons are explicitly overloaded to use `<` and the like so the compiler doesn't generate the IIFEs that make the conditions in `while` and `if` unreadable.
This commit alone manages to suppress another ~100kB worth of selfhost C++ code.

The new compile times are within the noise threshold of the old ones.
The new generated code does not affect the C++ parser, but the resulting C++ code is smaller, easier to read and more intuitive to search through. This in turn makes checking whether a modification in codegen.jakt was successful or not faster, since the code is more greppable.

### DIFF
--- a/runtime/jaktlib/prelude/operators.jakt
+++ b/runtime/jaktlib/prelude/operators.jakt
@@ -192,6 +192,18 @@ type T implements(
         return .compare_impl(rhs) as! Ordering
     }
 
+    [[name="operator(<)"]]
+    extern fn less_than(this, anon rhs: T) -> bool
+
+    [[name="operator(<=)"]]
+    extern fn less_than_or_equal(this, anon rhs: T) -> bool
+
+    [[name="operator(>)"]]
+    extern fn greater_than(this, anon rhs: T) -> bool
+
+    [[name="operator(>=)"]]
+    extern fn greater_than_or_equal(this, anon rhs: T) -> bool
+
     [[name="operator(+=)"]]
     extern fn add_assign(mut this, anon rhs: T) -> void
 
--- a/samples/basics/escape_sequence.jakt
+++ b/samples/basics/escape_sequence.jakt
@@ -2,7 +2,7 @@
 /// - output: "hello\nworld"
 
 fn main() {
-    let str_c: c_char = '\n'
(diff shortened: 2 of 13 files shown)


alimpfard:
Tests are failing because of some AK changes, [this](https://github.com/SerenityOS/jakt/pull/1595/commits/a40d2a27670eff6b5a0171e203afe25f7e55183b#diff-2ccdf1410db0d3a7d32bd68071c6060e12f221adfed757de4bf2e022565bafebR156), [this](https://github.com/SerenityOS/jakt/pull/1595/commits/a40d2a27670eff6b5a0171e203afe25f7e55183b#diff-16fda0fe34bd7d97af9a1ba1024ecb3d1eaffc229fa314881e1924e97e62461fR1) and [this](https://github.com/SerenityOS/jakt/pull/1595/commits/a40d2a27670eff6b5a0171e203afe25f7e55183b#diff-9713ae70fac2072b1e57293f3f8bdfa3a03c0facce95a99395e9040f2c280488R1) should fix it.
Feel free to yoink those into your PR if you want.

cg-jl:
oh, thanks! I forgot that I didn't have an up to date serenity repo 😅

alimpfard:
Very nice, looks great!
will merge after CI passes.

PR #1601, opened 2024-10-21
codegen: Convert `operator[]` to C++ index operator

cg-jl:
So that array indexing doesn't awkwardly call `operator[](index)`,
which is what it does since the Index trait was introduced.

### DIFF
--- a/selfhost/codegen.jakt
+++ b/selfhost/codegen.jakt
@@ -4489,6 +4489,16 @@ struct CodeGenerator {
         let expression_type = .program.get_type(expr.type())
         let name = call.name_for_codegen()
 
+		let use_cpp_index_operator = name.as_name_for_use() == "operator[]" and call.args.size() == 1 
+
+        guard not use_cpp_index_operator else {
+            .codegen_expression(expr, &mut output, syntactically_self_contained: true)
+            output.append('[')
+            .codegen_expression(call.args[0].1, &mut output)
+            output.append(']')
+            return
+        }
+
         let object_is_this = .expr_codegens_to_this_pointer(expr)
 
         let field_accessor = match name {


selfhost/codegen.jakt
```diff
@@ -4489,6 +4489,17 @@ struct CodeGenerator {
         let expression_type = .program.get_type(expr.type())
         let name = call.name_for_codegen()
 
+        if name.as_name_for_use() == "operator[]" {
+            guard call.args.size() == 1 else {
```
alimpfard, line 4493: how about
```suggestion
        guard name.as_name_for_use() != "operator[]" or call.args.size() != 1 else {
```

to allow `operator[]` with more than 1 argument? support for this is not landing in C++ any time soon, but why panic when we can just not.

cg-jl: I guess I could just write all the arguments separated by commas into the braces, then.

cg-jl: nvm, I will just use the equivalence where it's currently supported and leave the rest to be an `operator[](x, y)` call as you suggested.

cg-jl: Using `guard` to ensure the `return` maybe needs some naming to be easily parsed.

alimpfard: > I guess I could just write all the arguments separated by commas into the braces, then.

Ah that's currently incorrect (`x[foo, bar, baz]` is just `x[baz]` with whatever side effects `foo` and `bar` have).
calling `operator[](foo, bar, baz)` is the "correct" approach here.

alimpfard: And sure, the `if` is fine too, equivalent anyway.

PR #1602, opened 2024-10-21
typechecker+codegen: Emit new bodies for bindings with different types

cg-jl:
Fixes cases where the body is not duplicated when it should,
specially with instantiations that are inferred inside the body for
functions that support both instantiations and overloading. C++ can deal
with overloading, but we add the template parameters ourselves, which
makes the typing wrong for at least one pattern.

Introduces `BindingKey`, `MatchBuilder` and `BindingKeyBuilder`, which
together help keep match body code duplication at a low rate while being
as least intrusive as possible to the logic for each match typecheck:
`typecheck_match` has three big `for` loops with bodies that walk the
match cases in very different ways, and the deduplication system is
orthogonal to all of them. Most of it is implemented between
`MatchBuilder` and `BindingKeyBuilder`.

These builder states avoid creating scopes and duplicate typechecker
work, resulting in ~14% performance increase and 20MB less peak memory
usage when typechecking selfhost:

image

`poop` is https://github.com/andrewrk/poop .
`stage1_main` is the result of compiling stage1 from d65f014cc54b986f629fe676d914af01d442b9f7.

### DIFF
--- a/documentation/pattern-match-avoid-duplication.md
+++ b/documentation/pattern-match-avoid-duplication.md
@@ -1,0 +1,293 @@
+# How Jakt avoids code duplication in pattern matching codegen
+
+## Current solution
+
+Typechecker walks the patterns to get the type information out of them as it
+did before, but instead of adding binding variables to a scope, it adds
+them to a `BindingKeyBuilder`. Typechecker "submits" bindings to the builder,
+which then produces a `BindingKey` that tells us whether the set of bindings
+that were declared already exist within the same pattern group.
+
+Since we don't want to be passing a lot of state regarding the match to helper
+functions like `.typecheck_match_variant`, `BindingKeyBuilder` only contains a
+reference to the slice of cases that have been declared and refer to the same
+`ParsedMatchBody`. Remember that it could be multiple due to Constraint 1.
+
+To enforce that the set of cases checked by `BindingKeyBuilder` are the correct
+set, and that context is not lost when reading through the long loop bodies of
+`.typecheck_match`, all the state regarding match typecheck is in
+`MatchBuilder`. This struct has two functionalities:
+
+1.  It will not let us create a `BindingKeyBuilder` until we provide a proof
+    that we have started a case. Such proof tracks the start of the case
+    slice that `BindingKeyBuilder` should use. This is in place to avoid copy-paste
+    mishaps, since the logic for traversing the matches is big and changes depending on
+    the type to be matched.
+
+2.  When finishing a pattern, we give it a `BindingKey` along a match body and
+    said pattern, and it will register the pattern into the `CheckedMatchCase`
+    that the `BindingKey` tells it to. The parsed body will be (re-)typechecked
+    against the bindings for this finished pattern only when the bindings of the
+    new pattern have different types than the rest of the previously checked
+    patterns.
+
+`BindingKeyBuilder` keeps track of what bindings were declared by the
+typechecker, in case we have to generate a new `CheckedMatchBody`, and the set
+of candidates that could match the set of bindings that have been submitted.
+
+Since `BindingKeyBuilder` works in a binding-per-binding basis, we could have
+multiple sets that match the partial set of bindings. The `finish()` method is
+called when the build is done, so we know that there are no more bindings
+flying in and we can make a final pass on our candidates, selecting only the
+`CheckedMatchCase` that has the same amount of bindings as what we've gathered
+so far.
+
+The usage of `BindingKeyBuilder` helps us reduce the amount of dictionaries
+that are created and deleted during the typechecking phase, since builders that
+stay in their `Empty` state do not contain references to empty
+dictionaries.
+
+
+## Constraints
+
+The solution above is designed to feed codegen with more information on the
+structure of matches, avoid duplicate work whenever possible and keep
+correctness while doing so.
+
+We can summarize these features in two constraints, discussed in this section
+after a brief on pattern matching terminology.
+
+### Terminology
+
+
+```jakt
+match <expr> {
+    <pat1>(<binding1>...) | <pat2>(<binding2>...) => <body1>
+    <pat3> => <body2>
+}
+```
+
+- `expr` is the "matched value", or "value". It is the value we wish to branch
+on.
+- `pat1`, `pat2` and `pat3` are "patterns". They are different structures that
+the matched value may conform to.
+- `binding1` and `binding2` are "bindings". They are variables that extract
+different parts of the structure that the value is checked to conform to. We
+can then use those variables inside `<body1>`.
+- `body1` and `body2` are "bodies". They are the pieces of code that we want to
+execute on certain cases.
+- `pat1` and `pat2` form a "pattern group". Matching either of them will branch
+to the code specified in `<body1>`.
+- Pattern groups and bodies together form a "case". Each case has a single body
+and a group of patterns.
+
+
+### Constraint 1: Multiple binding types
+
+Jakt allows us to use bindings of different types in the same pattern group, as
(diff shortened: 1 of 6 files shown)


cg-jl:
I'm asking for more help to review the document. I'll be redoing it since the last three subsections are actually the design doc, which is hard to read, and the rest could go later as a way to describe the constraints of the problem and why those constraints.

selfhost/typechecker.jakt
```diff
@@ -7492,9 +7494,9 @@ struct Typechecker {
         Garbage => (None, expr)
     }
 
-    fn typecheck_var_decl(mut this, var: ParsedVarDecl, init: ParsedExpression, scope_id: ScopeId, safety_mode: SafetyMode, span: Span) throws -> CheckedStatement {
+    fn typecheck_var_decl_without_adding(mut this, var: ParsedVarDecl, init: ParsedExpression, scope_id: ScopeId, safety_mode: SafetyMode, span: Span) throws -> (CheckedVariable, CheckedExpression)? {
```
alimpfard, line 7497: nit: adding what?

selfhost/typechecker.jakt
```diff
@@ -12681,3 +12661,159 @@ fn dump_scope(anon scope_id: ScopeId, anon program: &CheckedProgram, indent: i64
...
+            yield BindingKeyBuilder::from_set(correct: next_correct, built, cases)
+        }
+        New(built: built_) => {
+            mut built = built_
+            built[name] = var_id
+            yield BindingKeyBuilder::New(built)
+        }
+    }
+
+    // NOTE: This could require IntoIterator, but Range (0..cases.size() in Empty case above) is Iterable, not IntoIterator.
```
alimpfard, line 12732: 🤔 aside: maybe we should make `Iterable<T>` implement `IntoIterator<T>`?

documentation/pattern-match-avoid-duplication.md
```diff
@@ -0,0 +1,266 @@
...
+            eprintln("result: {}", result)
+        }
+    }
+}
+```
+
+Notice that the first body of the match accepts both types of `a` and `b`
+(`i32` and `u64`):
+
+- `a + b` typechecks as long as `a` implements `Add` with `b` as its RHS.
```
alimpfard, line 59:
```suggestion
- `a + b` typechecks as long as the type of `a` implements `Add` with `b` as its RHS.
```

documentation/pattern-match-avoid-duplication.md
```diff
@@ -0,0 +1,266 @@
...
+
+
+## A previous solution: Why is it slow?
+
+The first implementation ran through all of the cases, typechecking the body
+for each of the patterns and submitting a new case per pattern, with their
+respective body.
+
+When generating code for a `match`, codegen sees a `CheckedExpression::Match`
+containing a list of `CheckedMatchCase`s. Each `CheckedMatchCase` exactly one
```
alimpfard, line 84:
```suggestion
containing a list of `CheckedMatchCase`s. Each `CheckedMatchCase` contains exactly one
```

documentation/pattern-match-avoid-duplication.md
```diff
@@ -0,0 +1,266 @@
...
+would receive the bodies already duplicated, severely limiting what can be done
+at late codegen stage. Take into account that the codegen stage has the most
+complete version of the program that Jakt can have, since it needs all the
+concrete types that are used in order to generate working C++. In #1597 we can
+see that a ~300 line Jakt function (`typecheck_binary_operation`) is compiled
+to ~5500 lines of C++ code! That happened because the bodies for each case are
+large, and the bodies were duplicated many times.
+
+To ensure that similar functions to `typecheck_binary_operation` do not make
+the C++ compiler sit through 18x the amount of Jakt code, we'll add a second
```
alimpfard, line 99:
```suggestion
the C++ compiler sift through 18x the amount of Jakt code, we'll add a second
```
?

documentation/pattern-match-avoid-duplication.md
```diff
@@ -0,0 +1,266 @@
...
+`||`.
+
+Now each `CheckedMatchCase` contains a list of `CheckedMatchPattern` and one
+body that uses the bindings declared within said pattern. The implementation
+typechecks all bodies like the last version, but only selects the last checked
+body.
+
+The problem with this approach is that now the typechecker portion of this
+solution does not respect Constraint 1, since the last checked body is not
+enough to represent the case if bindings from the same parsed pattern group can
```
alimpfard, line 163:
```suggestion
enough to represent the case if bindings from the same parsed pattern group
```

documentation/pattern-match-avoid-duplication.md
```diff
@@ -0,0 +1,266 @@
...
+have different binding types.
+
+This makes `tests/codegen/match_enum_same_body_different_impl.jakt` to fail
+codegen, since it calls `maybe_templated<JaktArray<i32>>` on all cases, when
```
alimpfard, line 167:
```suggestion
This causes `tests/codegen/match_enum_same_body_different_impl.jakt` to fail,
since it codegens a call to `maybe_templated<JaktArray<i32>>` on all cases, when
```

documentation/pattern-match-avoid-duplication.md
```diff
@@ -0,0 +1,266 @@
...
+
+Since we don't want to be passing a lot of state regarding the match to helper
+functions like `.typecheck_match_variant`, `BindingKeyBuilder` only contains a
+reference to the slice of cases that have been declared and refer to the same
+`ParsedMatchBody`. Remember that it could be multiple due to Constraint 1.
+
+To enforce that the set of cases checked by `BindingKeyBuilder` are the correct
+set, and that context is not lost when reading through the long loop bodies of
+`.typecheck_match`, all the state regarding match typecheck is in
+`MatchBuilder`. This struct has two functionalities: - It will make us
```
alimpfard, line 198: where second functionality

alimpfard (commented):
I'm not sure what you mean by it being hard to read, I think it's pretty comprehensive and easy to read (ignoring some grammatical mistakes, will point out a couple I noticed).

And yeah, it'd be nicer to move the "history" after the current design doc I think - assuming we're treating this as mainly a design doc and not a history post :)

selfhost/typechecker.jakt
cg-jl: Without... registering the variable in a scope. I'll rename the other version to `typecheck_and_register_var_decl`, or inline it if it's only used in one place.

selfhost/typechecker.jakt
cg-jl: Just saw #1603. If you end up finding a solution you're comfortable with I'll wait for your fix and then fix this immediately, otherwise I can leave it as a FIXME.

documentation/pattern-match-avoid-duplication.md
cg-jl: oh, I incorrectly removed the bullet point for the second paragraph.

cg-jl:
I'll do the rest of things that require access to the repo this afternoon. Thanks for the comprehensive review!

selfhost/typechecker.jakt
alimpfard: Nah, that's a big mess I don't want to get into right now

selfhost/typechecker.jakt
alimpfard: Sure, any names works so long as it's less ambiguous

cg-jl:
I've adjusted the comments in typechecker since I documented the functions before I introduced the BindingKeys. I also missed an annotation I used to remind myself to make the design doc

PR #1613, opened 2025-04-21
cmake: Bump CXX version to 23

Hendiadyoin1:
This is needed for some upcoming upstream AK changes using deducing-this

### DIFF
--- a/cmake/jakt-executable.cmake
+++ b/cmake/jakt-executable.cmake
@@ -15,7 +15,6 @@ function(add_jakt_compiler_flags target)
     -Werror
     -Wno-dangling-reference
     -Wno-deprecated-declarations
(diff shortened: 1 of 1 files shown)


alimpfard:
Maybe we should just add -Wno-ignored-value to the jakt-compiled code...

Hendiadyoin1:
Added a commit for that (+some shuffle for OCD sake/my indecision where in the list this warning belongs)

Hendiadyoin1:
https://github.com/microsoft/STL/blob/5762e6bcaf7f5f8b5dba0a9aabf0acbd0e335e80/stl/inc/memory#L532-L549
wth MSVC?
will need to fix this upstream ig
did not expect that code to run into msvc

PR #1618, opened 2025-07-24
editors/vscode: Rate limit run compiler tasks

cg-jl:
This prevents the extension from clogging the user's computer. Since avg
request on the compiler is 1s, rate limit is set to 500ms. Might want to
make it configurable.


alimpfard:
:thinking: we already have timeouts for compiler invocations though? 7a5867bcff43ab8170edb8dfc86a7546d737065b

PR #1619, opened 2025-07-26
ide+jakttest: Fix hover mislabeling

cg-jl:
Fixing a couple of bugs I'm seeing while using the extension. These ones are on hover:

- [x] Loop variable: hover resolves to something related to the range expression.

Before:

loop-var-bad

After:

loop-var-good

Images are from `typechecker::typecheck_inline_cpp`.

- [x] Loop variables when range comes from tuple destructuring in outer loop.
   Before:
loop variable from outer loop

After:

loop variable from outer loop, fixed

`codegen::codegen_struct`

- [x] Tuple destructure: it's resolving the internal VarDecl with the joined names.
Before:
loop variable from destructuring, fixed

After:

image

`codegen::codegen_enum`

- [x] parameters in `is` expression are also resolved to the temporaries created within.
      Maybe we want to flag variables as 'autogenerated'?

Before:

temporaries in `is` expr

After:

temporaries in `is` expr, fixed

`codegen::codegen_enum_destroy_variant`

---

I guess I could use this opportunity to setup testing. It can be done with shell + jq & get more complex as we need it.


selfhost/typechecker.jakt
```diff
@@ -6932,144 +6932,118 @@ struct Typechecker {
                     expr: range
                     call: ParsedCall(name: "iterator")
                     is_optional: false
-                    span: name_span
+                    span: range.span()
                 )
             }
         } else {
             resolved_iterable_result_type = iterable_trait_implementation!.implemented_type_args[0]
         }
```
cg-jl, line 6941: diff got messy here. I:
- removed the comments above the rewrite
- turned rewritten statement from `ParsedStatement::Block` into `ParsedBlock`
- patched a couple of spans so that ide resolves correctly.

selfhost/typechecker.jakt
```diff
@@ -7670,7 +7644,7 @@ struct Typechecker {
             loop_pre.control_flow = loop_pre.control_flow.updated(if_or_break.control_flow())
             loop_pre.statements.push(if_or_break)
 
-            return CheckedStatement::Loop(block: loop_pre, span)
+            return CheckedStatement::Loop(block: loop_pre,  span)
```
alimpfard, line 7661: extra space :)

selfhost/types.jakt
```diff
@@ -1167,6 +1167,8 @@ class CheckedVariable {
     public definition_span: Span
     public type_span: Span?
     public visibility: CheckedVisibility
+    // NOTE: could this be a flag in CheckedVisibility?
```
alimpfard, line 1170: I think it's fine to keep outside, I think visibility should stay semantics-only.

alimpfard (commented):
mostly lgtm, just a couple nits:

PR #1620, opened 2025-10-04
everywhere: Allow descructuring assignments

robryanx:
Currently we allow descructuring declarations for example:
```
let (x, y) = (1, 2)
```
This extends so that we can do destrucuting assignments for example:
```
mut x = 0
mut y = 0
(x, y) = (1, 2)
```
Additionally we can do nested destructuring for example:
```
mut x = 0
mut y = 0
mut z = 0
(x, (y, z)) = (1, (2, 3))
```
Will add this for declaration destructuring as part of a separate PR.

Note: This is similar to an older PR: https://github.com/SerenityOS/jakt/pull/1161
The main difference being that it specifically only allows `Var` or `JaktTuple` for nesting on the LHS. Also the earlier PR had fairly significant parser changes but this version handles it within the typechecker without introducing parsing changes.

### DIFF
--- a/selfhost/typechecker.jakt
+++ b/selfhost/typechecker.jakt
@@ -6743,7 +6743,7 @@ struct Typechecker {
         Continue(span) => .typecheck_continue(span)
         Break(span) => .typecheck_break(span)
         VarDecl(var, init, span) => .typecheck_and_register_var_decl(var, init, scope_id, safety_mode, span)
-        DestructuringAssignment(vars, var_decl, span) => .typecheck_destructuring_assignment(vars, var_decl, scope_id, safety_mode, span)
+        DestructuringDeclaration(vars, var_decl, span) => .typecheck_destructuring_declaration(vars, var_decl, scope_id, safety_mode, span)
         If(condition, then_block, else_statement, span) => .typecheck_if(condition, then_block, else_statement, scope_id, safety_mode, span)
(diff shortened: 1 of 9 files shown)


alimpfard:
Very nice, I'll take a more comprehensive look in a bit but generally looks good.

selfhost/typechecker.jakt
```diff
@@ -6730,7 +6730,20 @@ struct Typechecker {
...
+            // Detect tuple assignment expressed as a BinaryOp with Assign and a tuple on the LHS,
+            // and handle it via the tuple-assignment path.
+            match expr {
+                BinaryOp(lhs, op, rhs) => {
+                    if op is Assign and lhs is JaktTuple(values) {
+                        return .typecheck_tuple_assignment(lhs_values: values, rhs, scope_id, safety_mode, span)
+                    }
+                }
+                else => {}
+            }
```
alimpfard, line 6743: maybe better as an if?
```suggestion
            if expr is BinaryOp(lhs, op, rhs) and op is Assign and lhs is JaktTuple(values) {
                return .typecheck_tuple_assignment(lhs_values: values, rhs, scope_id, safety_mode, span)
            }
```

alimpfard (commented):
Just one comment!

alimpfard:
Thanks!

